PR #1, opened 2020-11-11
Initial implementation.

ro-tex:
# Overview

This is the initial implementation of a bare-bones accounts system. It's a really rudimentary implementation whose main goals are:
1. Get the conversation going.
2. Show something concrete, so we can build on it.
3. Crash into the first couple of walls, so we have more information going forward.

# Description

* The MR introduces the `User`, `Email`, `API`, and `DB` types that are going to be foundational in the future.
* For `User` we have the basic (limited) CRUD + find.
* No password handling, yet - we need to talk about that first.

The handlers are also limited - only user creation + GET are available at the moment.

# Food for thought

Given the amount of boilerplate code we'll have to write, we need to think really hard about writing that ourselves vs taking a ready-made package that deals with that.

Feature that we need to implement:
* secure password storage (hashing, slating, etc.)
* password policy limitations
* forgotten/reset password mechanic
* 2FA

# Alternative identity approaches:

* SkyID
* https://www.w3.org/TR/did-use-cases/


database/database.go
```diff
@@ -0,0 +1,255 @@
...
+
+var (
+	// SkynetDBHostEV holds the name of the environment variable for DB host.
+	SkynetDBHostEV = "SKYNET_DB_HOST"
+	// SkynetDBPortEV holds the name of the environment variable for DB port.
+	SkynetDBPortEV = "SKYNET_DB_PORT"
+	// SkynetDBUserEV holds the name of the environment variable for DB username.
+	SkynetDBUserEV = "SKYNET_DB_USER"
+	// SkynetDBPassEV holds the name of the environment variable for DB password.
+	SkynetDBPassEV = "SKYNET_DB_PASS"
```
ChrisSchinnerl, line 28: I'd probably call them `EnvDBHost`, `EnvDBPort` etc. to make it easier to find using autocomplete. Also `Skynet` is redundant within this application. In the string it's fine.

database/database.go
```diff
@@ -0,0 +1,255 @@
...
+	filter := bson.D{{"_id", oid}}
+	dr, err := db.users.DeleteOne(ctx, filter)
+	if err != nil {
+		return false, errors.AddContext(err, "failed to Delete")
+	}
+	return dr.DeletedCount == 1, nil
+}
+
+// UserFindAllByField finds a user by their ID.
+func (db *DB) UserFindAllByField(ctx context.Context, fieldName, fieldValue string) ([]*user.User, error) {
```
ChrisSchinnerl, line 119: I'd make this a `managedUsersByField` and then have custom `UserByEmail`, etc. methods. To essentially make it feel like any other Go type. That way you don't need any additional information about the fields.

user/user.go
```diff
@@ -0,0 +1,49 @@
...
+
+// Validate validates an email address.
+func (e Email) Validate() bool {
+	return EmailValidatorRegEx.MatchString(string(e))
+}
+
+// SetPassword sets the user's password.
+func (u *User) SetPassword(pw string) error {
+	// TODO Implement
+	u.password = sha256.Sum256([]byte(pw))
```
ChrisSchinnerl, line 47: We are going to move to a framework that does this for us right? 
If not we should use bcrypt with some good salt and maybe pepper

user/user.go
```diff
@@ -0,0 +1,49 @@
...
+	Hash [32]byte
+
+	// User represents a Skynet user.
+	User struct {
+		// ID is a hexadecimal string representation of the MongoDB id assigned
+		// to this user object. It is auto-generated by Mongo on insert.
+		ID        primitive.ObjectID `bson:"_id"`
+		FirstName string             `bson:"firstName" json:"firstName"`
+		LastName  string             `bson:"lastName" json:"lastName"`
+		Email     Email              `bson:"email" json:"email"`
```
ChrisSchinnerl, line 34: I think to keep it simple we might only need a first and last name for the MVP

database/database.go
```diff
@@ -0,0 +1,255 @@
...
+func (db *DB) Disconnect(ctx context.Context) error {
+	if db.db == nil {
+		return ErrNoDBConnection
+	}
+	return db.db.Client().Disconnect(ctx)
+}
+
+// UserDeleteByID deletes a user by their ID and returns `false` in case there
+// was an error or the user was not found.
+func (db *DB) UserDeleteByID(ctx context.Context, id string) (bool, error) {
```
ChrisSchinnerl, line 102: I don't think we are ever going to expose the ID through the API right? So we can probably provide a `User` as input here since we need to fetch it first anyway.

api/handlers.go
```diff
@@ -0,0 +1,111 @@
...
+	// TODO Limit to the user themselves.
+	if err := req.ParseForm(); err != nil {
+		WriteError(w, errors.New("Failed to parse parameters."), http.StatusBadRequest)
+	}
+	email := req.Form.Get("email")
+	if email == "" {
+		WriteError(w, errors.New("No email provided."), http.StatusBadRequest)
+		return
+	}
+	users, err := api.DB.UserFindAllByField(req.Context(), "email", email)
```
ChrisSchinnerl, line 26: Queries like that should probably be abstracted behind a specific interface method. 

e.g. `user, err := api.staticDB.UserByEmail(email)`

api/api.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/NebulousLabs/skynet-accounts/database"
+
+	"github.com/NebulousLabs/skynet-accounts/build"
+
+	"github.com/julienschmidt/httprouter"
+)
+
+// API is ...
+type API struct {
+	DB     *database.DB
```
ChrisSchinnerl, line 18: API doesn't need to expose its fields right? If not they should be staticDB and staticRouter according to Sia conventions.

api/handlers.go
```diff
@@ -0,0 +1,111 @@
...
+	if !email.Validate() {
+		WriteError(w, user.ErrInvalidEmail, http.StatusBadRequest)
+		return
+	}
+	user := &user.User{
+		FirstName: req.PostFormValue("firstName"),
+		LastName:  req.PostFormValue("lastName"),
+		Email:     email,
+	}
+	ins, err := api.DB.UserSave(req.Context(), user)
```
ChrisSchinnerl, line 54: Same here. I'd make this a `UserCreate` method to hide the database code within the database package. That would also make sure that the user's email is unique.

database/database.go
```diff
@@ -0,0 +1,255 @@
...
+		return false, errors.AddContext(err, "failed to Update")
+	}
+	if u.ID.IsZero() {
+		if ur.UpsertedID == nil {
+			log.Printf("Failed to Upsert a new user. UpsertResult: %+v\n", ur)
+			return false, errors.New("failed to upsert a new user")
+		}
+		u.ID = ur.UpsertedID.(primitive.ObjectID)
+	}
+	insertedNewUser := ur.UpsertedCount == 1
```
ChrisSchinnerl, line 239: sanity check for >1 ?

database/database.go
```diff
@@ -0,0 +1,255 @@
...
+	True = true
+	// False is a helper, so we can easily provide a *bool for UpdateOptions.
+	False = false
+)
+
+// DB represents a MongoDB database connection.
+type DB struct {
+	ctx   context.Context
+	db    *mongo.Database
+	users *mongo.Collection
```
ChrisSchinnerl, line 54: If the fields are never assigned to after init they should be static and otherwise they need a mutex.

ChrisSchinnerl (commented):
lgtm overall

Makefile
```diff
@@ -0,0 +1,106 @@
...
+release:
+	go install -tags='netgo' -ldflags='-s -w $(ldflags)' $(release-pkgs)
+release-race:
+	GORACE='$(racevars)' go install -race -tags='netgo' -ldflags='-s -w $(ldflags)' $(release-pkgs)
+release-util:
+	go install -tags='netgo' -ldflags='-s -w $(ldflags)' $(release-pkgs) $(util-pkgs)
+
+test:
+	go test -short -tags='debug testing netgo' -timeout=5s $(pkgs) -run=. -count=$(count)
+test-long: clean fmt vet lint-ci
```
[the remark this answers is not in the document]
ro-tex: I don't think so. The mental division I make is the following:
- `test`: short unit tests that are quickly executable and people should generally run all of them before each push.
- `test-long`: long test which might set up some internal structures and mocks. People should run the ones that are affected by their PR before pushing.
- `test-int`: integration tests which stand up all external services that we rely on, so we can test against them without using any mocks.

We can also merge the last two. I've set them apart for now because it seems useful to me to have the integration tests be separate and outline entire user stories.

api/handlers.go
```diff
@@ -0,0 +1,111 @@
...
+	if err := req.ParseForm(); err != nil {
+		WriteError(w, errors.New("Failed to parse parameters."), http.StatusBadRequest)
+	}
+	email := req.Form.Get("email")
+	if email == "" {
+		WriteError(w, errors.New("No email provided."), http.StatusBadRequest)
+		return
+	}
+	users, err := api.DB.UserFindAllByField(req.Context(), "email", email)
+	if err != nil && err != database.ErrUserNotFound {
```
m-cat, line 27:
```suggestion
	if err != nil && !errors.Contains(err, database.ErrUserNotFound) {
```

api/handlers.go
```diff
@@ -0,0 +1,111 @@
...
+	WriteJSON(w, users[0])
+}
+
+// userHandlerPOST creates a new user.
+func (api *API) userHandlerPOST(w http.ResponseWriter, req *http.Request, ps httprouter.Params) {
+	if err := req.ParseMultipartForm(64 * 1_000_000); err != nil {
+		WriteError(w, errors.New("Failed to parse multipart parameters."), http.StatusBadRequest)
+		return
+	}
+	email := (user.Email)(req.PostFormValue("email"))
```
m-cat, line 44: Feel like we should have a constructor for `user.Email` in case we add more logic to it.

api/routes.go
```diff
@@ -0,0 +1,14 @@
+package api
+
+// buildHTTPRoutes registers all HTTP routes and their handlers.
+func (api *API) buildHTTPRoutes() {
+	// TODO Rate-limit these from the start to prevent brute force and other forms of abuse.
+	api.Router.POST("/login", api.userLoginHandler)
+	api.Router.POST("/user", api.userHandlerPOST)
+	api.Router.GET("/user/", api.userHandlerGET)
```
m-cat, line 8:
```suggestion
	api.Router.GET("/user", api.userHandlerGET)
```
I think

api/api.go
```diff
@@ -0,0 +1,77 @@
...
+func New() *API {
+	router := httprouter.New()
+	router.RedirectTrailingSlash = true
+
+	api := &API{
+		Router: router,
+	}
+	api.buildHTTPRoutes()
+
+	db, err := database.New(context.Background())
```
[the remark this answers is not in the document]
ro-tex: That makes sense. What would be a good value? I would go for a 10s per DB operation and 30s per API call.

database/database.go
ro-tex: Makes sense.

api/routes.go
MSevey, line 8: Should this be `/user/:id` ?

database/database.go
```diff
@@ -0,0 +1,255 @@
...
+	}
+	// Upsert the user.
+	fields := bson.M{
+		"firstName": u.FirstName,
+		"lastName":  u.LastName,
+		"email":     u.Email,
+	}
+	if !u.ID.IsZero() {
+		fields["_id"] = u.ID
+		// If the user is new (u.ID is zero) then we can filer by email because
```
MSevey, line 221:
```suggestion
		// If the user is new (u.ID is zero) then we can filter by email because
```

user/user_test.go
```diff
@@ -0,0 +1,22 @@
...
+// TestEmail_Validate ensures email validation functions as expected.
+func TestEmail_Validate(t *testing.T) {
+	tests := []struct {
+		email string
+		valid bool
+	}{
+		{"valid@example.com", true},
+		{"valid@nebulous.tech", true},
+		{"invalid_nebulous.tech", false},
+		{"", false},
```
MSevey, line 14: useful additions https://gist.github.com/cjaoude/fd9910626629b53c4d25

database/database.go
ro-tex: Done.

database/database.go
```diff
@@ -0,0 +1,255 @@
...
+	opts, err := connectionOptionsFromEnv()
+	if err != nil {
+		return nil, errors.AddContext(err, "failed to get all necessary connection parameters")
+	}
+	return NewCustom(ctx, opts[SkynetDBUserEV], opts[SkynetDBPassEV], opts[SkynetDBHostEV], opts[SkynetDBPortEV], SkynetDBName)
+}
+
+// NewCustom returns a new DB connection based on the passed parameters.
+func NewCustom(ctx context.Context, user, pass, host, port, dbname string) (*DB, error) {
+	connStr := fmt.Sprintf("mongodb://%s:%s@%s:%s/?compressors=zlib&gssapiServiceName=mongodb&readPreference=nearest", user, pass, host, port)
```
MSevey, line 68: Could be a helper. Since there are other params in there it might be worth making those constants so then the string creation is more clear as to what the inputs are.

user/user.go
ro-tex: Oh, yes! This is just a placeholder, hence the "TODO" above.
@MSevey proposed an interesting idea in this direction on Discord (#accounts).

user/user.go
ro-tex: Having the email here is critical, IMO, because it is going to be our unique registration identifier, i.e. your username. Otherwise we can give people their internal IDs but those are hardly user-friendly and also I'd rather not expose those at all as they are implementation-depenedent.

Additionally, having emails here and requiring account verification via email will prevent a lot of potential abuse via automatic creation of bogus accounts in order to go around the speed limits.

database/database.go
ro-tex: Right, I was thinking about that. Yes, it makes sense. And if the user doesn't have an ID in it yet then we'll just error out.

api/handlers.go
ro-tex: Done. `staticDB` is also a good idea.

api/api.go
ro-tex: Done.

api/handlers.go
ro-tex: Yes, will do. I left this as `save` because it satisfies both `create` and `update` but it will be messy in the long run.

database/database.go
ro-tex: Done.

database/database.go
ro-tex: Makes sense.

The same should be valid for the `User` type. But there we want to lock it form the `DB` type. Would it be acceptable to expose the mutex in this case?

api/handlers.go
ro-tex: Hm, yes, it might be nice to have the `invalid email` error front-loaded. I'll add that.

api/routes.go
ro-tex: It doesn't really matter because the httprouter will redirect accordingly. But I agree it looks better. ;)

database/database.go
```diff
@@ -0,0 +1,255 @@
...
+	// SkynetDBName defines the name of Skynet's database.
+	SkynetDBName = "skynet"
+	// SkynetDBUsersCollection defines the name of the "users" collection within
+	// skynet's database.
+	SkynetDBUsersCollection = "users"
+
+	// ErrUserNotFound is returned when we can't find the user in question.
+	ErrUserNotFound = errors.New("user not found")
+	// ErrEmailAlreadyUsed is returned when we try to use an email to either
+	// either create or update a user and another user already uses this email.
```
m-cat, line 39:
```suggestion
	// create or update a user and another user already uses this email.
```

api/api.go
```diff
@@ -0,0 +1,77 @@
...
+	api.DB = db
+	if err != nil {
+		log.Println("Failed to connect to the database! Running without a database.")
+		// Assign an empty database instance. In its methods we can check for
+		// the availability of the DB and return errors instead of crashing.
+		api.DB = &database.DB{}
```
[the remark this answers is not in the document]
ro-tex: My idea was that for people who want to run a portal but don't want to run premium accounts or any kind of accounts, MongoDB won't be strictly necessary. This could be someone who's running a personal Skynet portal or an internal Skynet portal for their company/team. If they don't want to limit their users it's perfectly fine to forego Mongo and run with it.

Thinking about it again, those people won't really need to run `user-accounts` at all. I'll remove this and clean up the extra checks.

api/routes.go
ro-tex: I'm not sure about these routes at all.

Yes, probably `/user/:id` would be the right thing to do and then limit that to be accessed only by the user with that id. For that we need some kind of session code first. I'll change it to have `:id` and I'll leave it open for the moment with a TODO to limit it.

api/api.go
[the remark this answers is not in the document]
ro-tex: OK, this led me to stop embedding the context in the DB struct - we'll just pass a new derived context each time.

user/user_test.go
ro-tex: Wow, some of these are really terrible. I'm kind of torn about this. On one hand, I'd like to allow all valid email addresses. On the other, I don't want to allow invalid ones. Even the author's own suggested regex doesn't catch the odd ones and that ultimate regex he links to doesn't work out of the box in Go and I don't think it's worth it to invest the time to fix and escape it.

OTOH, if your email really is along the lines of `very.”(),:;<>[]”.VERY.”very@\ "very”.unusual@strange.example.com` then maybe not being serviced is kind of on you at that point? :)

There's also the question of why is `email@example.web` invalid.

Anyway, thanks a lot for the suggestion, it was a very fun rabbit hole to dive into! I think I'll add some of the sane suggestions but for the moment I don't think I'll change the regex. Unless someone offers a better one, of course.

m-cat (approved):

database/database.go
ro-tex: I'm not sure I completely follow.

Do you mean that the connection string can be moved to a helper method? I guess we can do that, I'm not 100% sure about the benefit if it's only used here.

Moving the other params to constants makes sense.

user/user_test.go
MSevey: > There's also the question of why is email@example.web invalid.

`.web` isn't a valid TLD i believe was the answer there. 

yea, some of those are really really wild haha

MSevey (approved):
LGTM

user/user.go
ChrisSchinnerl: Right I actually meant "not" instead of "only" ha. So only email and password should be fine in the beginning I think.

database/database.go
ChrisSchinnerl: Is the `User` type ever shared between threads? Does it even need a `Mutex`? Isn't every user I fetch from the DB a new copy of the current state within the DB?

user/user.go
ro-tex: Oh, right, yes. Those are optional anyway - the email is the only required field.

database/database.go
ro-tex: For the moment there are no threads int he service. I don't know if we'll ever need them.

Yes, the user struct is a new copy of the DB state. I'm not sure if it needs a mutex.

database/database.go
```diff
@@ -0,0 +1,304 @@
...
+	"gitlab.com/NebulousLabs/errors"
+	"go.mongodb.org/mongo-driver/bson"
+	"go.mongodb.org/mongo-driver/bson/primitive"
+	"go.mongodb.org/mongo-driver/mongo"
+	"go.mongodb.org/mongo-driver/mongo/options"
+)
+
+var (
+	// EnvDBHost holds the name of the environment variable for DB host.
+	EnvDBHost = "SKYNET_DB_HOST"
```
ChrisSchinnerl, line 22: These are exported but not used outside of the `database` package.

They should probably be moved to `main.go` anyway where environment variables are usually read from. Then main calls the constructor with the connection string.

database/database.go
```diff
@@ -0,0 +1,304 @@
...
+	mongoWriteConcern = "majority"
+	// mongoWriteConcernTimeout specifies a time limit, in milliseconds, for
+	// the write concern to be satisfied.
+	mongoWriteConcernTimeout = "1000"
+
+	// DBName defines the name of Skynet's database.
+	DBName = "skynet"
+	// DBUsersCollection defines the name of the "users" collection within
+	// skynet's database.
+	DBUsersCollection = "users"
```
ChrisSchinnerl, line 48: Same here. Everything should default to not being exported unless required.

database/database.go
```diff
@@ -0,0 +1,304 @@
...
+	ErrEmailAlreadyUsed = errors.New("email already in use by another user")
+	// ErrGeneralInternalFailure is returned when we do not want to disclose
+	// what kind of error occurred. This should always be coupled with another
+	// error output for internal use.
+	ErrGeneralInternalFailure = errors.New("general internal failure")
+
+	// True is a helper, so we can easily provide a *bool for UpdateOptions.
+	True = true
+	// False is a helper, so we can easily provide a *bool for UpdateOptions.
+	False = false
```
ChrisSchinnerl, line 63: These are not even used right?

user/user.go
```diff
@@ -0,0 +1,60 @@
+package user
```
ChrisSchinnerl, line 1: shouldn't this live in `database/user.go` ?

api/api.go
```diff
@@ -0,0 +1,89 @@
...
+	router.RedirectTrailingSlash = true
+
+	api := &API{
+		staticRouter: router,
+	}
+	api.buildHTTPRoutes()
+
+	ctx, cancel := context.WithTimeout(context.Background(), DefaultTimeoutDB)
+	defer cancel()
+	db, err := database.New(ctx)
```
ChrisSchinnerl, line 43: The database should be passed in. It should probably also be an interface that's passed in. That way it's more modular in case we ever swap out the database in the future. It also allows for easier mocking.

Makefile
```diff
@@ -0,0 +1,106 @@
...
+
+# markdown-spellcheck runs codespell on all markdown files that are not
+# vendored.
+markdown-spellcheck:
+	git ls-files "*.md" :\!:"vendor/**" | xargs codespell --check-filenames
+
+# lint runs golangci-lint (which includes golint, a spellcheck of the codebase,
+# and other linters), the custom analyzers, and also a markdown spellchecker.
+lint: markdown-spellcheck lint-analyze
+	golangci-lint run -c .golangci.yml
```
ChrisSchinnerl, line 42: This file is still missing right? I'd get linting working asap to avoid the issues we have in Sia right now. Where it's already so late that we are having troubles enabling it.

ChrisSchinnerl (approved):
Did a bit of a more thorough review. Feel free to resolve the comments in a follow up.

database/database.go
ro-tex: I was using one of them. We will be using them in the future.

database/database.go
ro-tex: Fixed.

database/database.go
ro-tex: I'll just remove them for the moment. YAGNI.

database/database.go
ro-tex: I'll unexport them and move them to main in a F/U.

api/api.go
ro-tex: F/U: https://github.com/NebulousLabs/skynet-accounts/issues/5

user/user.go
ro-tex: F/U: https://github.com/NebulousLabs/skynet-accounts/issues/5

database/database.go
ro-tex: F/U: https://github.com/NebulousLabs/skynet-accounts/issues/5

Makefile
ro-tex: Added.

I used Sia's file + exclusion for `composites` errors for `govet` because we need those composites in order to talk to Mongo.

database/database.go
ro-tex: F/U: #5

PR #6, opened 2020-11-23
Implement follow-ups on the initial commit

### LINKED ISSUE #5: F/U on Initial Implementation
## Comments to address

- [x] Should `User` have a mutex? https://github.com/NebulousLabs/skynet-accounts/pull/1#discussion_r524345080
- [x] Env var names moved to `main`. https://github.com/NebulousLabs/skynet-accounts/pull/1#discussion_r526917603
- [x] Should `user` live under `database`? https://github.com/NebulousLabs/skynet-accounts/pull/1#discussion_r526920973
- [x] DB should be passed in. https://github.com/NebulousLabs/skynet-accounts/pull/1#discussion_r526922828

ro-tex:
Note: I kept losing my todo notes, so I just committed them in this one. I need to address them soon anyway, so it should be fine for us to have some TODOs in the code while we're iterating quickly.

This MR does not cover the two `User`-related items in #5, we need to finalize our consensus on those first. Please feel free to chime in.

Closes #5

### DIFF
--- a/database/database.go
+++ b/database/database.go
@@ -5,7 +5,6 @@ import (
 	"context"
 	"fmt"
 	"net/url"
-	"os"
 
 	"github.com/NebulousLabs/skynet-accounts/build"
 	"github.com/NebulousLabs/skynet-accounts/user"
@@ -24,15 +23,6 @@ TODO
 */
 
 var (
-	// envDBHost holds the name of the environment variable for DB host.
-	envDBHost = "SKYNET_DB_HOST"
-	// envDBPort holds the name of the environment variable for DB port.
-	envDBPort = "SKYNET_DB_PORT"
-	// envDBUser holds the name of the environment variable for DB username.
-	envDBUser = "SKYNET_DB_USER" // #nosec
-	// envDBPass holds the name of the environment variable for DB password.
-	envDBPass = "SKYNET_DB_PASS" // #nosec
-
 	// mongoCompressors defines the compressors we are going to use for the
 	// connection to MongoDB
 	mongoCompressors = "zstd,zlib,snappy"
@@ -64,24 +54,26 @@ var (
 	ErrGeneralInternalFailure = errors.New("general internal failure")
 )
 
-// DB represents a MongoDB database connection.
-type DB struct {
-	staticDB    *mongo.Database
-	staticUsers *mongo.Collection
-}
-
-// New returns a new DB connection based on the environment variables.
-func New(ctx context.Context) (*DB, error) {
-	opts, err := connectionOptionsFromEnv()
-	if err != nil {
-		return nil, errors.AddContext(err, "failed to get all necessary connection parameters")
+type (
+	// DB represents a MongoDB database connection.
+	DB struct {
+		staticDB    *mongo.Database
+		staticUsers *mongo.Collection
 	}
-	return NewCustom(ctx, opts[envDBUser], opts[envDBPass], opts[envDBHost], opts[envDBPort], dbName)
-}
 
-// NewCustom returns a new DB connection based on the passed parameters.
-func NewCustom(ctx context.Context, user, pass, host, port, dbname string) (*DB, error) {
-	connStr := connectionString(user, pass, host, port)
+	// DBCredentials is a helper struct that binds together all values needed for
+	// establishing a DB connection.
+	DBCredentials struct {
+		User     string
+		Password string
+		Host     string
+		Port     string
+	}
+)
+
+// New returns a new DB connection based on the passed parameters.
+func New(ctx context.Context, creds DBCredentials) (*DB, error) {
+	connStr := connectionString(creds)
 	c, err := mongo.NewClient(options.Client().ApplyURI(connStr))
 	if err != nil {
 		return nil, errors.AddContext(err, "failed to create a new DB client")
@@ -90,7 +82,7 @@ func NewCustom(ctx context.Context, user, pass, host, port, dbname string) (*DB,
 	if err != nil {
 		return nil, errors.AddContext(err, "failed to connect to DB")
 	}
-	database := c.Database(dbname)
+	database := c.Database(dbName)
 	users := database.Collection(dbUsersCollection)
 	db := &DB{
 		staticDB:    database,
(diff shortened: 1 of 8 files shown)


main.go
```diff
@@ -1,22 +1,62 @@
...
+	"github.com/NebulousLabs/skynet-accounts/database"
+)
+
+var (
+	// envDBHost holds the name of the environment variable for DB host.
+	envDBHost = "SKYNET_DB_HOST"
+	// envDBPort holds the name of the environment variable for DB port.
+	envDBPort = "SKYNET_DB_PORT"
+	// envDBUser holds the name of the environment variable for DB username.
+	envDBUser = "SKYNET_DB_USER" // #nosec
```
m-cat, line 22: q: Why disable gosec here? 

nit: Can you add the error, e.g. `#nosec GXXX`?

test/database_test.go
```diff
@@ -204,13 +195,13 @@ func TestDatabase_UserDelete(t *testing.T) {
...
-	if err := errors.Compose(e1, e2, e3, e4); err != nil {
-		panic(err)
+// DBTestCredentials sets the environment variables to what we have defined in Makefile.
+func DBTestCredentials() database.DBCredentials {
+	return database.DBCredentials{
+		User:     "admin",
+		Password: "ivolocalpass",
+		Host:     "localhost",
+		Port:     "37017",
+		//Port:     "27017", // DEBUG
```
m-cat, line 205: You left in this comment

m-cat (dismissed):

main.go
ro-tex: A: It was suspecting that I'm hard-coding the credentials to the DB. This is an error that we're supposed to verify and disable if it's a false alarm.

> nit: Can you add the error, e.g. #nosec GXXX?

Good idea!

m-cat: Cool, with the error specified we only disable the one that we explicitly expect 🙂

m-cat (dismissed):

m-cat (approved):

m-cat (approved):

test/database_test.go
```diff
@@ -204,13 +193,12 @@ func TestDatabase_UserDelete(t *testing.T) {
...
-	e2 := os.Setenv("SKYNET_DB_PORT", "37017")
-	e3 := os.Setenv("SKYNET_DB_USER", "admin")
-	e4 := os.Setenv("SKYNET_DB_PASS", "ivolocalpass")
-	if err := errors.Compose(e1, e2, e3, e4); err != nil {
-		panic(err)
+// DBTestCredentials sets the environment variables to what we have defined in Makefile.
+func DBTestCredentials() database.DBCredentials {
+	return database.DBCredentials{
+		User:     "admin",
+		Password: "ivolocalpass",
```
peterjanbrone, line 200: probably best to not commit our names but rather use something generic

test/database_test.go
```diff
@@ -93,25 +86,23 @@ func TestDatabase_UserByID(t *testing.T) {
...
 		t.Fatal(err)
 	}
 
-	u := &user.User{
+	username := t.Name()
+	u := &database.User{
 		FirstName: username,
 		LastName:  "Pratchett",
-		Email:     (user.Email)(username + "@pratchett.com"),
+		Email:     (database.Email)(username + "@pratchett.com"),
```
peterjanbrone, line 99: are you referring to Terry Pratchett?

test/database_test.go
```diff
@@ -30,15 +24,16 @@ func TestDatabase_UserByEmail(t *testing.T) {
...
 	}
 	if err = db.UserCreate(nil, u); err != nil {
 		t.Fatal(err)
 	}
-	defer func(user *user.User) {
+	defer func(user *database.User) {
```
peterjanbrone, line 36: this is fine but I guess this is usually done in `teardown` or global init code like `beforeAll`

main.go
```diff
@@ -1,22 +1,62 @@
...
+	envDBPort = "SKYNET_DB_PORT"
+	// envDBUser holds the name of the environment variable for DB username.
+	envDBUser = "SKYNET_DB_USER"
+	// envDBPass holds the name of the environment variable for DB password.
+	envDBPass = "SKYNET_DB_PASS" // #nosec G101: Potential hardcoded credentials
 )
 
+// dbConnFromEnv creates a new DB connection based on credentials found in the
+// environment variables.
+func dbConnFromEnv(ctx context.Context) (*database.DB, error) {
```
peterjanbrone, line 29: (nit) I think having a `loadCredentials() database.DBCredentials` makes more sense and is easier to unit test separately, `dbConnFromEnv` looks a bit weird imo

database/database.go
```diff
@@ -263,34 +254,20 @@ func (db *DB) managedUsersByField(ctx context.Context, fieldName, fieldValue str
...
 		"mongodb://%s:%s@%s:%s/?compressors=%s&readPreference=%s&w=%s&wtimeoutMS=%s",
-		url.QueryEscape(user),
-		url.QueryEscape(pass),
-		host,
-		port,
+		url.QueryEscape(creds.User),
+		url.QueryEscape(creds.Password),
+		creds.Host,
+		creds.Port,
 		mongoCompressors,
```
peterjanbrone, line 271: I had to google `mongoCompressors` - never encountered those before - do we need those?

peterjanbrone (approved):
I missed the first PR though so reviewing this F/U was not very valuable really, it's really minor changes and renames. Overall though it definitely LGTM 👍

database/database.go
ro-tex: We don't strictly need it but since it allows compression of the communication between the replica set nodes it seemed like a good idea. Is there a reason not to have it? I don't have a strong opinion on this one and we can always change it.

main.go
ro-tex: Done in #8 .

test/database_test.go
ro-tex: TBH, I can just remove it. The email is unique, so it shouldn't be a problem. And the DB we're using for this doesn't have any persistence, so it won't carry over between runs.

test/database_test.go
ro-tex: The one and only! :)

test/database_test.go
ro-tex: Will change in #8.

ChrisSchinnerl (approved):

PR #8, opened 2020-11-24
User registration

### LINKED ISSUE #4: User registration and authentication
# Overview

- users will authenticate with email and password
- email needs to be verified
- requests will be authorized with JWT tokens with limited duration (24h)

# Details

### Email

- email verification required before an account is enabled
- send confirmation email endpoint
- re-send confirmation email endpoint (can be same as above)
- password reset email with limited duration (256bit token + 4 hours duration?)
  * make it possible for users to notify us that they haven't requested a password reset

### Passwords

- passwords need a sane security policy
- sane encryption (`bcrypt`+salt or a complete solution)

### Authorization

API requests will be authorized by a JWT token signed by a key shared among all portals.

In the future we can improve that by having each portal have an individual key and sign so that all other portals can verify. This will allow us to isolate compromised keys.

The JWT token will have a 24 hour lifetime for the moment, so we can quickly adapt our strategy, if something's amiss. In the future this duration will likely be increased and JWT tokens without an expiration date will be introduced (mainly for use by robots).

### SkyID

SkyID support will definitely come in the future but not in this ticket.

OAuth support *might* come but it's not planned right now.

ro-tex:
# Overview

Implement the entire login logic here, as well as token creation/verification/expiration.

# Description

## Flow

- Registration:
	- [x] validate email
	- [x] validate password (!empty)
	- [x] proper password hashing
	- [x] mark as unconfirmed (status/tier?)
	- [x] save to DB
	- [ ] ~~generate confirmation token~~
	- [ ] ~~email confirmation token with an expiration~~
- Confirmation:
	- [ ] ~~verify confirmation token~~
	- [ ] ~~update the User to `confirmed`~~
- Login:
	- [x] validate email/password/status(confirmed)
	- [x] issue a JWT token
- User details:
	- [x] get user info (verify it's the same user)
	- [x] update user info? (not email - that will be in a follow up because we need a more complicated confirmation)
	- [x] change password

## Details

This MR introduces the use of `.env` file which exports a number of environment variables. It must be noted that these values are only registered if there is no existing environment variable with that name, i.e. existing env vars take precedence over this file.

One **potential issue** with this is that the JWT tokens we're issuing are signed with the value of the `ACCESS_SECRET` env var. We should consider what's the most secure way to store this critical value:
* environment variable only
* .env file
* DB

## Follow-ups

The generation and sending of an email confirmation was left for a follow up. This should follow shortly after this MR is merged and should be merged before we release this to the public (even for alpha).

Closes #4


api/handlers.go
```diff
@@ -112,23 +140,106 @@ func (api *API) userHandlerPUT(w http.ResponseWriter, req *http.Request, ps http
...
+	}
+	token, err := IssueToken(u)
+	if err != nil {
+		fmt.Println(err)
+		// TODO WriteError doesn't set the response's error message properly. Or Postman doesn't read it properly?
+		WriteError(w, err, http.StatusUnprocessableEntity)
+		return
+	}
+	w.WriteHeader(http.StatusOK)
+	WriteJSON(w, token)
```
ro-tex, line 237: Something @kwypchlo brought up today was that I can set the user's cookie right here instead of relying on the FE to do that for me. I'll do that in a follow up.

Makefile
```diff
@@ -95,12 +109,12 @@ test:
...
+	GORACE='$(racevars)' go test -race --coverprofile='./cover/cover.out' -v -failfast -tags='testing debug netgo' -timeout=30s $(pkgs) -run=. -count=$(count)
 
 # test-int always returns a zero exit value! Only use it manually!
-test-int: clean fmt vet lint-ci start-mongo
-	@mkdir -p cover
-	GORACE='$(racevars)' go test -race --coverprofile='./cover/cover.out' -v -tags='testing debug netgo' -timeout=3600s $(integration-pkgs) -run=. -count=$(count) ; \
+test-int: clean fmt vet lint-ci test-long start-mongo start-mailslurp
+	GORACE='$(racevars)' go test -race -v -tags='testing debug netgo' -timeout=300s $(integration-pkgs) -run=. -count=$(count) ; \
 	make stop-mongo
+	#make stop-mailslurp
```
m-cat, line 118: Comment

Makefile
```diff
@@ -65,12 +65,26 @@ start-mongo:
...
+     -e MONGO_INITDB_ROOT_PASSWORD=aO4tV5tC1oU3oQ7u \
      mongo
 
 stop-mongo:
 	docker stop skynet-accounts-mongo-test-db
 
+# start-mailslurp starts a fake mail server listening on port 2500.
+# We use that server in our integration tests.
+start-mailslurp:
+	-docker stop mailslurp
```
m-cat, line 77: Should it be

```suggestion
	docker stop mailslurp
```

?

m-cat: Maybe use the `stop-mailslurp` target, in case we ever add e.g. some cleanup to stopping

api/jwt_test.go
```diff
@@ -0,0 +1,34 @@
...
+
+	"go.mongodb.org/mongo-driver/bson/primitive"
+
+	"github.com/NebulousLabs/skynet-accounts/database"
+)
+
+// TestToken ensures that IssueToken() and ValidateToken() function properly.
+// TODO This test is too basic and needs to be expanded.
+func TestToken(t *testing.T) {
+	id, err := primitive.ObjectIDFromHex("5fbd557f0518ba7ae9267e3b")
```
m-cat, line 14: We should hard-code an expected ID to match against, seems important to make sure it shouldn't ever change.

api/routes.go
```diff
@@ -1,14 +1,36 @@
...
 	// TODO Rate-limit these from the start to prevent brute force and other forms of abuse.
 	api.staticRouter.POST("/login", api.userLoginHandler)
-	api.staticRouter.POST("/user", api.userHandlerPOST)
-	api.staticRouter.GET("/user/:id", api.userHandlerGET)
-	api.staticRouter.PUT("/user/:id", api.userHandlerPUT)
-	api.staticRouter.POST("/user/:id/password", api.userChangePasswordHandler)
-	api.staticRouter.POST("/user/:id/password/reset/request", api.userPasswordResetRequestHandler)
-	api.staticRouter.POST("/user/:id/password/reset/verify", api.userPasswordResetCompleteHandler)
-	api.staticRouter.POST("/user/:id/password/reset/complete", api.userPasswordResetCompleteHandler)
+	api.staticRouter.POST("/user", (api.userHandlerPOST))
```
m-cat, line 15:
```suggestion
	api.staticRouter.POST("/user", api.userHandlerPOST)
```

Makefile
ro-tex: That's a good idea, I'll use the target. But the dash should stay - failing to clean up properly shouldn't fail the tests.

api/jwt_test.go
ro-tex: I'm not sure what you mean. Add a record with a known id to the DB and look for it in the tests? The tests will be run against a brand new, empty DB, so we'd have to insert it first. And this execution path is already tested when we test the `add user` functionality.

m-cat: Thought we were generating IDs, read this way too fast

Makefile
m-cat: Oh, I had never seen that dash in Makefiles. Just looked it up. Cool

ro-tex: Yeah, I didn't know about it, either. Chris told me.

API.md
```diff
@@ -0,0 +1,97 @@
+# Basic API Guide
+
+## Login
+
+* Endpoint: `/login`
+* Verb: `POST`
+* Requires existing session: `false`
+* GET params:
```
m-cat, line 8: nit: Can we remove the blank `GET params` here and below?

API.md
```diff
@@ -0,0 +1,97 @@
...
+* Requires existing session: `false`
+* GET params: 
+* POST params:
+    - `email`: string
+    - `password`: string
+* Returns: 200 JSON string (JWT token)
+```json
+"eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1c2VyX2lkIjoiNWZjNTFkOWNhMGYzNTFmYjYyMjkwMDlhIiwidGllciI6MSwiZXhwIjoxNjA2ODQwNjAxLCJpYXQiOjE2MDY3NTQyMDEsImlzcyI6IlNreW5ldCBBY2NvdW50cyJ9.0dhkOMHFBcoZSSBDrVBca5SOSeU3zMEOLVfNQWf_cOI"
+```
+also 400, 401, 422, 500
```
m-cat, line 16: Should we document what these error codes mean?

api/handlers.go
```diff
@@ -18,11 +18,25 @@ const (
 	MaxMultipartMem = 64_000_000
 )
 
+var (
+	// ErrAccountUnconfirmed is returned when a suer with an unconfirmed account
```
m-cat, line 22:
```suggestion
	// ErrAccountUnconfirmed is returned when a user with an unconfirmed account
```

api/handlers.go
```diff
@@ -40,17 +54,28 @@ func (api *API) userHandlerPOST(w http.ResponseWriter, req *http.Request, ps htt
 		WriteError(w, errors.New("Failed to parse multipart parameters."), http.StatusBadRequest)
 		return
 	}
-	email := (user.Email)(req.PostFormValue("email"))
-	if !email.Validate() {
-		WriteError(w, user.ErrInvalidEmail, http.StatusBadRequest)
+	email, err := database.NewEmail(req.PostFormValue("email"))
+	if err != nil {
+		WriteError(w, database.ErrInvalidEmail, http.StatusBadRequest)
```
m-cat, line 59: I think

```suggestion
		WriteError(w, err, http.StatusBadRequest)
```

is better, in case we ever add contexts in `NewEmail`.

api/handlers.go
```diff
@@ -112,23 +140,105 @@ func (api *API) userHandlerPUT(w http.ResponseWriter, req *http.Request, ps http
...
+			return
+		}
+	}
+	err = req.ParseMultipartForm(MaxMultipartMem)
+	if err != nil {
+		WriteError(w, errors.New("Failed to parse multipart parameters."), http.StatusBadRequest)
+		return
+	}
+	// Validate that the given old password is correct.
+	if pw := req.PostFormValue("oldPassword"); pw != "" {
```
m-cat, line 170: Looks like if `pw == ""` it skips the validation, so someone can just omit it and override a password without knowing it

m-cat (changes requested):
Reviewed everything so far except `jwt.go`

API.md
ro-tex: Yes, we should definitely do that but maybe a little later. We'll use Slate to document the API, just like in `siad`. I created this doc, so Karol could start building the front end. At first i thought it shouldn't even be under source control because it's temporary but then I decided to include it because we can just rework it into the slate format (I still need to look that up).

All in all, we don't need to worry about this doc too much for the moment.

API.md
ro-tex: Done.

api/handlers.go
ro-tex: Sure, makes sense. It will also make it more like the other endpoints.

I'm a little divided on the idea of context here because we're sending this data to the user which might potentially be an attacker, so it might make sense to not disclose which execution path they are hitting. But this might come later, no need to focus on defense prematurely.

api/handlers.go
ro-tex: Ha! Brilliant! Exactly the opposite behaviour than what we want! Thanks for that catch. :)

m-cat (approved):
Looks good!

api/handlers.go
```diff
@@ -60,32 +81,28 @@ func (api *API) userHandlerPOST(w http.ResponseWriter, req *http.Request, ps htt
...
+	var u *database.User
+	// Fetch the user by their id.
+	u, err = api.staticDB.UserByID(req.Context(), uid)
+	if err != nil {
+		// This is a Bad Request and not an Internal Server Error because
+		// the client has supplied an invalid user id.
+		WriteError(w, errors.AddContext(err, "failed to fetch user"), http.StatusBadRequest)
+		return
+	}
+	err = req.ParseMultipartForm(MaxMultipartMem)
```
ChrisSchinnerl, line 100: Why do we need this? `PostFormValue` should implicitly call that iff `r.PostForm == nil`.
It will also call it with about half the memory of `MaxMultipartMem` which seems reasonable.

api/handlers.go
```diff
@@ -111,24 +129,102 @@ func (api *API) userHandlerPUT(w http.ResponseWriter, req *http.Request, ps http
...
```
ChrisSchinnerl, line 148: same

api/handlers.go
```diff
@@ -111,24 +129,102 @@ func (api *API) userHandlerPUT(w http.ResponseWriter, req *http.Request, ps http
...
 // userLoginHandler starts a new session for a user.
-func (api *API) userLoginHandler(w http.ResponseWriter, req *http.Request, ps httprouter.Params) {
-	// TODO Implement
-	fmt.Println("User login.")
-	//req.ParseForm()
-	//req.Form.Set("async", "true")
-	//api.renterDownloadHandler(w, req, ps)
-	WriteJSON(w, struct{ msg string }{"Not implemented."})
+func (api *API) userLoginHandler(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
+	if err := req.ParseMultipartForm(MaxMultipartMem); err != nil {
```
ChrisSchinnerl, line 182: same

api/handlers.go
```diff
@@ -35,22 +45,33 @@ func (api *API) userHandlerGET(w http.ResponseWriter, req *http.Request, ps http
 }
 
 // userHandlerPOST creates a new user.
-func (api *API) userHandlerPOST(w http.ResponseWriter, req *http.Request, ps httprouter.Params) {
+func (api *API) userHandlerPOST(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
 	if err := req.ParseMultipartForm(MaxMultipartMem); err != nil {
```
ChrisSchinnerl, line 49: same

api/handlers.go
```diff
@@ -111,24 +129,102 @@ func (api *API) userHandlerPUT(w http.ResponseWriter, req *http.Request, ps http
...
+		return
+	}
+	// Set the new password.
+	err = u.SetPassword(newPass)
+	if err != nil {
+		WriteError(w, err, http.StatusInternalServerError)
+		return
+	}
+	// Persist the change.
+	err = api.staticDB.UserUpdatePassword(req.Context(), u)
```
ChrisSchinnerl, line 172: I think changing the password should be a single method on the database. 
We should never need to fetch a user from a databse at all actually. Unless we explicitly want to return user information.

e.g. if you want to update a user's pw you should call `staticDB.ChangePassword(id, oldPassword, newPassword)`
Then you would (within a single db txn) fetch the user, verify the password and set the password. 

Otherwise you end up with race conditions where 2 threads fetch a user, each one sets a different new password, both verify the old password and then they call update and whoever calls update last wins.

ChrisSchinnerl: tldr; 

Updates to the database should be atomic and that should be reflected in its interface.

API.md
```diff
@@ -0,0 +1,91 @@
...
+
+### Lookup own user data
+
+* Endpoint: `/user/:id`
+* Verb: `GET`
+* Requires existing session: `true`
+* Returns: 200 JSON object
+```json
+{
+    "ID": "5fbe7dca17defb7793e6b7d0",
```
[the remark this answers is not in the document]
ro-tex: Done.

API.md
```diff
@@ -0,0 +1,91 @@
...
+```
+also 400, 401, 500
+
+### Update own user data
+
+* Endpoint: `/user/:id`
+* Verb: `PUT`
+* Requires existing session: `true`
+* POST params:
+    - `id`: string, identifier, won't be changed
```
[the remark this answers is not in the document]
ro-tex: Right, this is already done but the docs are stale. Updated.

Makefile
```diff
@@ -65,12 +65,26 @@ start-mongo:
      --name skynet-accounts-mongo-test-db \
      -p 127.0.0.1:37017:27017 \
      -e MONGO_INITDB_ROOT_USERNAME=admin \
-     -e MONGO_INITDB_ROOT_PASSWORD=ivolocalpass \
+     -e MONGO_INITDB_ROOT_PASSWORD=aO4tV5tC1oU3oQ7u \
```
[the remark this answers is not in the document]
ro-tex: We start a fresh Mongo container without any persistence and we run a set of operations against it. After that the container is discarded together with all of its data. In order to talk to Mongo we need a username and  a password and we can set those via these parameters. The username and password can be completely arbitrary and don't need to be secret.

api/api.go
```diff
@@ -28,23 +37,24 @@ type API struct {
 	staticRouter *httprouter.Router
 }
 
+// ctxValue is a helper type which makes it safe to register values in the
+// context. If we don't use a custom unexported type it's possible for others
+// to get our value.
+type ctxValue string
```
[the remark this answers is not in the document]
ro-tex: I haven't really played around with this but it sounds like a fun exploration task.
For the moment I changed the docstring to say `easy` instead of `possible`.

api/routes.go
```diff
@@ -1,14 +1,36 @@
...
+func Validate(h httprouter.Handle) httprouter.Handle {
+	return func(w http.ResponseWriter, req *http.Request, ps httprouter.Params) {
+		token, err := ValidateToken(extractToken(req))
+		if err != nil {
+			WriteError(w, errors.New("Unauthorized"), http.StatusUnauthorized)
+			return
+		}
+		// Embed the verified token in the context of the request.
+		ctx := context.WithValue(req.Context(), ctxValue("token"), token)
+		h(w, req.Clone(ctx), ps)
```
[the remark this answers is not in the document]
ro-tex: Nice, I didn't know this existed.

api/handlers.go
```diff
@@ -60,32 +81,28 @@ func (api *API) userHandlerPOST(w http.ResponseWriter, req *http.Request, ps htt
...
-		email := (user.Email)(req.PostFormValue("email"))
-		u, err = api.staticDB.UserByEmail(req.Context(), email)
-		if err != nil {
-			WriteError(w, errors.AddContext(err, "failed to fetch user"), http.StatusBadRequest)
-			return
-		}
+
+	var u *database.User
+	// Fetch the user by their id.
+	u, err = api.staticDB.UserByID(req.Context(), uid)
```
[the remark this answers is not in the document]
ro-tex: Leftover from old logic.

api/handlers.go
ro-tex: Oh, you're right. I think I wasn't getting my parameters for some reason and I read that this is what I needed to do. And then it worked, so I kept it. Probably it was something completely unrelated like a bad Postman setup.

Removed.

api/handlers.go
ro-tex: Hm, this should also extend to "update user data" then. I'll update it.

ro-tex: I'm actually not sure about "update user data" - that might have more complicated logic that definitely doesn't fit into a single DB txn.

For the moment I would hold off on adding transactions, so we can get this off the ground. I've added `TODO`.

ro-tex: > We should never need to fetch a user from a databse at all actually. Unless we explicitly want to return user information.

We need to fetch the user in order to get their personal salt. Otherwise we cannot verify that the password is correct. I don't think we can do `bcrypt` stuff within Mongo requests.

> Otherwise you end up with race conditions where 2 threads fetch a user, each one sets a different new password, both verify the old password and then they call update and whoever calls update last wins.

We don't have race conditions because we don't have threads for the moment. That's why we removed the mutex from the user struct. We have *some form* of multi-threading in the form of multiple requests that work with the same user record, though.

The only way to avoid the race conditions is to lock/unlock the record in the database or to use transactions. The second approach is what we should do, IMO. But I would argue (as I did above) that we can skip that for the moment and do it in a follow up. If you think that this is required then I'll add transactions.

api/handlers.go
```diff
@@ -35,22 +39,29 @@ func (api *API) userHandlerGET(w http.ResponseWriter, req *http.Request, ps http
...
 		Email:     email,
+		Tier:      database.TierFree,
+	}
+	err = u.SetPassword(pw)
+	if err != nil {
+		WriteError(w, errors.AddContext(err, "failed to set password"), http.StatusInternalServerError)
+		return
 	}
-	err := api.staticDB.UserCreate(req.Context(), u)
+	err = api.staticDB.UserCreate(req.Context(), u)
```
ChrisSchinnerl, line 64: F/U I'd change this to `UserCreate(ctx, firstname, lastname, email, tier, password)` for safety and convenience. It shouldn't be necessary to manually create users outside of the database package.

ChrisSchinnerl (approved):

m-cat (approved):

api/handlers.go
ro-tex: F/U: #12

README.md
```diff
@@ -2,6 +2,22 @@
...
+
+All local secrets are loaded from a `.env` file in the root directory of the project.
+
+Those are (example values):
+```.env
+SKYNET_DB_HOST="localhost"
+SKYNET_DB_PORT="27017"
+SKYNET_DB_USER="username"
+SKYNET_DB_PASS="password"
+SKYNET_PEPPER="some random string for pepper"
```
peterjanbrone, line 17: sorry but what is "pepper".. if that's some joke on a salt we should definitely get rid of it

ro-tex: Oh, you and your unrelenting hatred towards anything fun!

https://en.wikipedia.org/wiki/Pepper_(cryptography)

PR #13, opened 2020-12-02
Move atomic operations to DB

### LINKED ISSUE #12: F/U I'd change this to `UserCreate(ctx, firstname, lastname, email, tier, password)` for safety and convenience. It shouldn't be necessary to manually create users outside of the database package.
F/U I'd change this to `UserCreate(ctx, firstname, lastname, email, tier, password)` for safety and convenience. It shouldn't be necessary to manually create users outside of the database package.

_Originally posted by @ChrisSchinnerl in https://github.com/NebulousLabs/skynet-accounts/pull/8#discussion_r534238867_

ro-tex:
We want to have all operations that should be atomic under the `DB` struct, so we can easily wrap them in transactions later.

Based on #8 (only review after that one is merged).

Closes #12


ro-tex:
Diffs are bugged, will open a new PR.

PR #14, opened 2020-12-03
Set and read secure cookies.

### LINKED ISSUE #11: Set the user's cookie on login
Something @kwypchlo brought up today was that I can set the user's cookie right here instead of relying on the FE to do that for me. I'll do that in a follow up.

_Originally posted by @ro-tex in https://github.com/NebulousLabs/skynet-accounts/pull/8#r532747939_

ro-tex:
Based on #8 (only review after that one is merged).

Closes #11


ro-tex:
Diff is bugged, will open a new PR.

PR #15, opened 2020-12-04
Move atomic operations to DB

### LINKED ISSUE #12: F/U I'd change this to `UserCreate(ctx, firstname, lastname, email, tier, password)` for safety and convenience. It shouldn't be necessary to manually create users outside of the database package.

ro-tex:
We want to have all operations that should be atomic under the `DB` struct, so we can easily wrap them in transactions later.

Closes #12

### DIFF
--- a/database/database_test.go
+++ b/database/database_test.go
@@ -30,7 +30,7 @@ func TestDB_passwordHashAndSalt(t *testing.T) {
 	}
 
 	// FAILURE CASE:
-	// Ensure failing to set a password doesn't affect the user'salt salt.
+	// Ensure failing to set a password doesn't affect the user's salt.
 	salt = u.Salt
 	db.staticDep = &test.DependencyHashPassword{}
 	_, _, err = db.passwordHashAndSalt("some_new_pass")
--- a/database/database_test.go
+++ b/database/database_test.go
@@ -35,6 +35,6 @@ func TestDB_passwordHashAndSalt(t *testing.T) {
 	db.staticDep = &test.DependencyHashPassword{}
 	_, _, err = db.passwordHashAndSalt("some_new_pass")
 	if err == nil || !strings.Contains(err.Error(), "DependencyHashPassword") {
-		t.Fatalf("expected to fail with  %salt but got %v\n", "DependencyHashPassword", err)
+		t.Fatalf("expected to fail with %s but got %v\n", "DependencyHashPassword", err)
 	}
 }
--- a/database/database_test.go
+++ b/database/database_test.go
@@ -1,0 +1,40 @@
+package database
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/NebulousLabs/skynet-accounts/test"
+	"gitlab.com/NebulousLabs/fastrand"
+)
+
+// TestDB_passwordHashAndSalt ensures passwordHashAndSalt() works as expected.
+func TestDB_passwordHashAndSalt(t *testing.T) {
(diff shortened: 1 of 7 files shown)


database/database_test.go
```diff
@@ -0,0 +1,40 @@
...
+		Email:    "foo@bar.baz",
+		Password: pw,
+		Salt:     salt,
+	}
+	if err = u.VerifyPassword(password); err != nil {
+		t.Fatal("failed to verify password", err)
+	}
+
+	// FAILURE CASE:
+	// Ensure failing to set a password doesn't affect the user'salt salt.
```
m-cat, line 33:
```suggestion
	// Ensure failing to set a password doesn't affect the user's salt.
```

database/database_test.go
```diff
@@ -0,0 +1,40 @@
...
+	salt = u.Salt
+	db.staticDep = &test.DependencyHashPassword{}
+	_, _, err = db.passwordHashAndSalt("some_new_pass")
+	if err == nil || !strings.Contains(err.Error(), "DependencyHashPassword") {
+		t.Fatalf("expected to fail with  %salt but got %v\n", "DependencyHashPassword", err)
```
m-cat, line 38:
```suggestion
		t.Fatalf("expected to fail with %s but got %v\n", "DependencyHashPassword", err)
```

database/user_test.go
```diff
@@ -65,7 +62,11 @@ func TestEmail_Validate(t *testing.T) {
 func TestUser_saltAndPepper(t *testing.T) {
 	initEnv()
 	salt := []byte("this is a salt")
-	pepper := []byte("this is some pepper")
+	pepper, ok := os.LookupEnv(envPepper)
+	if !ok {
+		os.Setenv(envPepper, "this is a salt")
```
m-cat, line 67: Why set this env var here? Also, isn't `pepper` empty?

m-cat (dismissed):

database/database_test.go
ro-tex: GoLand's rename feature is annoyingly buggy...

database/user_test.go
ro-tex: The idea is that the environment variable might or might not be set.

If it's set - I want to use it as it is.

If it's not set I need to set it (because `saltAndPepper` depends on it) and then unset it again, in order to clean up the environment.

In theory, the pepper env var should not be empty because of the call to `initEnv` which sets it.

ro-tex: I might be more cautious than I should be and I can just set the env var if it's not set and be done with it. If you think that would be god enough, I'll just make that change.

m-cat: I think I would just set the expected `pepper`, we don't need to set/unset the env var because we set it in `initEnv`. We also do an env var lookup for pepper in `saltAndPepper()`, so there is no point in doing it here. I would refactor out the pepper from `initEnv` and use it as the input pepper on line 78. Currently the `pepper` on line 78 would be empty if this lookup check failed

m-cat (approved):

ro-tex:
The changes in this PR are no longer needed, considering the move to Kratos. Closing.

PR #16, opened 2020-12-04
 Set and read secure cookies

### LINKED ISSUE #11: Set the user's cookie on login

ro-tex:
Closes #11


api/cookie.go
```diff
@@ -0,0 +1,73 @@
...
+
+const (
+	// cookieName is the name of the cookie where we store the user's JWT token.
+	cookieName = "SkynetJWTCookie"
+	// cookieValidity specifies for how long the cookie is valid. In seconds.
+	cookieValidity = 604800 // one week
+
+	// envCookieDomain holds the name of the environment variable for the
+	// domain name of the portal
+	envCookieDomain = "COOKIE_COMAIN"
```
m-cat, line 19:
```suggestion
	envCookieDomain = "COOKIE_DOMAIN"
```

m-cat: Why have env vars to store the names of other env vars? I feel like we have a lot of env vars and it's getting confusing

m-cat (changes requested):
Just 1 comment

api/cookie.go
ro-tex: Ha ha ha, Clippy!

m-cat (approved):

ro-tex:
The changes in this PR are no longer needed, considering the move to Kratos. Closing.

PR #17, opened 2020-12-10
Kratos

ro-tex:
# Overview

This PR adds the necessary endpoints and supporting code for an initial accounts implementation in tandem with ORY/Kratos and ORY/Oathkeeper.

# Details

This PR contains the changes in the former "cookies" and "user_create" branches with the only goal to clean up all the account-management aspects and keep the rest of the code, as minimal as it may be.

We should probably squish this on merge - most of the commits here are just noise.


API.md
```diff
@@ -1,86 +1,57 @@
...
-### Lookup own user data
+This request combines the "get user data" and "create user" requests - if the
+users exists in the DB, their data will be returned. If they don't exist in the
+DB, an account will be created on the Free tier.
 
 * Endpoint: `/user`
 * Verb: `GET`
 * Requires existing session: `true`
-* Returns: 200 JSON object
+* Returns:
```
m-cat, line 14: FYI, the formatting seems off in a couple places:

Screen Shot 2020-12-10 at 14 41 50

api/jwt.go
```diff
@@ -3,77 +3,227 @@ package api
...
-		Tier   int    `json:"tier"`
+var (
+	// oathkeeperPubKeys is the public RS key exposed by Oathkeeper for JWT
+	// validation. It's available at oathkeeperPubKeyUrl.
+	oathkeeperPubKeys *jwk.Set = nil
 
-		jwt.StandardClaims
-	}
+	// oathkeeperPubKeyUrl is the URL on which we can find the public key.
+	oathkeeperPubKeyUrl = "http://oathkeeper:4456/.well-known/jwks.json"
```
m-cat, line 20: niggle: I have some lint enabled that complains here

```suggestion
	oathkeeperPubKeyURL = "http://oathkeeper:4456/.well-known/jwks.json"
```

database/database.go
```diff
@@ -42,9 +38,9 @@ var (
 
 	// ErrUserNotFound is returned when we can't find the user in question.
 	ErrUserNotFound = errors.New("user not found")
-	// ErrEmailAlreadyUsed is returned when we try to use an email to either
-	// create or update a user and another user already uses this email.
-	ErrEmailAlreadyUsed = errors.New("email already in use by another user")
+	// ErrUserAlreadyExists is returned when we try to use a sub to create a
+	// user and a user already exists with this sub. This should never happen.
+	ErrUserAlreadyExists = errors.New("sub already belongs to an existing user")
```
m-cat, line 43: Will readers of this error know what "sub" means?

m-cat (dismissed):
Looks good from what I can tell. I'm not sure I totally understand how Kratos fits in here, I would add a section on that and on JWT tokens in API.md.

API.md
```diff
@@ -1,86 +1,57 @@
...
-    "lastName": "Doe",
-    "email": "john@doe.com",
-    "tier": 0
-}
-```
-also 400 and 500
+### Get
 
-### Lookup own user data
+This request combines the "get user data" and "create user" requests - if the
```
peterjanbrone, line 7: This is an extremely weird API design. `GET` requests don't go and create new entities in the database. 

I'm also confused as to how this would work, don't we need some user information to create a `user` object with?  Like, an email at least? 

How is this request authenticated, it says that it requires a session, how can one acquire a session if he's not a user yet.

API.md
```diff
@@ -1,86 +1,57 @@
...
-    "email": "john@doe.com",
     "tier": 1
 }
 ```
-also 400, 401, 500
+  - 404 (when there is no such user and we fail to create it)
+  - 500 (on any other error)
 
-### Update own user data
+### Update own user data (TODO)
```
peterjanbrone, line 24: Do we need this in a first iteration? Needlessly overcomplicates imo.

API.md
```diff
@@ -1,86 +1,57 @@
...
 ```json
 {
-   "firstName": "John",
-   "lastName": "Doe",
-   "email": "john@doe.com",
    "tier": 1
 }
 ```
-also 400, 401, 500
+  - 400
```
peterjanbrone, line 38: You dropped `400` here, while the request requires a session, doesn't that imply the request can be unauthorized?

API.md
```diff
@@ -1,86 +1,57 @@
...
+  - 500
 
-## Password
+## Reports
```
peterjanbrone, line 41: I'm not a fan of the naming here. I'd call it `Track` and have the endpoint be `/track/upload/:skylink`.

API.md
```diff
@@ -1,86 +1,57 @@
...
+### Report an upload (TODO)
 
-* Endpoint: `/user/password`
+* Endpoint: `/report/upload/:skylink`
 * Verb: `POST`
 * Requires existing session: `true`
-* POST params:
-    - `oldPassword`: string
-    - `newPassword`: string
+* POST params: none
```
peterjanbrone, line 48: Why wouldn't we have the skylink in the POST body here? It's kind of strange that we're passing in data as a url parameter and keep the request body empty. If there's a reason though I guess I don't really mind as much, just lookeda bit off.

API.md
```diff
@@ -1,86 +1,57 @@
...
+* Returns: 204 or 400, 401, 500
+
+### Report a download (TODO)
+
+* Endpoint: `/report/download/:skylink`
```
peterjanbrone, line 53: Some comments apply on download

API.md
```diff
@@ -1,86 +1,57 @@
...
 
-* Endpoint: `/user/password`
+* Endpoint: `/report/upload/:skylink`
 * Verb: `POST`
 * Requires existing session: `true`
-* POST params:
-    - `oldPassword`: string
-    - `newPassword`: string
+* POST params: none
+* Returns: 204 or 400, 401, 500
```
peterjanbrone, line 49: `204` is indeed a good status code here, but I've always considered it to be boasting with http status codes. Especially in this case, a 200 is what I'd expect if I look at the route. I'm not 100% sure but for me at least I link a 204 to routes where you could potentially receive a large amount of data, like a download for example.

api/handlers.go
```diff
@@ -1,219 +1,75 @@
...
-		WriteError(w, err, http.StatusBadRequest)
-		return
-	}
-	if err != nil {
-		WriteError(w, err, http.StatusInternalServerError)
-		return
+	if errors.Contains(err, database.ErrUserNotFound) {
+		u, err = api.staticDB.UserCreate(req.Context(), sub, database.TierFree)
+		if err != nil {
+			WriteError(w, errors.AddContext(err, "user not found, failed to create"), http.StatusNotFound)
```
peterjanbrone, line 28: Hm, I think this should return `424 Failed Dependency`

api/handlers.go
```diff
@@ -1,219 +1,75 @@
...
-	}
-	claims = t.Claims.(jwt.MapClaims)
-	if reflect.ValueOf(claims["user_id"]).Kind() != reflect.String {
-		err = errors.New("the token does not contain the user_id we expect")
-	}
-	id = claims["user_id"].(string)
-	token = t
-	return
-}
+// TODO This will be needed bu only for changing the user's tier and expiration times. It will be driven by payments.
```
peterjanbrone, line 35:
```suggestion
// TODO This will be needed but only for changing the user's tier and expiration times. It will be driven by payments.
```

api/handlers.go
```diff
@@ -1,219 +1,75 @@
...
+//// userHandlerPUT updates an existing user.
+//func (api *API) userHandlerPUT(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
```
peterjanbrone, line 37: If this is commented out, perhaps we should get rid of it in the documentation as well. Also I believe the documentation doesn't mention user tier and expiration time as fields that you could update on a user, but rather mentions first and last name plus email.

api/jwt.go
peterjanbrone: TIL what niggle means 👍

api/jwt.go
```diff
@@ -3,77 +3,227 @@ package api
...
+// given token among the public keys provided by Oathkeeper.
+func keyForToken(token *jwt.Token) (interface{}, error) {
+	if _, ok := token.Method.(*jwt.SigningMethodRSA); !ok {
+		return nil, errors.New(fmt.Sprintf("unexpected signing method: %v", token.Header["alg"]))
+	}
+	keySet, err := oathkeeperPublicKeys()
+	if err != nil {
+		return nil, err
+	}
+	// Should this be replaced by a deferred recover in case the value is not
```
peterjanbrone, line 104: I would mark open questions like these as `TODO`.

peterjanbrone: To answer the question, I'm kind of surprised that a `Header` can have anything other than a string. Usually it's a string dictionary right? I guess it won't be much of an issue, aren't we only providing JWTs here that are given out by ORYs oathkeeper service?

api/jwt.go
```diff
@@ -3,77 +3,227 @@ package api
...
+			return nil, err
+		}
+		oathkeeperPubKeys = set
+	}
+	return oathkeeperPubKeys, nil
+}
+
+// tokenFromRequest extracts the JWT token from the request and returns it.
+// Returns an empty string if there is no token.
+func tokenFromRequest(r *http.Request) string {
```
peterjanbrone, line 145: this is slightly better:

```
authHeader := r.Header.Get("Authorization")
parts := strings.Split(authHeader, "Bearer")
if len(parts) != 2 {
    // return error
}

return strings.TrimSpace(parts[1])
```

I think error handling on the header value is definitely nice to have, we can be more informative that way.
Without it I believe an invalid token, and an improperly formatted Authorization header are treated the same.

api/jwt.go
```diff
@@ -3,77 +3,227 @@ package api
...
+	claims = t.Claims.(jwt.MapClaims)
+	if reflect.ValueOf(claims["sub"]).Kind() != reflect.String {
+		err = errors.New("the token does not contain the sub we expect")
+	}
+	defer func() {
+		if e := recover(); e != nil {
+			sub = ""
+			claims = nil
+			token = nil
+			err = errors.New("jwt claims don't contain a valid sub")
```
peterjanbrone, line 223: I think we'd want to include `e` here in the error message right?

api/routes.go
```diff
@@ -5,28 +5,22 @@ import (
...
-	// TODO Rate-limit these from the start to prevent brute force and other forms of abuse.
-	api.staticRouter.POST("/login", api.userLoginHandler)
-	api.staticRouter.POST("/user", api.userHandlerPOST)
-	api.staticRouter.GET("/user", Validate(api.userHandlerGET))
-	api.staticRouter.PUT("/user", Validate(api.userHandlerPUT))
-	api.staticRouter.POST("/user/password", Validate(api.userChangePasswordHandler))
-	//api.staticRouter.POST("/password/reset/request", api.passwordResetRequestHandler)
-	//api.staticRouter.POST("/password/reset/verify", api.passwordResetCompleteHandler)
-	//api.staticRouter.POST("/password/reset/complete", api.passwordResetCompleteHandler)
+	api.staticRouter.GET("/user", validate(api.userHandlerGET))
```
peterjanbrone, line 12: Documentation and implementation are diverging quite a bit here, I get that you want to save as much of the documentation you wrote already but honestly most of this stuff is easily copy pastable so I'd probably just get rid of as much as possible and keep accounts to a bare minimum of LoC.

database/database.go
peterjanbrone: +1

database/database.go
```diff
@@ -56,6 +52,7 @@ type (
 	DB struct {
 		staticDB    *mongo.Database
 		staticUsers *mongo.Collection
+		staticDep   lib.Dependencies
```
peterjanbrone, line 55: Why are all these fields listed as `static`. The `DB` struct doesn't have a mutex

database/database.go
```diff
@@ -93,20 +90,18 @@ func (db *DB) Disconnect(ctx context.Context) error {
...
 	}
-	// Emails must be unique. If we hit this then we have a serious
+	// Subs must be unique. If we hit this then we have a serious
 	// programmer error which endangers customer's data and finances.
 	// We should error out in order to prevent exposing the wrong user data.
 	if len(users) > 1 {
-		build.Critical(fmt.Sprintf("More than one user found for email '%s'!", email))
+		build.Critical(fmt.Sprintf("More than one user found for sub '%s'!", sub))
 		// The error message is intentionally cryptic.
 		return nil, ErrGeneralInternalFailure
```
peterjanbrone, line 106: This is fine but I like `Failed Dependency` a lot for this type of stuff personally. It makes it look like it's bricked, but not entirely our fault perhaps :D less shameful than a 500 error

database/database.go
```diff
@@ -140,41 +135,42 @@ func (db *DB) UserByID(ctx context.Context, id string) (*User, error) {
...
 	if err != nil && !errors.Contains(err, ErrUserNotFound) {
-		return errors.AddContext(err, "failed to query DB")
+		return nil, errors.AddContext(err, "failed to query DB")
 	}
 	if len(users) > 0 {
-		return ErrEmailAlreadyUsed
+		return nil, ErrUserAlreadyExists
+	}
+	u := &User{
+		ID:   primitive.ObjectID{},
```
peterjanbrone, line 149: Have you inspected your mongo collection locally? This looks weird, every document in mongo has a field called `_id` which is its primary key. I think you might be defining users that look like this:

```
{
      _id: ObjectId(),
      ID: ObjectId(),
      Sub: sub,
      Tier: tier,
}
```

Also keep in mind that I haven't seen anywhere where we configure indices, so both `ID` and `Sub` won't have an index for them, resulting in a full COLLSCAN on every request. That's pretty detrimental to performance. 

In mongo you can use 'explain' to see the query stage pipeline, if you are using robo3t you can just do `db.collection('users').find().explain()`. You can look at the `winningplan` to see whether it's making proper use of an index or not.

database/database.go
```diff
@@ -140,41 +135,42 @@ func (db *DB) UserByID(ctx context.Context, id string) (*User, error) {
...
 	}
 	ir, err := db.staticUsers.InsertOne(ctx, fields)
 	if err != nil {
-		return errors.AddContext(err, "failed to Insert")
+		return nil, errors.AddContext(err, "failed to Insert")
 	}
 	u.ID = ir.InsertedID.(primitive.ObjectID)
 	// Sanity check because races exist.
-	users, err = db.managedUsersByField(ctx, "email", string(u.Email))
+	users, err = db.managedUsersByField(ctx, "sub", sub)
```
peterjanbrone, line 164: You should make the `sub` field unique, this way you'd get a `UniqueConstraintViolationException` on the second insert. It's perfectly possible to be 100% atomic in mongo, they even have a new feature called `Transactions` that allow you to be ACID across collections!

database/database.go
```diff
@@ -140,41 +135,42 @@ func (db *DB) UserByID(ctx context.Context, id string) (*User, error) {
...
 	if len(users) > 1 {
 		// Race detected! Email no longer unique in DB. Delete new user.
```
peterjanbrone, line 166: I don't think it's a good idea to go and delete users when you found they exist twice. Wouldn't it be possible to end up going from 2 -> 0 users? Also simply deleting a user.. I don't think we should be doing that ever.

peterjanbrone: Perhaps we should consider soft-deletion 🤔  (in general, not in this particular situation)

database/user.go
```diff
@@ -1,113 +1,25 @@
...
-		ID        primitive.ObjectID `bson:"_id,omitempty" json:"-"`
-		FirstName string             `bson:"firstName" json:"firstName"`
-		LastName  string             `bson:"lastName" json:"lastName"`
-		Email     Email              `bson:"email" json:"email"`
-		Tier      int                `bson:"tier" json:"tier"`
-		Password  []byte             `bson:"password" json:"-"`
-		Salt      []byte             `bson:"salt" json:"-"`
-
-		dep lib.Dependencies
+		ID   primitive.ObjectID `bson:"_id,omitempty" json:"-"`
```
peterjanbrone, line 21: why are we hiding it in `json`?

database/database.go
peterjanbrone: Oh I see how the ID -> _id works, you specify it using `bson`. Ignore that, the rest still applies

test/database_test.go
```diff
@@ -162,13 +117,10 @@ func TestDatabase_UserDelete(t *testing.T) {
 		t.Fatal(err)
 	}
 
+	sub := "this is a sub"
```
peterjanbrone, line 120: (nit) use an actual `sub` here, that way the reader knows how it looks and has a better feel for what it is

database/database.go
```diff
@@ -193,79 +189,25 @@ func (db *DB) UserDelete(ctx context.Context, u *User) error {
...
-	}
-	// TODO What if we have a race a user gets this email right at this point?
 	// Update the user.
 	filter := bson.M{"_id": u.ID}
 	update := bson.M{"$set": bson.M{
-		"_id":       u.ID,
-		"firstName": u.FirstName,
-		"lastName":  u.LastName,
-		"email":     u.Email,
+		"_id":  u.ID,
```
peterjanbrone, line 198: I'm surprised mongo this works, it's not something you'd usually do. A document's `_id` is never overwritten. it's redundant so I would definitely remove it here.

database/database.go
```diff
@@ -193,79 +189,25 @@ func (db *DB) UserDelete(ctx context.Context, u *User) error {
...
+		"tier": u.Tier,
 	}}
 	ur, err := db.staticUsers.UpdateOne(ctx, filter, update)
 	if err != nil {
-		return errors.AddContext(err, "failed to Update")
+		return errors.AddContext(err, "failed to update")
 	}
 	if ur.UpsertedCount > 1 || ur.ModifiedCount > 1 {
 		build.Critical(fmt.Sprintf("updated more than one user! filter: %v, update: %v, update result:%v\n", filter, update, ur))
```
peterjanbrone, line 206: This is redundant, or it should be redundant, due to your `filter`. You are filtering on the primary key, so you could never ever have an `UpsertedCount` of more than one. Now that we are on the subject, I believe you wouldn't want to `upsert` at all here, and thus specify `upsert` to be false as you are trying to update an existing user object.

peterjanbrone (changes requested):

database/database.go
```diff
@@ -1,24 +1,20 @@
 package database
```
peterjanbrone, line 1: I haven't found a place where we ensure the collections? That'd be the place where you also ensure they have proper indices.

api/jwt.go
ro-tex: @m-cat can you share the lint config? I write stuff like that automatically sometimes and depend on the linter to help me out. But the linter can only do that if it's properly configured.

PS: TIL about niggle as well - it is an awesome word, it applies to so many things in my life! :D

database/database.go
ro-tex: I thought it was a well-known reference to identity string in JWTs. But I can change it to something clearer. The issue is that I can't say `id` here because that would imply the user id in Mongo, not the external user id we get from ORY.

Any suggestions?

API.md
ro-tex: I'll start in LIFO order because it makes more sense:

> How is this request authenticated, it says that it requires a session, how can one acquire a session if he's not a user yet.

This is something @m-cat also noted, I don't really explain how the authentication works. This is because the document was initially meant entirely for Karol who knows this quite well. I have now added a section to `API.md` that clears it out. For the sake of a complete conversation, I'll repeat some of it here:
 - All requests are authenticated by a JWT token passed in the headers.
 - This token is issued by ORY (I use that in place of switching between Kratos and Oathkeeper).
 - This token holds some infor about the user, most important of which is the `sub` or the user identity in ORY.

> I'm also confused as to how this would work, don't we need some user information to create a user object with? Like, an email at least?

No, all such information is stored in ORY. The only thing we need in our DB is the `sub` of the user or their unique id from ORY. That is enough for us to uniquely identify them without needlessly duplicating the user data already stored in ORY (email, names, etc.).

> This is an extremely weird API design. GET requests don't go and create new entities in the database.

We are stepping away from the normal RESTful path here. The reason for that is that we'll probably never work with entity ids and all that - we'll always operate on the current user which we can get from the JWT in the request. 

We can also split this into three calls (`GET /user` -> 404, `POST /user` without any additional data because we don't need it -> 201, `GET /user` -> 200) but that doesn't seem like a really useful flow to me. REST is a great design pattern but doesn't necessarily work well everywhere. For this particular case folding these three calls into a single one seems to work well. If we want to split them in the future we're very much free to do it.

API.md
ro-tex: I think the description is a little off. Maybe "Update user data" is better.

This endpoint is meant to allow us to update important user information such as tier, data/bandwidth limits, membership expiration dates and so on. It's not meant for changing names, emails, etc. - all that is handled by Kratos. This is what enables the operation of premium accounts.

database/database.go
ro-tex: Done.

database/user.go
ro-tex: Why would the calling service ever need this? We only use JSON to talk to the portal and the portal should never need the MongoID. It's an implementation detail that is not a concern for the caller.

test/database_test.go
ro-tex: Done.

database/database.go
ro-tex: I guess it won't be the fault of a dependency in this case - it seems like something we messed up, not a DB that's failing?

database/database.go
kwypchlo: it might be well known to people who know jwt well :)

database/database.go
```diff
@@ -193,75 +189,18 @@ func (db *DB) UserDelete(ctx context.Context, u *User) error {
...
-		return ErrEmailAlreadyUsed
-	}
-	// TODO What if we have a race a user gets this email right at this point?
 	// Update the user.
 	filter := bson.M{"_id": u.ID}
 	update := bson.M{"$set": bson.M{
-		"_id":       u.ID,
-		"firstName": u.FirstName,
-		"lastName":  u.LastName,
-		"email":     u.Email,
```
kwypchlo, line 221: why did you drop these fields?

API.md
peterjanbrone: Type I meant to say you dropped `401`.

api/routes.go
peterjanbrone: I'm talking about the documentation and implementation in general. The documentation lists some endpoints that are not available. You can ignore the comment I was just getting at "perhaps we should not invest any time yet in the documentation" but the point about Karol having a reference point is a good one.

database/database.go
peterjanbrone: The driver will probably have some examples in the repo, or in the documentation.

peterjanbrone: https://github.com/mongodb/mongo-go-driver/blob/master/examples/documentation_examples/examples.go

database/database.go
peterjanbrone: I never heard about it, and I knew what claims were, never encountered it before 🤷‍♂️

database/database.go
peterjanbrone: Don't we only do that for structs that have mutexes? I don't mind keeping it though. I thought we marked as static only when there's a mutex, so people trying to access the field know it's safe, which indeed can only be if it never changes if the struct has a mutex.

database/database.go
peterjanbrone: Yeah, we would be the dependency 😁  returning 500 error is just not done in my book. But perhaps that's just something that someone once told me that I now think is not done. I was once told by previous mentors that returning a 500 on purpose when designing an API is something you generally don't do.

Again, feel free to resolve.

database/database.go
peterjanbrone: I don't really mind but I think it's a bit weird we have endless discussions on database performance, with manual benchmark comparisons, and then not treat a missing index as thé highest priority.

As long as there's an issue for it and it's labeled as high prio I'm fine with anything.

database/database.go
peterjanbrone: Same comment as earlier. But I don't mind F/U, but we really have to F/U. 
The follow up is literally less than 20 lines of code by the way.

database/user.go
peterjanbrone: yeah but we are also hiding the sub, so basically the user is represented by its tier here, don't we need at least some form of unique id to link to a user outside of the API. 

A stupid example is if a react frontend were to list these users and you'd have to uniquely identify each row.. in react you have to identify children like that.. there a user id of sorts would come in handy.

Also debugging.

database/database.go
ro-tex: Changed.

API.md
ro-tex: Fixed.

database/database.go
ro-tex: In my experience a 500 denotes any error that's caused by the service's own code.

api/jwt.go
m-cat: > can you share the lint config?
@ro-tex Looks like my IDE is running `go-golint`. I don't think I have a config for it.

m-cat: And indeed there are other instances of this e.g. `isJsonErr -> isJSONErr`.

Minor stuff but at least y'all learned a new word 🙂

database/database.go
ro-tex: F/U: https://github.com/NebulousLabs/skynet-accounts/issues/18

database/database.go
ro-tex: The reason we discuss DB performance and benchmark is because we're more or less locking ourselves to that DB. If not by design then by the effort of switching. So, it's a very good idea to research before making the choice.
But this performance consideration is a long-term one, for when we have tens of thousands of users. It doesn't really matter for the first two weeks.

I'll open a F/U for this (high prio): https://github.com/NebulousLabs/skynet-accounts/issues/18

database/database.go
ro-tex: F/U: https://github.com/NebulousLabs/skynet-accounts/issues/18

database/database.go
ro-tex: These are managed by Kratos. The user can edit their names, change their password and email and get the respective confirmations and checks via Kratos. We also have them in the JWT token, so we don't need to store them in the DB.

test/database_test.go
```diff
@@ -18,7 +18,8 @@ func TestDatabase_UserBySub(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	sub := "this is a random test sub that shoudn't exist in the DB"
+	// A random sub that shouldn't exist in the DB.
+	sub := "695725d4-a345-4e68-919a-7395cb68484c"
```
m-cat, line 22: nit: Could make this a file-wide constant instead of repeating it in every test

database/user.go
ro-tex: That makes sense - I added the `sub`.

api/jwt.go
ro-tex: It seems that `go-golint` is ` golang.org/x/lint/golint`. Added to `make lint`. 👍

m-cat (dismissed):

test/database_test.go
ro-tex: Later :)

m-cat (dismissed):

api/jwt.go
ChrisSchinnerl: Wait why can't we just check if the field exists? e.g.
```go
subEntry, exists := claims["sub"]
if !exists {
  return errors.New("claim doesn't contain a sub")
}
sub = subEntry.(string)
```
?

We should unnecessary `recover` defers in production whenever possible.

database/database.go
ChrisSchinnerl: It just means that the field won't be touched after initializing it. Which implies that it's safe to access without locking. I don't really care that much whether we have static fields within structs that don't have a mutex but it might still be a useful indication. Especially if we ever add locking to a struct later on.

database/database.go
ChrisSchinnerl: Yeah I don't think we need to worry about this. If we try and create a user with an existing sub it should just fail with an error. This check doesn't really protect us anyway cause things might happen in the database between getting the user and deleting it. So we can just hope that the database enforces unique subs.

api/jwt.go
ro-tex: Oh, good idea.

The only thing is whether we can be sure that if it's there it would be a string and not, say, an int? If the type is defined as `map[string]interface{}` and the value is `{"foo":1}` then it might be an int, right?

It seems safer to just check it with `reflect`:
```go
	if reflect.ValueOf(claims["sub"]).Kind() != reflect.String {
		claims = nil
		err = errors.New("jwt claims don't contain a valid sub")
		return
	}
	sub = claims["sub"].(string)
```


OTOH, this is a validated token coming from Kratos and we know the `sub` will be a string. I'll just use your approach, because it's simpler.

m-cat (approved):

API.md
peterjanbrone: I think a cleaner approach would be to use a hook that inserts the database upon registration. You can resolve the comment, but I do think it's not the cleanest approach. Perhaps we should resolve in a follow-up.

database/database.go
peterjanbrone: I think you could argue it would actually be nice to have this information duplicated across both databases. That way we can query everything with mongo, and don't have to resort to writing manual queries to mysql (or cockroach - not sure what we landed on)..

If it is part of the JWT, perhaps we can update the user object on login every time. That's not necessarily a bad thing because we can take advantage of this event and add `lastLoginAt` or something 🤷‍♂️  super useful to have all of this information in mongo imo.

peterjanbrone (approved):
lookin' good

API.md
ro-tex: F/U: https://github.com/NebulousLabs/skynet-accounts/issues/19

database/database.go
ro-tex: I was thinking a lot about this and I don't see a case in which we would need this information and not have a JWT token to get it from. If we have it in Mongo we would need to check with the JWT every time before we use it in order to make sure our copy is up to date. Which renders having it entirely moot.

Put another way, this is an issue of separation of concerns - all user identity information is handled by Kratos while accounts handles only the status and history of the user.

PR #20, opened 2020-12-14
Ensure DB schema on service start.

### LINKED ISSUE #18: DB collections setup
# Overview

Add a mechanism in code that ensures all needed collections and indices are in place.

# Description

This resolves these discussions:
 * https://github.com/NebulousLabs/skynet-accounts/pull/17#discussion_r540789920 (ensure the correct collections and indices exist)
 * https://github.com/NebulousLabs/skynet-accounts/pull/17#discussion_r540784037 (indices for `sub` and `_id`)
 * https://github.com/NebulousLabs/skynet-accounts/pull/17#discussion_r540784679 (`sub` should be unique by index)

Resources:
 * https://github.com/mongodb/mongo-go-driver/blob/master/examples/documentation_examples/examples.go

# Details

## Collections

* `users` - contains all user account information such as tier, expiration time, used space, etc.
* `skylinks` - info about each skylink, such as link itself and size (in the future we might have more here)
* `uploads` - user id, skylink id, datetime
* `downloads` - user id, skylink id, datetime

### Collection `users`

* `_id` - I'm not sure if I need to make this one manually, the `_id` is the primary key, so there should be an index on it already
* `sub` - unique

### Collection `skylinks`

* `_id`
* `skylink` - unique

### Collection `uploads`

* `user_id`
* `skylink_id`

### Collection `downloads`

* `user_id`
* `skylink_id`

ro-tex:
This PR doesn't add indexes on the `_id` field because MongoDB [creates one automatically](https://docs.mongodb.com/manual/indexes/#default-id-index).

Closes #18

### DIFF
--- a/database/database.go
+++ b/database/database.go
@@ -285,11 +285,11 @@ func ensureDBSchema(ctx context.Context, db *mongo.Database) error {
 		dbDownloadsCollection: {
 			{
 				Keys:    bson.D{{"user_id", 1}},
-				Options: options.Index().SetName("uploadsUserID"),
+				Options: options.Index().SetName("downloadsUserID"),
 			},
 			{
 				Keys:    bson.D{{"skylink_id", 1}},
-				Options: options.Index().SetName("uploadsSkylinkID"),
+				Options: options.Index().SetName("downloadsSkylinkID"),
 			},
 		},
 	}
--- a/database/database.go
+++ b/database/database.go
@@ -36,13 +36,13 @@ var (
 	// dbUsersCollection defines the name of the "users" collection within
 	// skynet's database.
 	dbUsersCollection = "users"
-	// dbUsersCollection defines the name of the "skylinks" collection within
+	// dbSkylinksCollection defines the name of the "skylinks" collection within
 	// skynet's database.
 	dbSkylinksCollection = "skylinks"
-	// dbUsersCollection defines the name of the "uploads" collection within
+	// dbUploadsCollection defines the name of the "uploads" collection within
 	// skynet's database.
 	dbUploadsCollection = "uploads"
-	// dbUsersCollection defines the name of the "downloads" collection within
+	// dbDownloadsCollection defines the name of the "downloads" collection within
 	// skynet's database.
 	dbDownloadsCollection = "downloads"
 
--- a/database/database.go
+++ b/database/database.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/NebulousLabs/skynet-accounts/build"
 	"github.com/NebulousLabs/skynet-accounts/lib"
+	"github.com/sirupsen/logrus"
 
 	"gitlab.com/NebulousLabs/errors"
 	"go.mongodb.org/mongo-driver/bson"
@@ -35,6 +36,15 @@ var (
 	// dbUsersCollection defines the name of the "users" collection within
 	// skynet's database.
 	dbUsersCollection = "users"
+	// dbUsersCollection defines the name of the "skylinks" collection within
+	// skynet's database.
+	dbSkylinksCollection = "skylinks"
+	// dbUsersCollection defines the name of the "uploads" collection within
+	// skynet's database.
+	dbUploadsCollection = "uploads"
+	// dbUsersCollection defines the name of the "downloads" collection within
+	// skynet's database.
+	dbDownloadsCollection = "downloads"
 
 	// ErrUserNotFound is returned when we can't find the user in question.
 	ErrUserNotFound = errors.New("user not found")
@@ -77,6 +87,10 @@ func New(ctx context.Context, creds DBCredentials) (*DB, error) {
 		return nil, errors.AddContext(err, "failed to connect to DB")
 	}
 	database := c.Database(dbName)
+	err = ensureDBSchema(ctx, database)
+	if err != nil {
+		return nil, err
+	}
 	users := database.Collection(dbUsersCollection)
 	db := &DB{
 		staticDB:    database,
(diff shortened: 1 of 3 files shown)


database/database.go
```diff
@@ -35,6 +36,15 @@ var (
 	// dbUsersCollection defines the name of the "users" collection within
 	// skynet's database.
 	dbUsersCollection = "users"
+	// dbUsersCollection defines the name of the "skylinks" collection within
```
m-cat, line 39: Update docstrings (x3)

ro-tex: Caught red-handed copy-pasting...
Fixed.

database/database.go
```diff
@@ -237,3 +251,76 @@ func connectionString(creds DBCredentials) string {
...
+			},
+			{
+				Keys:    bson.D{{"skylink_id", 1}},
+				Options: options.Index().SetName("uploadsSkylinkID"),
+			},
+		},
+		dbDownloadsCollection: {
+			{
+				Keys:    bson.D{{"user_id", 1}},
+				Options: options.Index().SetName("uploadsUserID"),
```
m-cat, line 288:
```suggestion
				Options: options.Index().SetName("downloadsUserID"),
```

(x2)

m-cat (changes requested):
Just one comment

database/database.go
ro-tex: Copy-paste strikes again! No wonder, tbh, all the uploads/downloads code is identical. I actually considered writing it once and calling it with a parameter but it would have turned out more complicated that the current form. :)

m-cat (approved):

database/database.go
```diff
@@ -237,3 +251,76 @@ func connectionString(creds DBCredentials) string {
...
+		},
+		dbSkylinksCollection: {
+			{
+				Keys:    bson.D{{"skylink", 1}},
+				Options: options.Index().SetName("skylinksSkylinkUnique").SetUnique(true),
+			},
+		},
+		dbUploadsCollection: {
+			{
+				Keys:    bson.D{{"user_id", 1}},
```
peterjanbrone, line 277: (nit) are all fields in our documents snakecase? I would've expected `userId` here actually. I would actually change these to `user` or `userId` because we'll be accessing them in objects and using them in queries. We'll end up with stuff like 
```
const upload = await uploadsDB.findBy(...);
console.log(upload.user_id)
```

in any case we should be consistent, if we're consistent I'm no tbothered by it

database/database.go
```diff
@@ -237,3 +251,76 @@ func connectionString(creds DBCredentials) string {
...
+		coll, err := ensureCollection(ctx, db, collName)
+		if err != nil {
+			return err
+		}
+		iv := coll.Indexes()
+		names, err := iv.CreateMany(ctx, models)
+		if err != nil {
+			return errors.AddContext(err, "failed to create indexes")
+		}
+		logrus.Debugf("Created new indexes: %v\n", names)
```
peterjanbrone, line 306: I'm a bit confused by the usage of `logrus` here, you added it for just this line. Are we going to use it throughout? If we log the successful creation of indices, shouldn't we be logging failed attempts and/or errors as well?

peterjanbrone (approved):
thanks for adding this so quickly!

database/database.go
```diff
@@ -35,6 +36,15 @@ var (
 	// dbUsersCollection defines the name of the "users" collection within
 	// skynet's database.
 	dbUsersCollection = "users"
+	// dbSkylinksCollection defines the name of the "skylinks" collection within
+	// skynet's database.
+	dbSkylinksCollection = "skylinks"
+	// dbUploadsCollection defines the name of the "uploads" collection within
+	// skynet's database.
+	dbUploadsCollection = "uploads"
```
peterjanbrone, line 44: `dbEvents` 🤔 ? in favour of having uploads and downloads?

database/database.go
```diff
@@ -237,3 +251,76 @@ func connectionString(creds DBCredentials) string {
...
+// See https://docs.mongodb.com/manual/indexes/
+// See https://docs.mongodb.com/manual/core/index-unique/
+func ensureDBSchema(ctx context.Context, db *mongo.Database) error {
+	// schema defines a mapping between a collection name and the indexes that
+	// must exist for that collection.
+	schema := map[string][]mongo.IndexModel{
+		dbUsersCollection: {
+			{
+				Keys:    bson.D{{"sub", 1}},
+				Options: options.Index().SetName("usersSubUnique").SetUnique(true),
```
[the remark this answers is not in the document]
ro-tex: I don't know what names mongo auto-generates, mysql's names are not very readable. So, I'd prefer a custom name.

On both other counts - no problem, I'll remove the prefix and I'll make them snake case.

About the prefix - in MySQL indexes are listed globally, so it's helpful to prefix them. When I created these in Mongo I saw that we didn't need that but I was just too lazy to remove it.

database/database.go
ro-tex: The fields are snake case in Mongo but not in Go. Here's the `Upload` type:
```go
type Upload struct {
	ID        primitive.ObjectID  `bson:"_id,omitempty" json:"_id"`
	UserID    primitive.ObjectID  `bson:"user_id,omitempty" json:"user_id"`
	SkylinkID primitive.ObjectID  `bson:"skylink_id,omitempty" json:"skylink_id"`
	Timestamp primitive.Timestamp `bson:"timestamp" json:"timestamp"`
}
```
So, this is how the usage looks like:
```go
func (db *DB) UploadsByUser(ctx context.Context, user User) ([]Upload, error) {
	if user.ID.IsZero() {
		return nil, errors.New("invalid user")
	}
	filter := bson.D{{"user_id", user.ID}}
	c, err := db.staticUploads.Find(ctx, filter)
	if err != nil {
		return nil, err
	}
	uploads := make([]Upload, 0)
	err = c.All(ctx, &uploads)
	if err != nil {
		return nil, err
	}
	return uploads, nil
}
```
And then you can use the resulting `[]Upload` in a way idiomatic to Go.

database/database.go
ro-tex: I wanted to have some structured logging. `logrus` looked like a decent mature choice. I intend to add logging to many more places, so we can know what's going on.

database/database.go
ro-tex: I don't see a use case in which we would need both in the same query. No need to strain those indexes if we don't need to. Unless there's a reason to prefer that?

peterjanbrone: Some things would be slightly easier, like aggregating the amount of downloads/uploads per user, or finding latest user activitiy. Events would also be more generic, we could add more types of events over time.
Ignore it, we can always rename and refactor later.

PR #21, opened 2020-12-14
Uploads and downloads

ro-tex:
This PR adds the uploads and downloads - the collections in the database, the handlers and endpoints for registering them and for getting lists of those per user.

When a user uploads a skylink we check in the DB to see whether we know this skylink already (that's possible if it's a repin or exactly the same content). If that's the case we just add the skylink's size to the user's used space tally. Otherwise we start a background thread that fetches the metadata of the skylink from the local siad and updates both the skylink's record and the user's used space.

~~The lists returned by `/user/uploads` and `/user/downloads` don't have pagination or anything for the moment (KISS) - we'll add that when we need it.~~


database/skylink_test.go
```diff
@@ -0,0 +1,45 @@
...
+// TestValidateSkylink ensures validateSkylink properly returns the skylink hash
+func TestValidateSkylink(t *testing.T) {
+	tests := []struct {
+		name  string
+		in    string
+		out   string
+		valid bool
+	}{
+		{
+			name:  "valid0",
```
m-cat, line 14: nit: Think we can remove the name field and just use the `in` value in the error message.

api/handlers.go
```diff
@@ -31,3 +32,124 @@ func (api *API) userHandlerGET(w http.ResponseWriter, req *http.Request, _ httpr
...
+	if err != nil {
+		WriteError(w, err, http.StatusInternalServerError)
+		return
+	}
+	_, err = api.staticDB.UploadCreate(req.Context(), *u, *skylink)
+	if err != nil {
+		WriteError(w, err, http.StatusInternalServerError)
+		return
+	}
+	if skylink.Size == 0 {
```
m-cat, line 105: Should this be

```suggestion
	if skylink.Size != 0 {
```

?

m-cat: Nevermind! The name `Size` confused me.

m-cat (commented):
Looks good so far. Is it just me or is the diff broken again? If so, maybe raise a new PR

database/skylink_test.go
ro-tex: Good idea, we will have a clue what's going on before even opening the code.

api/handlers.go
ro-tex: I added a little more info in the comment. Hopefully that would make it a little less confusing.

ro-tex:
> Looks good so far. Is it just me or is the diff broken again? If so, maybe raise a new PR

It looks good to me now but yesterday I was unable to open it at all. GitHub is having some issues..

m-cat (dismissed):

api/handlers.go
```diff
@@ -31,3 +32,123 @@ func (api *API) userHandlerGET(w http.ResponseWriter, req *http.Request, _ httpr
 	}
 	WriteJSON(w, u)
 }
+
+// userUploadsHandler returns all uploads made by the current user.
+func (api *API) userUploadsHandler(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
```
peterjanbrone, line 37: I'm not sure why we would not just add pagination from the start. It's just two query string parameters `offset` (or `skip`) and `limit`. Pagination is a pretty basic luxury to have when serving large JSON responses.

api/handlers.go
```diff
@@ -31,3 +32,123 @@ func (api *API) userHandlerGET(w http.ResponseWriter, req *http.Request, _ httpr
...
+	sub, _, _, err := tokenFromContext(req)
+	if err != nil {
+		WriteError(w, err, http.StatusInternalServerError)
```
peterjanbrone, line 40: Why are we returning 500 internal server errors here? In this particular case I would return `400 Bad Request`. I've always been thought it's bad API design to return 500 errors.

api/handlers.go
```diff
@@ -31,3 +32,123 @@ func (api *API) userHandlerGET(w http.ResponseWriter, req *http.Request, _ httpr
...
+		return
+	}
+	u, err := api.staticDB.UserBySub(req.Context(), sub)
+	// TODO Should we create the user in case it's not found?
```
peterjanbrone, line 44: I would return `404` or `400`. Kind of leaning towards a `400` because I think ` `404` is at times dubious between route not found or (in this case) user not found.

api/handlers.go
```diff
@@ -31,3 +32,123 @@ func (api *API) userHandlerGET(w http.ResponseWriter, req *http.Request, _ httpr
...
+	if err != nil {
+		WriteError(w, err, http.StatusInternalServerError)
+		return
+	}
+	ups, err := api.staticDB.UploadsByUser(req.Context(), *u)
+	if err != nil {
+		WriteError(w, err, http.StatusInternalServerError)
```
peterjanbrone, line 51: This would/could be a failed dependency perhaps?

api/handlers.go
```diff
@@ -31,3 +32,123 @@ func (api *API) userHandlerGET(w http.ResponseWriter, req *http.Request, _ httpr
...
+	ups, err := api.staticDB.DownloadsByUser(req.Context(), *u)
+	if err != nil {
+		WriteError(w, err, http.StatusInternalServerError)
+	}
+	WriteJSON(w, ups)
+}
+
+// trackUploadHandler registers a new upload in the system.
+func (api *API) trackUploadHandler(w http.ResponseWriter, req *http.Request, ps httprouter.Params) {
+	sl := ps.ByName("skylink")
```
peterjanbrone, line 78: on `trackDownload` we return a "missing parameter .." 400, we should do that here as well

api/handlers.go
```diff
@@ -31,3 +32,123 @@ func (api *API) userHandlerGET(w http.ResponseWriter, req *http.Request, _ httpr
...
+	sub, _, _, err := tokenFromContext(req)
```
peterjanbrone, line 79: I don't really understand how we're doing the authentication of these routes. Is there a middleware of sorts that handles authentication. I'm not sure we need to be asserting that the user exists. If we authenticate a user we could set its id on the request scope and simply insert the "tracking document" here with that user id. Saving a trip to the database. I feel there's quite a lot of stuff going on here. Like 3 or 4 database calls? Is this being triggered asynchronously or is this a blocking action. On upload it doesn't matter much but on download I think it would be less than ideal to have this be a blocking action.

api/handlers.go
```diff
@@ -31,3 +32,123 @@ func (api *API) userHandlerGET(w http.ResponseWriter, req *http.Request, _ httpr
...
+	if err != nil {
+		WriteError(w, err, http.StatusInternalServerError)
+		return
+	}
+	_, err = api.staticDB.UploadCreate(req.Context(), *u, *skylink)
+	if err != nil {
+		WriteError(w, err, http.StatusInternalServerError)
+		return
+	}
+	if skylink.Size == 0 {
```
peterjanbrone, line 103: In what scenario would the database contain a Skylink with size 0. Wouldn't we have fetched the metadata when we inserted the Skylink in the first place?

api/handlers.go
```diff
@@ -31,3 +32,123 @@ func (api *API) userHandlerGET(w http.ResponseWriter, req *http.Request, _ httpr
...
+	if sl == "" {
+		WriteError(w, errors.New("missing parameter 'skylink'"), http.StatusBadRequest)
+		return
+	}
+	sub, _, _, err := tokenFromContext(req)
+	if err != nil {
+		WriteError(w, err, http.StatusInternalServerError)
+		return
+	}
+	skylink, err := api.staticDB.Skylink(req.Context(), sl)
```
peterjanbrone, line 134: Here we don't seem to care if the Skylink size is 0. I take it that's because on download we don't need to keep track of something like "amount downloaded" like we have to in upload with "used storage"?

api/jwt.go
```diff
@@ -120,17 +121,21 @@ func keyForToken(token *jwt.Token) (interface{}, error) {
 // Encoding RSA pub key: https://play.golang.org/p/mLpOxS-5Fy
 func oathkeeperPublicKeys() (*jwk.Set, error) {
 	if oathkeeperPubKeys == nil {
+		fmt.Println("fetching JWKS from oathkeeper")
```
peterjanbrone, line 124: are these print statements left in here intentionally? weren't we using a log library (logrus)?

api/jwt.go
```diff
@@ -139,18 +144,37 @@ func oathkeeperPublicKeys() (*jwk.Set, error) {
...
-	if len(parts) != 2 {
-		return "", errors.New("invalid authorization header")
+	if len(parts) == 2 {
+		return strings.TrimSpace(parts[1]), nil
+	}
+
+	// Check the cookie for a token.
+	cookie, err := r.Cookie(CookieName)
+	if err != nil {
+		return "", err
```
peterjanbrone, line 159: could do with some context on the error here I guess

api/jwt.go
```diff
@@ -224,3 +248,19 @@ func tokenFromContext(req *http.Request) (sub string, claims jwt.MapClaims, toke
 	token = t
 	return
 }
+
+// TokenExpiration extracts and returns the `exp` claim of the given token.
+// NOTE: It does NOT validate the token!
+func TokenExpiration(t *jwt.Token) (int64, error) {
```
peterjanbrone, line 254: (nit) why upper case?

api/routes.go
```diff
@@ -2,32 +2,49 @@ package api
...
 }
 
 // validate ensures that the user making the request has logged in.
 func validate(h httprouter.Handle) httprouter.Handle {
 	return func(w http.ResponseWriter, req *http.Request, ps httprouter.Params) {
-		t, err := tokenFromRequest(req)
+		tokenStr, err := tokenFromRequest(req)
 		if err != nil {
-			WriteError(w, err, http.StatusBadRequest)
+			fmt.Println("error while fetching token from request", err)
```
peterjanbrone, line 28: use logger?

database/download.go
```diff
@@ -0,0 +1,87 @@
...
+import (
+	"context"
+	"time"
+
+	"gitlab.com/NebulousLabs/errors"
+	"go.mongodb.org/mongo-driver/bson"
+	"go.mongodb.org/mongo-driver/bson/primitive"
+)
+
+// Download ...
```
peterjanbrone, line 12: are the `....` a leftover TODO to extend later?

database/download.go
```diff
@@ -0,0 +1,87 @@
...
+	return downloads, nil
+}
+
+// DownloadsByUser fetches all downloads by this user
+func (db *DB) DownloadsByUser(ctx context.Context, user User) ([]Download, error) {
+	if user.ID.IsZero() {
+		return nil, errors.New("invalid user")
+	}
+	filter := bson.D{{"user_id", user.ID}}
+	c, err := db.staticDownloads.Find(ctx, filter)
```
peterjanbrone, line 77: Adding pagination should be as easy as to use `FindAndCount` here with offset and skip passed in as options. Not sure if you'll have `FindAndCount` at your disposal but most libraries have some sort of built in query to support list routes.

database/skylink.go
```diff
@@ -0,0 +1,94 @@
...
+	"regexp"
+
+	"gitlab.com/NebulousLabs/errors"
+	"go.mongodb.org/mongo-driver/bson"
+	"go.mongodb.org/mongo-driver/bson/primitive"
+	"go.mongodb.org/mongo-driver/mongo"
+)
+
+var (
+	skylinkRE *regexp.Regexp = nil
```
peterjanbrone, line 14: why is this nil?

database/skylink.go
```diff
@@ -0,0 +1,94 @@
...
+
+// Skylink represents a skylink object in the DB.
+type Skylink struct {
+	ID      primitive.ObjectID `bson:"_id,omitempty" json:"-"`
+	Skylink string             `bson:"skylink" json:"skylink"`
+	Size    int64              `bson:"size" json:"size"`
+}
+
+// Skylink gets the DB object for the given skylink.
+// If it doesn't exist it creates it.
```
peterjanbrone, line 25: Oh I guess this is why the filesize is sometimes 0 and there's no message scheduled. I guess you're not scheduling a message here to have loose coupling, which is nice of course. You could've also put a service on top of it though and have this return `null` in case it's not found and then have the service schedule a metadata fetch + insert a skylink object

database/skylink.go
```diff
@@ -0,0 +1,94 @@
...
+		Skylink: skylinkHash,
+	}
+	// Try to find the skylink in the database.
+	filter := bson.D{{"skylink", skylinkHash}}
+	sr := db.staticSkylinks.FindOne(ctx, filter)
+	err = sr.Decode(&skylinkRec)
+	if err == mongo.ErrNoDocuments {
+		// It's not there, insert it.
+		var ior *mongo.InsertOneResult
+		ior, err = db.staticSkylinks.InsertOne(ctx, skylinkRec)
```
peterjanbrone, line 42: I think this'll be subject to duplciate key exceptions if two connections are inserting the same skylink at the same time. You could use `upsert` to circumvent this issue, and I think we should.

database/skylink.go
```diff
@@ -0,0 +1,94 @@
...
+	if err != nil {
+		return errors.AddContext(err, "failed to update")
+	}
+	return nil
+}
+
+// validateSkylink extracts the skylink hash from the given skylink that might
+// have protocol, path, etc. within it.
+func validateSkylink(skylink string) (string, error) {
+	if skylinkRE == nil {
```
peterjanbrone, line 86: why don't we simly set the regex in the var opposed to lazy loading it like this?

database/user.go
```diff
@@ -18,8 +26,147 @@ type (
...
+func (db *DB) UserBySub(ctx context.Context, sub string) (*User, error) {
+	users, err := db.managedUsersByField(ctx, "sub", sub)
+	if err != nil {
+		return nil, err
+	}
+	// Subs must be unique. If we hit this then we have a serious
+	// programmer error which endangers customer's data and finances.
+	// We should error out in order to prevent exposing the wrong user data.
+	if len(users) > 1 {
+		build.Critical(fmt.Sprintf("More than one user found for sub '%s'!", sub))
```
peterjanbrone, line 47: we should be enforcing this on database level using a unique index on the `sub` field

metafetcher/fetcher.go
```diff
@@ -0,0 +1,136 @@
...
+// a given user.
+type Message struct {
+	UserID    primitive.ObjectID
+	SkylinkID primitive.ObjectID
+	Attempts  uint8
+}
+
+// MetaFetcher is a background task that listens for messages on its queue and
+// then processes them.
+type MetaFetcher struct {
```
peterjanbrone, line 31: (nit) why abbreviate to `Meta` - I would call it SkylinkMetadataFetcher to make it super obvious

metafetcher/fetcher.go
```diff
@@ -0,0 +1,136 @@
...
+	go mf.threadedStartQueueWatcher(ctx)
+
+	return &mf
+}
+
+// threadedStartQueueWatcher starts a loop over the Queue that processes each
+// incoming message in a separate goroutine.
+func (mf *MetaFetcher) threadedStartQueueWatcher(ctx context.Context) {
+	for m := range mf.Queue {
+		// Process each message in a separate goroutine because fetching the
```
peterjanbrone, line 60: This made me think about something, it might be a good idea to verify whether we can squash the message somehow. Because this process takes such a long time, I think a decent portion of the queue will be redundant messages for the same Skylink. We could keep a limited sized cache, or do a database query before executing the actual HEAD call perhaps?

peterjanbrone: Oh I guess we have the added complexity the user's storage needs to be updated. That's not ideal by the way, the metadata fetcher should not be responsible for updating that it feels like to me.

metafetcher/fetcher.go
```diff
@@ -0,0 +1,136 @@
...
+		Method: http.MethodHead,
+		URL:    skylinkURL,
+		Header: http.Header{"User-Agent": []string{"Sia-Agent"}},
+	}
+	client := http.Client{}
+	res, err := client.Do(&req)
+	if err != nil || res.StatusCode > 399 {
+		mf.logger.Tracef("Failed to fetch skyfile. Skylink: %s, error: %v\n", sl.Skylink, err)
+		if m.Attempts >= maxAttempts {
+			mf.logger.Debugf("Message exceeded its maximum number of attempts, dropping: %v\n", m)
```
peterjanbrone, line 100: Should we consider `Errorf` here?

peterjanbrone (changes requested):

api/handlers.go
ro-tex: Added.

api/handlers.go
ro-tex: I would go for "Unauthorized" for this particular case. It covers both the case where a token was not provided, did not contain the needed claims or was not valid.

We've talked about 500s before - I disagree that it's bad design to return those. An internal server error is a valid response that informs the user what happened. I do agree that we should strive to never return them but that's different from not having them on the API - if our service somehow failed then we should return a 500.

Again, in this case I agree that a 500 is not the best response, so I'm changing it to Unauthorized. If you think that bad Request is better then we can discuss that.

api/handlers.go
ro-tex: The thinking here was focused on the way this is going to be used, rather that strictly following REST rules. From REST point of view this endpoint is shady in two ways:
1. It suggest that it's going to create a user if it doesn't exist, so it combines the functionality of GET and POST.
2. It doesn't follow the standard pattern of `/resource/resource_id` because that doesn't make much sense in our use case - a user is not supposed to access the data of another user.

If we want to go strict REST, we need to change this to `/user/user_id` but that will require our code in nginx to inspect the cookies, decrypt the right one, parse the JSON, extract the id and then make the call. Given the use case, this is adding a lot of complexity without a real benefit. I would prefer to use REST as guidelines and write the code in the most convenient way for our setup, meaning that we have just `GET /user` which returns the data of the currently authenticated user (or return Unauthorized).

There is still the question of whether this GET call should also create the user if it doesn't exist. On one hand, this is unintuitive and unexpected for a GET call. On the other hand, it's very convenient, especially when you consider that with the GET call we already have all the data we need and it will simplify the flow a lot. Otherwise the logic in nginx will be this:
1. GET /user
2. If the response is 404, POST /user (with the same params) && GET /user
3. Do the rest of the work.

I don't have a strong opinion on this one. Maybe @kwypchlo can weigh in here.

api/handlers.go
ro-tex:
```
424 Failed Dependency (WebDAV)
    The request failed due to failure of a previous request.
```
I don't think there is a previous request that has failed here.

All 4xx responses indicate that the service is in good health and the problem is on the user's side. So, they should either fix their request (400), get authenticated (401), get more rights (403), make sure they are trying to get an existing resource (404), try again later (429) or something else that's in their control. The failed dependency would mean "make sure your previous requests pass and try again". What we have here is our internal infrastructure preventing the request from being fulfilled. This is unlikely to be fixed by actions on the user's side, so it should be a 5xx.

api/handlers.go
ro-tex: 👍

api/handlers.go
ro-tex: 1. We are validating the token in a middleware called `validate` that's defined in `routes.go`. A valid token does not mean that we have the user in the DB, only that Kratos knows about them.
2. Tracking documents and so on are not something I'm currently familiar with. That's a little unfortunate but I haven't had the time to read up on Mongo's functionality, so I'm using bare-bones approach for the moment. We can definitely improve on this in the future. For the moment delivering a functioning first version seems to be more important.
3. Auto-creating the user is something that's already done in the code. This is already touched upon in other discussions in this PR but I believe that's a good approach in our use case - if the user doesn't exist create a new free user and carry on the request.
4. This is blocking on the service's side. I'm not sure if it needs to be blocking on nginx level. It will happen both on upload and on download but it will happen only once (and after that we'll have the user in the DB).

api/handlers.go
ro-tex: Fetching a skylink's metadata is slow and might fail. That's why we do it in the background.

The flow is like this:
1. A new skylink is registered as either uploaded or downloaded.
2. We insert it in the DB with size 0 and add a background job that will handle that.
3. We return a response to the request and base the data in that response on the current amount of downloads and uploads made by the user without taking into account this skylink. This might allow users to go a little over their quota but we're OK with that (confirmed with David).

The background job does the following:
1. Downloads the metadata of the skylink.
2. Updates the data about the skylink in the DB.
3. Updates the upload/download usage by all users who uploaded/downloaded the skylink.

api/handlers.go
ro-tex: That's correct.

api/jwt.go
ro-tex: I should transition them to logrus and prune them a bit. But yes, they should not be prints.

api/jwt.go
ro-tex: added

api/jwt.go
ro-tex: fixed

database/download.go
ro-tex: Pagination added.

database/skylink.go
ro-tex: Hm, good question. Fixed.

database/skylink.go
ro-tex: This is a good idea. I guess we can also scan the table with skylinks for entries with zero size and just process them. We just need to make sure there are no fetches currently queued. Then we can fetch the sizes and update the users who uploaded them. Or we can also skip updating the users on this pass and just recalculate the uploaded amount by each user once per day after we've updated all skylink sizes. That will make sure we have the right counts.

This looks like a good candidate for an independent F/U.

ro-tex: F/U: https://github.com/NebulousLabs/skynet-accounts/issues/22

metafetcher/fetcher.go
ro-tex: `Error` is a bit severe for this. Maybe `Warn`? We intend to have a daily/weekly cover for this in the form of a scan that attempts to update the sizes on all zero-sized skylinks (#22).

metafetcher/fetcher.go
ro-tex: Because I hate long lines, haha.
I don't think that the extra characters there would resolve any ambiguity or bring anything useful to the table.

database/user.go
ro-tex: You're right - we should. And we are. I'll remove this check.

metafetcher/fetcher.go
ro-tex: 1. You are right, we can check the DB before making an expensive HTTP call. I'll add that. This doesn't get in the way of the rest of the logic.
2. I can agree with the idea that the fetcher shouldn't be updating the uploaded amount of the user but this would mean that the user can go *way* over their upload limit before we start limiting them. I'm not sure this is acceptable.

database/skylink.go
ro-tex: Good point. Added.

api/cookie.go
```diff
@@ -0,0 +1,73 @@
...
+	envCookieHashKey = "COOKIE_HASH_KEY"
+	// envCookieEncKey holds the name of the env var which holds the key we use
+	// to encrypt cookies.
+	envCookieEncKey = "COOKIE_ENC_KEY"
+)
+
+var (
+	// _sc holds an instance of `securecookie`, so we don't instantiate it more
+	// than once.
+	_sc *securecookie.SecureCookie = nil
```
ChrisSchinnerl, line 31: A more idiomatic solution would be

```go
sc = func() *securecookie.SecureCookie {
	var hashKey = []byte(os.Getenv(envCookieHashKey))
	var blockKey = []byte(os.Getenv(envCookieEncKey))
	return securecookie.New(hashKey, blockKey)
}()
```

or alternatively to use the `init` function. https://golang.org/doc/effective_go.html#init
Then you don't need the underscore prefix to indicate a special variable.

api/handlers.go
```diff
@@ -2,32 +2,153 @@ package api
...
+	ups, err := api.staticDB.UploadsByUser(req.Context(), *u, offset, limit)
+	if err != nil {
+		WriteError(w, err, http.StatusInternalServerError)
+	}
+	WriteJSON(w, ups)
+}
+
+// userDownloadsHandler returns all downloads made by the current user.
+func (api *API) userDownloadsHandler(w http.ResponseWriter, req *http.Request, ps httprouter.Params) {
+	sub, _, _, err := tokenFromContext(req)
```
ChrisSchinnerl, line 53: A potentially nicer way to handle this and a good follow-up would be to introduce a `RequireToken` middleware.
Similar to the `RequirePassword` middleware in Sia. 

Then you could extend this method to 
```func (api *API) userDownloadsHandler(w http.ResponseWriter, req *http.Request, ps httprouter.Params, token Token) {```

and register it with ```router.GET("/foo/bar", RequireToken(api.userDownloadsHandler))```

This saves you the call to `tokenFromContext` in every handler that needs it.

File: api/handlers.go [hunk 27]
```diff
@@ -2,32 +2,153 @@ package api
...
 	}
-	u, err := api.staticDB.UserBySub(req.Context(), sub)
-	if err != nil && !errors.Contains(err, database.ErrUserNotFound) {
+	sl := ps.ByName("skylink")
+	if sl == "" {
+		WriteError(w, errors.New("missing parameter 'skylink'"), http.StatusBadRequest)
+		return
+	}
+	skylink, err := api.staticDB.Skylink(req.Context(), sl)
+	if err == database.ErrInvalidSkylink {
```
ChrisSchinnerl, line 85: Might be good to start using `errors.Contains` from the beginning to avoid having to refactor code in the future.

api/handlers.go
ChrisSchinnerl: If all we use the token for is fetching users you might also just pass in the User to the handler.

api/cookie.go
ro-tex: I thought about the `init` approach but I've never heard anything good about it and multiple sources complain about the added "magic" and weird bugs they managed to create with it. I'd rather stay away from it where possible and the proposed function solution is pretty nice and readable.

api/handlers.go
ro-tex: Sounds good, added #23 as F/U.

api/handlers.go
kwypchlo: What is the case where `err` would not be `nil`? Database down for example?

ro-tex: Any DB error or an invalid user. The latter one shouldn't happen because of the auto-create mechanic we're using.

README.md
```diff
@@ -15,6 +15,9 @@ SKYNET_DB_PORT="27017"
 SKYNET_DB_USER="username"
 SKYNET_DB_PASS="password"
 SKYNET_ACCOUNTS_PORT=3000
+COOKIE_DOMAIN="siasky.xyz"
```
peterjanbrone, line 18: should we change this to be `siasky.net` in the repo?

peterjanbrone (approved):

README.md
ro-tex: Done.

peterjanbrone (approved):

api/routes.go
```diff
@@ -3,31 +3,47 @@ package api
...
 		if err != nil {
+			api.staticLogger.Traceln("error while validating token", err)
 			WriteError(w, err, http.StatusUnauthorized)
 			return
 		}
+		// If we don't have a valid cookie with reasonably long remaining TTL
+		// then set one.
+		c, err := req.Cookie(CookieName)
+		exp, _ := tokenExpiration(token)
+		if err != nil || !c.Expires.Equal(time.Unix(exp, 0)) {
```
m-cat, line 41: I don't understand this part of the condition:

```go
!c.Expires.Equal(time.Unix(exp, 0)
```

Isn't the chance that the two times are equal very low?

main.go
```diff
@@ -56,19 +61,38 @@ func main() {
 	if !ok {
 		port = "3000"
 	}
+	portal, ok := os.LookupEnv(envPortal)
+	if !ok {
+		portal = "https://siasky.net"
```
m-cat, line 66: nit: Can make this a constant

m-cat (approved):

api/routes.go
ro-tex: You are right, I should be checking if the expiration time is *before* the current moment, no equal to it.
I'll fix it in a F/U.

main.go
ro-tex: The `portal` var? Sure, I'll F/U.

PR #24, opened 2021-02-04
Upload/Download Response DTOs

ro-tex:
* Introduce Response DTOs, so we can control how the data is represented in our responses.
* Tweak the data content and structure according to @kwypchlo's needs.
* Always use UTC time.
* Switch the time representation in the DB from primitive.Timestamp to time.Time.

Example responses:
```json
[
    {
        "id": "601c10a1fc760d69e53f1092",
        "skylink": "HELLO__5MJyl1DKyfbuQMY_XBOBC9bR7idiU6isp6LXxEw",
        "name": "This is my HELLO skyfile!",
        "size": 132412341234,
        "uploaded_on": "2021-02-04T15:20:01.577Z"
    }
]
```
```json
[
    {
        "id": "601c1626df96cb02cd2e84a0",
        "skylink": "HELLO__5MJyl1DKyfbuQMY_XBOBC9bR7idiU6isp6LXxEw",
        "name": "This is my HELLO skyfile!",
        "size": 132412341234,
        "downloaded_on": "2021-02-04T15:43:34.871Z"
    }
]
```

### DIFF
--- a/api/cookie.go
+++ b/api/cookie.go
@@ -36,9 +36,6 @@ var (
 // writeCookie is a helper function that writes the given JWT token as a
 // secure cookie.
 func writeCookie(w http.ResponseWriter, token string, exp int64) error {
-	if exp <= 0 || time.Unix(exp, 0).Before(time.Now().UTC()) {
-		exp = time.Now().UTC().Unix()
-	}
 	encodedValue, err := secureCookie.Encode(CookieName, token)
 	if err != nil {
 		return err
--- a/api/cookie.go
+++ b/api/cookie.go
@@ -36,8 +36,8 @@ var (
 // writeCookie is a helper function that writes the given JWT token as a
 // secure cookie.
 func writeCookie(w http.ResponseWriter, token string, exp int64) error {
-	if exp <= 0 || time.Unix(exp, 0).Before(time.Now()) {
-		exp = time.Now().Unix()
+	if exp <= 0 || time.Unix(exp, 0).Before(time.Now().UTC()) {
+		exp = time.Now().UTC().Unix()
 	}
 	encodedValue, err := secureCookie.Encode(CookieName, token)
 	if err != nil {
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -187,7 +187,7 @@ func fetchOffset(form url.Values) (int, error) {
 	return offset, nil
 }
 
-// fetchLimit extracts the offset from the params and validates its value.
+// fetchLimit extracts the limit from the params and validates its value.
 func fetchLimit(form url.Values) (int, error) {
 	limit, _ := strconv.Atoi(form.Get("limit"))
 	if limit < 0 {
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -2,6 +2,7 @@ package api
 
 import (
 	"net/http"
+	"net/url"
 	"strconv"
 
 	"github.com/NebulousLabs/skynet-accounts/database"
@@ -28,7 +29,7 @@ func (api *API) userHandler(w http.ResponseWriter, req *http.Request, _ httprout
 }
 
 // userUploadsHandler returns all uploads made by the current user.
-func (api *API) userUploadsHandler(w http.ResponseWriter, req *http.Request, ps httprouter.Params) {
+func (api *API) userUploadsHandler(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
 	sub, _, _, err := tokenFromContext(req)
 	if err != nil {
 		WriteError(w, err, http.StatusUnauthorized)
@@ -39,20 +40,29 @@ func (api *API) userUploadsHandler(w http.ResponseWriter, req *http.Request, ps 
 		WriteError(w, err, http.StatusInternalServerError)
 		return
 	}
-	offset, err1 := fetchOffset(ps)
-	limit, err2 := fetchLimit(ps)
+	if err = req.ParseForm(); err != nil {
+		WriteError(w, err, http.StatusBadRequest)
+	}
+	offset, err1 := fetchOffset(req.Form)
+	limit, err2 := fetchLimit(req.Form)
 	if err = errors.Compose(err1, err2); err != nil {
 		WriteError(w, err, http.StatusBadRequest)
 	}
-	ups, err := api.staticDB.UploadsByUser(req.Context(), *u, offset, limit)
+	ups, total, err := api.staticDB.UploadsByUser(req.Context(), *u, offset, limit)
 	if err != nil {
 		WriteError(w, err, http.StatusInternalServerError)
 	}
-	WriteJSON(w, ups)
+	response := database.UploadsResponseDTO{
+		Items:      ups,
+		Offset:     offset,
+		PageSize:   limit,
+		TotalCount: total,
+	}
+	WriteJSON(w, response)
 }
 
 // userDownloadsHandler returns all downloads made by the current user.
-func (api *API) userDownloadsHandler(w http.ResponseWriter, req *http.Request, ps httprouter.Params) {
+func (api *API) userDownloadsHandler(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
 	sub, _, _, err := tokenFromContext(req)
 	if err != nil {
 		WriteError(w, err, http.StatusInternalServerError)
@@ -63,16 +73,25 @@ func (api *API) userDownloadsHandler(w http.ResponseWriter, req *http.Request, p
 		WriteError(w, err, http.StatusInternalServerError)
 		return
 	}
-	offset, err1 := fetchOffset(ps)
-	limit, err2 := fetchLimit(ps)
+	if err = req.ParseForm(); err != nil {
+		WriteError(w, err, http.StatusBadRequest)
+	}
+	offset, err1 := fetchOffset(req.Form)
+	limit, err2 := fetchLimit(req.Form)
 	if err = errors.Compose(err1, err2); err != nil {
 		WriteError(w, err, http.StatusBadRequest)
 	}
-	ups, err := api.staticDB.DownloadsByUser(req.Context(), *u, offset, limit)
+	downs, total, err := api.staticDB.DownloadsByUser(req.Context(), *u, offset, limit)
 	if err != nil {
 		WriteError(w, err, http.StatusInternalServerError)
 	}
-	WriteJSON(w, ups)
+	response := database.DownloadsResponseDTO{
+		Items:      downs,
+		Offset:     offset,
+		PageSize:   limit,
+		TotalCount: total,
+	}
+	WriteJSON(w, response)
 }
 
 // trackUploadHandler registers a new upload in the system.
@@ -160,8 +179,8 @@ func (api *API) trackDownloadHandler(w http.ResponseWriter, req *http.Request, p
 }
 
 // fetchOffset extracts the offset from the params and validates its value.
-func fetchOffset(ps httprouter.Params) (int, error) {
-	offset, _ := strconv.Atoi(ps.ByName("offset"))
+func fetchOffset(form url.Values) (int, error) {
+	offset, _ := strconv.Atoi(form.Get("offset"))
 	if offset < 0 {
 		return 0, errors.New("Invalid offset")
 	}
@@ -169,8 +188,8 @@ func fetchOffset(ps httprouter.Params) (int, error) {
 }
 
 // fetchLimit extracts the offset from the params and validates its value.
-func fetchLimit(ps httprouter.Params) (int, error) {
-	limit, _ := strconv.Atoi(ps.ByName("limit"))
+func fetchLimit(form url.Values) (int, error) {
+	limit, _ := strconv.Atoi(form.Get("limit"))
 	if limit < 0 {
 		return 0, errors.New("Invalid limit")
 	}
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -39,8 +39,11 @@ func (api *API) userUploadsHandler(w http.ResponseWriter, req *http.Request, ps 
 		WriteError(w, err, http.StatusInternalServerError)
 		return
 	}
-	offset, _ := strconv.Atoi(ps.ByName("offset"))
-	limit, _ := strconv.Atoi(ps.ByName("limit"))
+	offset, err1 := fetchOffset(ps)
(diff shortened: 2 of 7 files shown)


kwypchlo (dismissed):

api/cookie.go
```diff
@@ -36,8 +36,8 @@ var (
 // writeCookie is a helper function that writes the given JWT token as a
 // secure cookie.
 func writeCookie(w http.ResponseWriter, token string, exp int64) error {
-	if exp <= 0 || time.Unix(exp, 0).Before(time.Now()) {
-		exp = time.Now().Unix()
+	if exp <= 0 || time.Unix(exp, 0).Before(time.Now().UTC()) {
```
peterjanbrone, line 39: why is exp an `int64` and not a `uint64`?

api/cookie.go
```diff
@@ -36,8 +36,8 @@ var (
...
+		exp = time.Now().UTC().Unix()
```
peterjanbrone, line 40: not sure why we care the expiry is before the current time if we're going to set it to the current time anyway? those would immediately expire right?

database/download.go
```diff
@@ -7,15 +7,24 @@ import (
...
-	Timestamp primitive.Timestamp `bson:"timestamp" json:"timestamp"`
+	ID        primitive.ObjectID `bson:"_id,omitempty" json:"_id"`
+	UserID    primitive.ObjectID `bson:"user_id,omitempty" json:"user_id"`
+	SkylinkID primitive.ObjectID `bson:"skylink_id,omitempty" json:"skylink_id"`
+	Timestamp time.Time          `bson:"timestamp" json:"timestamp"`
+}
+
+// DownloadResponseDTO is the DTO we send as response to the caller.
+type DownloadResponseDTO struct {
+	ID        string    `bson:"string_id" json:"id"`
```
peterjanbrone, line 23: why is this annotated with `bson`? This has nothing to do with mongo right?

database/download.go
```diff
@@ -52,48 +61,84 @@ func (db *DB) DownloadCreate(ctx context.Context, user User, skylink Skylink) (*
 }
 
 // DownloadsBySkylink fetches all downloads of this skylink
-func (db *DB) DownloadsBySkylink(ctx context.Context, skylink Skylink, offset, limit int) ([]Download, error) {
+func (db *DB) DownloadsBySkylink(ctx context.Context, skylink Skylink, offset, limit int) ([]DownloadResponseDTO, error) {
```
peterjanbrone, line 64: This is not usually how list functions work, how do you page with this? Usually what you return is an object that contains `items` and `count`. Most mongo drivers and libraries have a `findAndCount` for this reason. Allowing you to query using an `offset` and `limit` and still get the overall count.

database/download.go
```diff
@@ -52,48 +61,84 @@ func (db *DB) DownloadCreate(ctx context.Context, user User, skylink Skylink) (*
...
+//		{ $replaceRoot: { newRoot: { $mergeObjects: [ { $arrayElemAt: [ "$fromSkylinks", 0 ] }, "$$ROOT" ] } } },
+//		{ $project: { fromSkylinks: 0 } },
+//		{ $addFields: { string_id: { $toString: "$_id" } } }
+//	])
+//
+// This query will get all downloads by the current user, skip $skip of them
+// and then fetch $limit of them, allowing us to paginate. It will then
+// join with the `skylinks` collection in order to fetch some additional
+// data about each download. The last line converts the [12]byte `_id` to hex,
+// so we can easily handle it in JSON.
```
peterjanbrone, line 105: The driver should handle this for you actually, pretty sure it would call `toHexString()` on the `ObjectId`. But I don't have that much against the `$toString` you did here either.

database/download.go
```diff
@@ -52,48 +61,84 @@ func (db *DB) DownloadCreate(ctx context.Context, user User, skylink Skylink) (*
...
+func (db *DB) downloadsBy(ctx context.Context, matchStage bson.D, offset, limit int) ([]DownloadResponseDTO, error) {
+	if offset < 0 {
```
peterjanbrone, line 107: I'm not really sure why we're handling these negative cases. If the `offset` and/or `limit` query string parameters are invalid, the user should be made aware then and there, well before executing this function and through a 400 status code. This helper should just build the aggregation pipeline imo and not do validation of the input params.

database/upload.go
```diff
@@ -7,15 +7,24 @@ import (
...
-	Timestamp primitive.Timestamp `bson:"timestamp" json:"timestamp"`
+	ID        primitive.ObjectID `bson:"_id,omitempty" json:"_id"`
+	UserID    primitive.ObjectID `bson:"user_id,omitempty" json:"user_id"`
+	SkylinkID primitive.ObjectID `bson:"skylink_id,omitempty" json:"skylink_id"`
+	Timestamp time.Time          `bson:"timestamp" json:"timestamp"`
+}
+
+// UploadResponseDTO is the DTO we send as response to the caller.
+type UploadResponseDTO struct {
+	ID        string    `bson:"string_id" json:"id"`
```
peterjanbrone, line 23: same remark with the `bson`

database/upload.go
```diff
@@ -52,48 +61,84 @@ func (db *DB) UploadCreate(ctx context.Context, user User, skylink Skylink) (*Up
...
+//		{ $project: { fromSkylinks: 0 } },
+//		{ $addFields: { string_id: { $toString: "$_id" } } }
+//	])
+//
+// This query will get all uploads by the current user, skip $skip of them
+// and then fetch $limit of them, allowing us to paginate. It will then
+// join with the `skylinks` collection in order to fetch some additional
+// data about each upload. The last line converts the [12]byte `_id` to hex,
+// so we can easily handle it in JSON.
+func (db *DB) uploadsBy(ctx context.Context, matchStage bson.D, offset, limit int) ([]UploadResponseDTO, error) {
```
peterjanbrone, line 106: This code is completely identical to `downloadsBy` except for the return param type and the collection on which we call the aggregation pipeline right? When dealing with aggregation pipelines I usually devise helper functions that help construct the pipeline itself, rather than do the entire thing. That way you end up with things like:

```
const pipeline = getUploadsPipeline(inputA, inputB)
const uploadsDB = await this.mongo.getCollection('uploadsDB') 
return await uploadsDB.aggregate(pipeline)
```

🤷‍♂️  might clean up the code a bit I don't know

peterjanbrone: these helpers look very complex, with the large doc blocks and comments and stuff, but it's "just a list function"

api/cookie.go
ro-tex: Go's `time.Now()` returns int64 and `time.Unix` takes int64. I didn't see a reason to convert twice.

ro-tex: Hm, I guess the use of `int64` is there to allow referenced to times before Jan 1st, 1970.
Running `fmt.Println(time.Unix(-1000000, 0))` prints out `1969-12-20 11:13:20 +0100 CET`

api/cookie.go
ro-tex: Hm, you're right. I'm not sure what I was thinking back then. I'll remove the check.

database/download.go
ro-tex: It does. When we fetch data from mongo we want to deserialize it into this DTO before we return it to the handler. We need to know how to map the fields, so `cursor.All` can populate that for us.

Is there a better way to do it?

database/upload.go
ro-tex: same answer. I don't think we should keep duplicating these comments when the cases are identical.

database/download.go
peterjanbrone: wasn't aware stuff got deserialized into these structs, 
figured you were populating them but I missed:
```
downloads := make([]DownloadResponseDTO, 0)
err = c.All(ctx, &downloads)		err = c.All(ctx, &downloads)
```

kwypchlo (dismissed):

database/download.go
ro-tex: Oh, you're right! I was trying before with `int64` instead of `time.Time` and it didn't work but it does work with `time.Time`, so I'm getting rid of the `$toString`. Thanks!

database/download.go
ro-tex: Good point, I should validate in the handler.

database/upload.go
ro-tex: I was thinking about doing that but it seemed like it could bite me in the future if I wanted to do different things with the uploads and downloads. Well, YAGNI, I tried your helper approach and the code looks much cleaner. Let's go with that until we need to separate them.

kwypchlo (approved):

database/download.go
ro-tex: I changed the response to look like this:
```json
{
    "items": [
        {
            "id": "60227e000fb1b80cf4f795f4",
            "skylink": "HELLO__5MJyl1DKyfbuQMY_XBOBC9bR7idiU6isp6LXxEw",
            "name": "This is my HELLO skyfile!",
            "size": 132412341234,
            "downloadedOn": "2021-02-09T12:20:16.869Z"
        }
    ],
    "offset": 1,
    "pageSize": 1,
    "totalCount": 7
}
```

Unfortunately, I couldn't manage to do it within the same aggregation pipeline. I'm not sure if it's even possible or desirable. I'm doing it with a separate aggregation call that just filters the results by the same `matchStage` criteria and then counts them.

database/download.go
```diff
@@ -51,52 +68,42 @@ func (db *DB) DownloadCreate(ctx context.Context, user User, skylink Skylink) (*
...
 	}
-	c, err := db.staticDownloads.Find(ctx, filter, &opts)
+	pipeline := generateUploadsDownloadsPipeline(matchStage, offset, limit)
+	c, err := db.staticDownloads.Aggregate(ctx, pipeline)
 	if err != nil {
-		return nil, err
+		return nil, 0, err
 	}
-	downloads := make([]Download, 0)
+	downloads := make([]DownloadResponseDTO, 0)
```
peterjanbrone, line 103: I guess you could initialize this with the count now. Not sure what happens if between the count and the query a document got inserted or removed 🤔

peterjanbrone (approved):

api/handlers.go
```diff
@@ -39,17 +40,29 @@ func (api *API) userUploadsHandler(w http.ResponseWriter, req *http.Request, ps
...
+	limit, err2 := fetchLimit(req.Form)
+	if err = errors.Compose(err1, err2); err != nil {
+		WriteError(w, err, http.StatusBadRequest)
+	}
+	ups, total, err := api.staticDB.UploadsByUser(req.Context(), *u, offset, limit)
 	if err != nil {
 		WriteError(w, err, http.StatusInternalServerError)
 	}
-	WriteJSON(w, ups)
+	response := database.UploadsResponseDTO{
```
peterjanbrone, line 55: I think we need to return the `limit` and the `offset`, the user requested those params so no need to return them.
Also if we call the param `limit`, I wouldn't call the return variable `PageSize`?

database/database.go
```diff
@@ -208,3 +211,65 @@ func ensureCollection(ctx context.Context, db *mongo.Database, collName string)
...
 	return coll, nil
 }
+
+// generateUploadsDownloadsPipeline generates a mongo pipeline for transforming
+// an `Upload` or `Download` struct into the respective
+// `<Up/Down>loadResponseDTO` struct.
+//
+// The Mongo query we want to ultimately execute is:
+//	db.downloads.aggregate([
+//		{ $match: { "user_id": ObjectId("5fda32ef6e0aba5d16c0d550") }},
```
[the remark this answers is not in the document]
ro-tex: Yes, we have indexes on `user_id` and `skylink_id` for both `uploads` and `downloads` collections.

What is the special implication of `$lookup`?

database/database.go
```diff
@@ -208,3 +211,65 @@ func ensureCollection(ctx context.Context, db *mongo.Database, collName string)
...
+	}
+	projectStage := bson.D{{"$project", bson.D{{"fromSkylinks", 0}}}}
+	return mongo.Pipeline{matchStage, skipStage, limitStage, lookupStage, replaceStage, projectStage}
+}
+
+// count returns the number of documents in the given collection that match the
+// given matchStage.
+func count(ctx context.Context, coll *mongo.Collection, matchStage bson.D) (int, error) {
+	pipeline := mongo.Pipeline{matchStage, bson.D{{"$count", "count"}}}
+	c, err := coll.Aggregate(ctx, pipeline)
```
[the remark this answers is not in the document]
ro-tex: There is the `collection.CountDocuments()` which does that but you're right, it uses the aggregation pipeline underneath. And if I reuse the `matchStage` the entire flow becomes quite nice, IMO.

database/download.go
ro-tex: Yes, nice catch. A tiny optimisation but a free one. Except that I shouldn't init it with count (which is the total count of objects in the DB) but with limit.

If the number of records in the DB changes between the two DB requests we're going to have a bit of a dicrepancy, I guess, i.e. `count: 5, items: <arr with len 6>` or something like that. It doesn't seem like a critical thing in this particular case. We could counter that with a transaction or something like that but I'd rather keep it simple for the moment.

api/handlers.go
ro-tex: `limit` renamed to `pageSize` for consistency.

If @kwypchlo doesn't need the `pafeSize` and `offset` in the response, I'll remove them.

database/database.go
```diff
@@ -208,3 +211,65 @@ func ensureCollection(ctx context.Context, db *mongo.Database, collName string)
...
+			{"localField", "skylink_id"}, // field in the downloads collection
+			{"foreignField", "_id"},      // field in the skylinks collection
+			{"as", "fromSkylinks"},
+		}},
+	}
+	replaceStage := bson.D{
+		{"$replaceRoot", bson.D{
+			{"newRoot", bson.D{
+				{"$mergeObjects", bson.A{
+					bson.D{{"$arrayElemAt", bson.A{"$fromSkylinks", 0}}}, "$$ROOT"},
```
m-cat, line 254: The double `$$` is intended?

m-cat (approved):
I'm not familiar with Mongo, but looks good from a Go perspective.

database/database.go
ro-tex: Yes. No clue why, tbh. I am balancing the "dig deeper, so you understand why" with "how do I just...". :)

kwypchlo (approved):

PR #30, opened 2021-02-11
Track registry operations + login/logout

ro-tex:


api/routes.go
```diff
@@ -3,19 +3,23 @@ package api
...
 	"context"
 	"net/http"
-	"time"
 
 	"github.com/julienschmidt/httprouter"
 )
 
 // buildHTTPRoutes registers all HTTP routes and their handlers.
 func (api *API) buildHTTPRoutes() {
+	api.staticRouter.POST("/login", api.loginHandler)
```
kwypchlo, line 12: This endpoint also requires validate right? It should validate whether the authorisation header is provided and valid before logging user in.

```suggestion
	api.staticRouter.POST("/login", api.validate(api.loginHandler))
```

api/routes.go
```diff
@@ -3,19 +3,23 @@ package api
...
+	api.staticRouter.POST("/logout", api.validate(api.logoutHandler))
```
kwypchlo, line 13: This endpoint on the other hand should not check whether kratos session is available - actually, it will already be deleted by the time we call /logout on accounts.

```suggestion
	api.staticRouter.POST("/logout", api.logoutHandler)
```

kwypchlo (changes requested):

api/routes.go
ro-tex: The endpoint does validate internally. It basically repeats the steps `validate` does but keeps the string representation of the token in order to put it in the cookie. I've removed the `validate` here in order to avoid doing it twice.

api/routes.go
ro-tex: This validation isn't validating the Kratos's session, it's only validating either the bearer token from the headers or the cookie. If they are both invalid there is nothing for this endpoint to do.

kwypchlo: does it validate whether the authorisation header exists at all ? because when hitting this endpoint there will be no authorisation header

api/routes.go
kwypchlo: you shouldn't do it in the endpoint again if you validate it here I guess but maybe I didn't read thorough enough, I'll leave it to a second reviewer

api/routes.go
ro-tex: The `validate` method checks first the header and then, if it's not there, checks the cookie.

My idea here is that:
1. If we have a valid token in the header then we'll remove the cookie. You're saying that this won't happen. That's fine.
2. If we have a valid cookie then we'll remove it. This is our happy case.
3. If we have neither of those then there will be nothing to do.

The only case that's not supported here is if you have a valid cookie but you also pass an invalid authorization header, e.g. an expired one.

If you think that we don't need to do any validation whether the caller is allowed to remove the cookie (it's a cookie on their machine, so I guess that's OK?) then I'll remove the validation and any call to this endpoint will result in a removed cookie.

m-cat (dismissed):

ro-tex:
I cleaned up the history a bit, so we can keep the thought process but remove the useless experimentation.
I'll open a new PR.

PR #34, opened 2021-02-22
Implement user statistics endpoint

ro-tex:
* Introduced a new `/user/stats` endpoint that reports on various user stats, such as storage and bandwidth used.
* Used storage is no longer stored with the user's record, making it moot for the fetcher to update the user.
* Increment download size of the most recent download instead of creating a new one.

Most of these are pretty straightforward, with the exception of the incremental downloads. What is happening there is that when `accounts` receives a request to track a new download, before it does so it would check whether there is a recent download by the same user of the same skylink. "Recent" is adjustable and is currently set to 10 minutes. If there is such a recent upload, we won't create a new one and instead we are just going to increment the size of the recent download with the newly downloaded amount of bytes. This will lead to streaming video downloads being aggregated into one. That will keep user's download history in line with what they might reasonable expect from it.

### DIFF
--- a/database/download.go
+++ b/database/download.go
@@ -2,13 +2,12 @@ package database
 
 import (
 	"context"
-	"fmt"
 	"time"
 
 	"gitlab.com/NebulousLabs/errors"
 	"go.mongodb.org/mongo-driver/bson"
 	"go.mongodb.org/mongo-driver/bson/primitive"
-	"go.mongodb.org/mongo-driver/mongo"
+	"go.mongodb.org/mongo-driver/mongo/options"
 )
 
 const (
@@ -24,8 +23,8 @@ type Download struct {
 	UserID    primitive.ObjectID `bson:"user_id,omitempty" json:"userId"`
 	SkylinkID primitive.ObjectID `bson:"skylink_id,omitempty" json:"skylinkId"`
 	Bytes     int64              `bson:"bytes" json:"bytes"`
-	Created   time.Time          `bson:"timestamp" json:"timestamp"`
-	Updated   time.Time          `bson:"updated" json:"-"`
+	CreatedAt time.Time          `bson:"created_at" json:"timestamp"`
+	UpdatedAt time.Time          `bson:"updated_at" json:"-"`
 }
 
 // DownloadResponseDTO  is the representation of a download we send as response
@@ -71,7 +70,7 @@ func (db *DB) DownloadCreate(ctx context.Context, user User, skylink Skylink, by
 	// Check if there exists a download of this skylink by this user, updated
 	// within the DownloadUpdateWindow and keep updating that, if so.
 	down, err := db.DownloadRecent(ctx, skylink.ID)
-	if err == nil && down.Updated.Add(DownloadUpdateWindow).After(time.Now().UTC()) {
+	if err == nil && down.UpdatedAt.Add(DownloadUpdateWindow).After(time.Now().UTC()) {
 		// We found a recent download of this skylink. Let's update it.
 		return db.DownloadIncrement(ctx, down, bytes)
 	}
@@ -82,8 +81,8 @@ func (db *DB) DownloadCreate(ctx context.Context, user User, skylink Skylink, by
 		UserID:    user.ID,
 		SkylinkID: skylink.ID,
 		Bytes:     bytes,
-		Created:   time.Now().UTC(),
-		Updated:   time.Now().UTC(),
+		CreatedAt: time.Now().UTC(),
+		UpdatedAt: time.Now().UTC(),
 	}
 	_, err = db.staticDownloads.InsertOne(ctx, down)
 	return err
@@ -136,25 +135,17 @@ func (db *DB) downloadsBy(ctx context.Context, matchStage bson.D, offset, pageSi
 
 // DownloadRecent returns the most recent download of the given skylink.
 func (db *DB) DownloadRecent(ctx context.Context, skylinkId primitive.ObjectID) (*Download, error) {
-	matchStage := bson.D{{"$match", bson.D{{"skylink_id", skylinkId}}}}
-	sortStage := bson.D{{"$sort", bson.D{
-		{"updated", -1},
-		{"timestamp", -1},
-	}}}
-	limitStage := bson.D{{"$limit", 1}}
-	pipeline := mongo.Pipeline{matchStage, sortStage, limitStage}
-
-	c, err := db.staticDownloads.Aggregate(ctx, pipeline)
-	if err != nil {
+	filter := bson.D{{"skylink_id", skylinkId}}
+	opts := options.FindOneOptions{
+		Sort: bson.D{{"updated_at", -1}},
+	}
+	sr := db.staticDownloads.FindOne(ctx, filter, &opts)
+	if err := sr.Err(); err != nil {
+		// This includes the "no documents found" case.
 		return nil, err
 	}
-	if ok := c.Next(ctx); !ok {
-		// No results found. This is expected.
-		return nil, errors.New(fmt.Sprintf("no downloads found for skylink with id %v", skylinkId))
-	}
 	var d Download
-	err = c.Decode(&d)
-	if err != nil {
+	if err := sr.Decode(&d); err != nil {
 		return nil, errors.AddContext(err, "failed to parse value from DB")
 	}
 	return &d, nil
--- a/database/download.go
+++ b/database/download.go
@@ -2,11 +2,20 @@ package database
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	"gitlab.com/NebulousLabs/errors"
 	"go.mongodb.org/mongo-driver/bson"
 	"go.mongodb.org/mongo-driver/bson/primitive"
+	"go.mongodb.org/mongo-driver/mongo"
+)
+
+const (
+	// DownloadUpdateWindow defines a time window during which instead of
+	// creating a new download record for the given skylink, we'll update the
+	// previous one, as long as it has been updated within the window.
+	DownloadUpdateWindow = 10 * time.Minute
 )
 
 // Download describes a single download of a skylink by a user.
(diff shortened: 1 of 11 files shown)


test/user_test.go
```diff
@@ -28,63 +27,167 @@ func TestUserBandwidth(t *testing.T) {
...
-	// Register a download.
+	// Register a small download.
 	smallDownload := int64(1 + fastrand.Intn(4*database.MiB))
 	_, err = db.DownloadCreate(ctx, *u, *skylinkSmall, smallDownload)
 	if err != nil {
 		t.Fatal("Failed to download.", err)
 	}
-	// Check bandwidth.
-	details, err = db.UserDetails(ctx, u)
+	expectedDownloadBandwidth += database.BandwidthDownloadCost(smallDownload)
```
peterjanbrone, line 81: maybe the `database` package is not the best location for this helper

database/download.go
```diff
@@ -50,25 +60,33 @@ func (db *DB) DownloadByID(ctx context.Context, id primitive.ObjectID) (*Downloa
...
 	}
 	if skylink.ID.IsZero() {
-		return nil, errors.New("invalid skylink")
+		return errors.New("invalid skylink")
+	}
+
+	// Check if there exists a download of this skylink by this user, updated
+	// within the DownloadUpdateWindow and keep updating that, if so.
+	down, err := db.DownloadRecent(ctx, skylink.ID)
+	if err == nil && down.Updated.Add(DownloadUpdateWindow).After(time.Now().UTC()) {
```
peterjanbrone, line 74: I guess this is fine but you could add the update window to the query imo

database/download.go
```diff
@@ -50,25 +60,33 @@ func (db *DB) DownloadByID(ctx context.Context, id primitive.ObjectID) (*Downloa
...
+		// We found a recent download of this skylink. Let's update it.
+		return db.DownloadIncrement(ctx, down, bytes)
 	}
-	up := Download{
+
+	// We couldn't find a recent download of this skylink, updated within
```
peterjanbrone, line 79: Should this be atomic? I think you'll quickly run into issues here where two concurrent calls to this endpoint wreak havoc in your database? Erroring out one of the two calls and such most probably if your database has any unqiue constraints.

We have unique constraints on all of these collections right? E.g. in downloads an entry should be unique on user and skylink for example.

database/download.go
```diff
@@ -50,25 +60,33 @@ func (db *DB) DownloadByID(ctx context.Context, id primitive.ObjectID) (*Downloa
...
 		UserID:    user.ID,
 		SkylinkID: skylink.ID,
 		Bytes:     bytes,
-		Timestamp: time.Now().UTC(),
-	}
-	ior, err := db.staticDownloads.InsertOne(ctx, up)
-	if err != nil {
-		return nil, err
+		Created:   time.Now().UTC(),
+		Updated:   time.Now().UTC(),
```
peterjanbrone, line 86: Are you updating this field on update? I think but am not 100% sure you are not so `Updated` will always remain the same as `Created`

database/user.go
```diff
@@ -70,14 +68,21 @@ type (
...
 		SubscribedUntil time.Time          `bson:"subscribed_until" json:"subscribedUntil"`
 	}
-	// UserDetails builds on top of User and provides additional information,
-	// typically more expensive to fetch than a single DB query.
-	UserDetails struct {
-		User
-		BandwidthUsed int64 `bson:"-" json:"bandwidthUsed"`
+	// UserStats contains statistical information about the user.
+	UserStats struct {
+		StorageUsed        int64 `bson:"-" json:"storageUsed"`
```
peterjanbrone, line 75: in siad we have lowercase characters for all json fields, I kind of like that, should we be consistent here?

database/user.go
```diff
@@ -70,14 +68,21 @@ type (
...
+		NumRegReads        int64 `bson:"-" json:"numRegReads"`
```
peterjanbrone, line 76: what's the deal with the `bson:"-"`?

database/user.go
```diff
@@ -70,14 +68,21 @@ type (
...
+		NumRegWrites       int64 `bson:"-" json:"numRegWrites"`
+		NumUploads         int   `bson:"-" json:"numUploads"`
+		NumDownloads       int   `bson:"-" json:"numDownloads"`
+		TotalUploadsSize   int64 `bson:"-" json:"totalUploadsSize"`
+		TotalDownloadsSize int64 `bson:"-" json:"totalDownloadsSize"`
+		BandwidthUploads   int64 `bson:"-" json:"bwUploads"`
```
peterjanbrone, line 82: not a fan of the abbreviation in the fields here

database/download.go
```diff
@@ -2,11 +2,20 @@ package database
...
 	"go.mongodb.org/mongo-driver/bson"
 	"go.mongodb.org/mongo-driver/bson/primitive"
+	"go.mongodb.org/mongo-driver/mongo"
+)
+
+const (
+	// DownloadUpdateWindow defines a time window during which instead of
+	// creating a new download record for the given skylink, we'll update the
+	// previous one, as long as it has been updated within the window.
+	DownloadUpdateWindow = 10 * time.Minute
```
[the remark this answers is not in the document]
ro-tex: Yes, this is very arbitrary. I'm very much open to changing this to anything that makes more sense. 

I chose the period because it fits the use case of watching a movie and stopping for a quick bio break or to get a snack.

database/download.go
```diff
@@ -15,7 +24,8 @@ type Download struct {
 	UserID    primitive.ObjectID `bson:"user_id,omitempty" json:"userId"`
 	SkylinkID primitive.ObjectID `bson:"skylink_id,omitempty" json:"skylinkId"`
 	Bytes     int64              `bson:"bytes" json:"bytes"`
-	Timestamp time.Time          `bson:"timestamp" json:"timestamp"`
+	Created   time.Time          `bson:"timestamp" json:"timestamp"`
```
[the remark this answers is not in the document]
ro-tex: Yes, you're right. I initially left it like this for development purposes (I didn't want to have to go over the records in the DB and update them manually) and then I forgot to change it.

I'll rename the `bson` fields to match the properties but for the moment I won't touch the `json` ones as that will change the API and that's not needed right now because we're not reporting the `updatedAt` time there.

database/download.go
```diff
@@ -115,3 +133,42 @@ func (db *DB) downloadsBy(ctx context.Context, matchStage bson.D, offset, pageSi
...
+func (db *DB) DownloadRecent(ctx context.Context, skylinkId primitive.ObjectID) (*Download, error) {
+	matchStage := bson.D{{"$match", bson.D{{"skylink_id", skylinkId}}}}
+	sortStage := bson.D{{"$sort", bson.D{
+		{"updated", -1},
+		{"timestamp", -1},
+	}}}
+	limitStage := bson.D{{"$limit", 1}}
+	pipeline := mongo.Pipeline{matchStage, sortStage, limitStage}
+
+	c, err := db.staticDownloads.Aggregate(ctx, pipeline)
```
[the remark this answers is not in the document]
ro-tex: Changed to `FindOne`.

database/download.go
```diff
@@ -115,3 +133,42 @@ func (db *DB) downloadsBy(ctx context.Context, matchStage bson.D, offset, pageSi
...
+	var d Download
+	err = c.Decode(&d)
+	if err != nil {
+		return nil, errors.AddContext(err, "failed to parse value from DB")
+	}
+	return &d, nil
+}
+
+// DownloadIncrement increments the size of the download by additionalBytes.
+func (db *DB) DownloadIncrement(ctx context.Context, d *Download, additionalBytes int64) error {
```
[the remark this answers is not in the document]
ro-tex: All sizes are `int64` for consistency. This is in line with the approach in the standard library (`os.FileInfo().Size()` returns `int64` despite files always having non-negative sizes.)

database/user.go
```diff
@@ -397,49 +396,57 @@ func (db *DB) userDownloadBandwidth(ctx context.Context, id primitive.ObjectID,
...
 	for c.Next(ctx) {
 		if err = c.Decode(&result); err != nil {
-			return 0, errors.AddContext(err, "failed to decode DB data")
+			err = errors.AddContext(err, "failed to decode DB data")
+			return
 		}
-		bandwidth += BandwidthDownloadCost(result.Size)
+		count++
+		totalSize += result.Size
+		totalBandwidth += BandwidthDownloadCost(result.Size)
```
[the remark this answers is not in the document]
ro-tex: I wasn't sure how to name this. Both on upload and download we perform actions which use more bandwidth than the raw size of the file. I was on the fence between `cost` and `used` but `used` sounded like something we've recorded but that's not the case. That's why I went with `cost`.

What would make more sense to you?

test/user_test.go
ro-tex: Moved all bandwidth-related costs, vars and functions to a new module called `skynet`.

database/download.go
ro-tex: Good suggestion. While implementing this I also noticed that the `updated_at` was never actually updated, so that's fixed as well.

database/download.go
ro-tex: I'm not sure I'm following.

There is no unique constraint on user_id+skylink_id, so if we add a new download at this point it will get its own download_id with the same user_id and skylink_id as another one, if added in-between these calls.

In the race case where the browser sends two GETs very close together and:
1. we check the DB on the first GET, we see no record, we start creating it...
2. we check the DB on the second GET, we see no record (the first one is still pending creation), we start creating it...

We'll end up with two separate download records. This will pollute the user's download history slightly but it won't affect them in any other way because each created download will have its real size (as reported by nginx), so the user won't be charged for bandwidth they didn't consume.

database/download.go
ro-tex: Correct, I wasn't. Fixed.

database/user.go
ro-tex: The name convention of the JSON side of things is requested by Karol in order to be in line with what's usually used in JS. Personally, I don't mind either way. If you have a strong opinion on this one you can ping @kwypchlo and you can hash it out. I'm happy to follow any convention.

database/user.go
ro-tex: I don't really need them. I just got used to the duality of bson/json and tunnel-visioned on this one. Removed.

database/user.go
ro-tex: Not strictly needed here but it's still fewer bytes to send over the wire with `bandwidth` and `number` repeated over and over. I don't feel strongly about this, if you don't like it I can change to full words.

m-cat (approved):

peterjanbrone (approved):

PR #36, opened 2021-02-24
Integrate Stripe

ro-tex:
This PR adds Stripe integration for accounts. The way that works is by implementing a webhook on which we listen for evens sent by Stripe. Those events tell us when a user's subscription changes. When that happens we adjust the user's tier accordingly.

Details:
* All JWT logic moved to a separate package.
* We now store some user details locally (names and email), so we can identify user accounts more easily. We might also use them for other purposes later on.


api/stripe.go
```diff
@@ -0,0 +1,291 @@
...
+	"github.com/stripe/stripe-go/v71/sub"
+	"gitlab.com/NebulousLabs/errors"
+)
+
+var (
+	// StripeTestMode tells us whether to use Stripe's test mode or prod mode
+	// plan and price ids.
+	StripeTestMode = false
+
+	// TODO These should be in the DB.
```
ro-tex, line 29: This TODO is not forgotten, we just don't need to handle it right now.

main.go
```diff
@@ -40,6 +43,15 @@ var (
 	// envPortal holds the name of the environment variable for the portal to
 	// use to fetch skylinks.
 	envPortal = "PORTAL_URL"
+	// envKratosAddr hold the name of the environment variable for Kratos's
+	// address. Defaults to "kratos:4433".
+	envKratosAddr = "KRATOS_ADDR"
+	// envKratosAddr hold the name of the environment variable for Oathkeeper's
```
m-cat, line 49:
```suggestion
	// envOathkeeperAddr hold the name of the environment variable for Oathkeeper's
```

test/database_test.go
```diff
@@ -84,7 +84,7 @@ func TestDatabase_UserByID(t *testing.T) {
 	}
 }
 
-// TestDatabase_UserUpdate ensures UserUpdate works as expected.
+// TestDatabase_UserUpdate ensures UserSave works as expected.
 func TestDatabase_UserUpdate(t *testing.T) {
```
m-cat, line 88:
```suggestion
func TestDatabase_UserSave(t *testing.T) {
```

database/download.go
```diff
@@ -34,7 +34,7 @@ type DownloadResponseDTO struct {
 	Skylink   string    `bson:"skylink" json:"skylink"`
 	Name      string    `bson:"name" json:"name"`
 	Size      uint64    `bson:"size" json:"size"`
-	Timestamp time.Time `bson:"timestamp" json:"downloadedOn"`
+	Timestamp time.Time `bson:"created_at" json:"downloadedOn"`
```
m-cat, line 37:
```suggestion
	CreatedAt time.Time `bson:"created_at" json:"downloadedOn"`
```

m-cat: The `json` doesn't match, either

ro-tex: Yes, we changed course after this was implemented. I'll rename the field in the IDE, rather than merging this, so it changes all the uses.

api/stripe.go
```diff
@@ -0,0 +1,313 @@
...
+	u.Tier = database.TierFree
+	u.SubscribedUntil = time.Time{}
+	u.SubscriptionStatus = ""
+	u.SubscriptionCancelAt = time.Time{}
+	u.SubscriptionCancelAtPeriodEnd = false
+	var latestCreated time.Time
+	var latestSub string
+	// Pick the latest active plan and set the user's tier based on that.
+	subs := it.SubscriptionList().Data
+	for _, subsc := range subs {
```
peterjan, line 211: Why do you loop over all subs and decorate the user potentially multiple times? 
Seems more logical to find the most recent sub and then decorate it once?

I'm also slightly confused on why we reset the user's subscription fields to default,
before assigning it the new ones.

api/stripe.go
```diff
@@ -0,0 +1,313 @@
...
+	}
+	for _, subsc := range subs {
+		if subsc == nil || subsc.ID == latestSub {
+			continue
+		}
+		subsc, err = sub.Cancel(subsc.ID, &p)
+		if err != nil {
+			api.staticLogger.Warnf("Failed to cancel sub with id %s for user %s with Stripe customer id %s. Error: %s", subsc.ID, u.ID.Hex(), s.Customer.ID, err.Error())
+		} else {
+			api.staticLogger.Tracef("Successfullu cancelled sub with id %s for user %s with Stripe customer id %s.", subsc.ID, u.ID.Hex(), s.Customer.ID)
```
peterjan, line 241:
```suggestion
			api.staticLogger.Tracef("Successfully cancelled sub with id %s for user %s with Stripe customer id %s.", subsc.ID, u.ID.Hex(), s.Customer.ID)
```

api/stripe.go
```diff
@@ -0,0 +1,313 @@
...
+		}
+	}
+	api.staticLogger.Tracef("Subscribed user id %s, tier %d, until %s.", u.ID, u.Tier, u.SubscribedUntil.String())
```
peterjan, line 244: (nit) logging before success

api/stripe.go
```diff
@@ -0,0 +1,313 @@
...
+		}
+		sPrices = append(sPrices, sp)
+	}
+	api.WriteJSON(w, sPrices)
+}
+
+// readStripeEvent reads the event from the request body and verifies its
+// signature.
+func readStripeEvent(w http.ResponseWriter, req *http.Request) (*stripe.Event, int, error) {
+	const MaxBodyBytes = int64(65536)
```
peterjan, line 172: (nit) define up top I guess

peterjan (approved):
LGTM 👍  decided not to add too many nits or suggestions as I think the overall think just about works now right

api/stripe.go
```diff
@@ -0,0 +1,313 @@
...
+		if err != nil {
+			api.staticLogger.Debugln("Failed to process sub:", err)
+		}
+		api.WriteSuccess(w)
+		return
+	}
+
+	// Here we handle the entire class of subscription_schedule events.
+	// See https://stripe.com/docs/api/subscription_schedules/object
+	if strings.Contains(event.Type, "subscription_schedule") {
```
m-cat, line 115: q: Should this be `custom.subscription_schedule` to match `custom.subscription` above?

m-cat (approved):

api/stripe.go
ro-tex: Those are two different groups of events - `customer` and `subscription_schedule` (https://stripe.com/docs/api/events/types). Both can be sent on subscription update, so I'm listening to both.

api/stripe.go
ro-tex: The main goal was to avoid the `else` in `if <has sub> <decorate with sub> else <decorate with free>`. But I don't mind changing it. Will F/U in #38.

PR #38, opened 2021-03-15
Endpoints for speed limits and unpinning files

### LINKED ISSUE #32: Allow users to unpin files
Files pinned by premium users are pinned permanently and their contracts will be auto-renewed by the portal. They count towards the `storage used` quota of the user.

We should allow users to remove the permanent status of files they upload and have them removed from their `storage used` quota. This would inform the portal to not renew the file contract once it expires.

ro-tex:
## Overview

This PR introduces three new endpoint:
* `GET /user/limits` returns the speed limits for the user based on their account tier. It takes care to downgrade the speed of the user if they exceed their quotas.
* `DELETE /skylink/:skylink` marks the given skylink as unpinned by the current user (all uploads)
* `DELETE /user/upload/:uploadId` marks the given upload as unpinned

~~It also changes the way we calculate storage and bandwidth used by the user to be closer to the real world usage. See https://discord.com/channels/542938080349519882/542939275587485696/821722719069405184 and the subsequent discussion.~~  
Uploaded files now count towards the uploader's quota with their normal size (as seen when the file is on disk) and not based on on the raw size on the Sia network.

### User Limits

Each account tier comes with preset limits that imposes on the user's speed for uploads, downloads, and access to the registry. Users have usage quotas set on the files they upload - total upload size and total number of uploads. Once they exceed any of those, their speed limitations get dropped to those of an anonymous user.

The `GET /user/limit` endpoint fetches the user's tier from the DB, checks if the user exceeds their quotas and returns the respective values, based on a hardcoded map. If we decide to keep this long-term we'll move to a config file. Hardcoding the map seemed like a decent way to save time.

### Unpinning skylinks

The way the `DELETE /skylink/:skylink` endpoint works is by scanning the list of uploads by this user and marking all uploads of this skylink made by this user as "unpinned". `siad` is not notified of this and no other action is taken, so the files are not immediately delete. The same is true about the `DELETE /user/upload/:uploadId` but limited to single upload. The second endpoint was requested by Karol in order to make sure a specific upload in the list is being unpinned, as that makes a lot more sense for the purposes of the UI.

The intended way for this to work is that in the future the portal will scan the list of uploads for each skylink and will only refresh the contracts for those skylinks that are still being pinned by an active subscriber. All unpinned skylinks will be treated as if uploaded by an anonymous user. Once a skylink is unpinned, it no longer counts towards the user's storage quota.

### Storage and bandwidth

* ~~DB will store both the user-facing file size and the actually used storage (base sector * 10 + chunks * 3)~~
* ~~API will report on the actual used storage / 3, so it's more in line with the file sizes users see on their filesystem. We can adjust pricing based on that.~~
* ~~Add an `.env` param that adjusts that modifier.~~
* Multiple uploads of the same content (binary identical uploads) count towards the user's storage quota only once but do count towards the total number of uploads quota multiple times.

## Additional Follow-ups:

* https://github.com/NebulousLabs/skynet-accounts/pull/36#discussion_r595985187

Closes #32

### DIFF
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -248,6 +248,31 @@ func (api *API) userUploadsHandler(w http.ResponseWriter, req *http.Request, _ h
 	api.WriteJSON(w, response)
 }
 
+// userUploadDeleteHandler unpins a single upload by this user.
+func (api *API) userUploadDeleteHandler(w http.ResponseWriter, req *http.Request, ps httprouter.Params) {
+	sub, _, _, err := jwt.TokenFromContext(req.Context())
+	if err != nil {
+		api.WriteError(w, err, http.StatusUnauthorized)
+		return
+	}
+	u, err := api.staticDB.UserBySub(req.Context(), sub, false)
+	if errors.Contains(err, database.ErrUserNotFound) {
+		api.WriteError(w, err, http.StatusNotFound)
+		return
+	}
+	if err != nil {
+		api.WriteError(w, err, http.StatusInternalServerError)
+		return
+	}
+	uid := ps.ByName("uploadId")
+	_, err = api.staticDB.UnpinUpload(req.Context(), uid, *u)
+	if err != nil {
+		api.WriteError(w, err, http.StatusInternalServerError)
+		return
+	}
+	api.WriteSuccess(w)
+}
+
 // userDownloadsHandler returns all downloads made by the current user.
 func (api *API) userDownloadsHandler(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
 	sub, _, _, err := jwt.TokenFromContext(req.Context())
@@ -432,7 +457,7 @@ func (api *API) trackRegistryWriteHandler(w http.ResponseWriter, req *http.Reque
 	api.WriteSuccess(w)
 }
 
-// skylinkDeleteHandler unpins a skylink uploaded by the user.
+// skylinkDeleteHandler unpins all uploads of a skylink uploaded by the user.
 func (api *API) skylinkDeleteHandler(w http.ResponseWriter, req *http.Request, ps httprouter.Params) {
 	sub, _, _, err := jwt.TokenFromContext(req.Context())
 	if err != nil {
@@ -461,7 +486,7 @@ func (api *API) skylinkDeleteHandler(w http.ResponseWriter, req *http.Request, p
 		api.WriteError(w, err, http.StatusInternalServerError)
 		return
 	}
-	_, err = api.staticDB.UnpinUpload(req.Context(), *skylink, *u)
+	_, err = api.staticDB.UnpinUploads(req.Context(), *skylink, *u)
 	if err != nil {
 		api.WriteError(w, err, http.StatusInternalServerError)
 		return
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -95,6 +95,16 @@ func (api *API) userHandler(w http.ResponseWriter, req *http.Request, _ httprout
 	api.WriteJSON(w, u)
 }
 
+// userLimitsHandler returns the speed limits which apply for this user.
+func (api *API) userLimitsHandler(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
+	u := api.userFromRequest(req)
+	if u == nil {
+		api.WriteJSON(w, database.SpeedLimits[database.TierAnonymous])
+		return
+	}
+	api.WriteJSON(w, database.SpeedLimits[u.Tier])
+}
+
 // userStatsHandler returns statistics about an existing user.
 func (api *API) userStatsHandler(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
 	sub, _, _, err := jwt.TokenFromContext(req.Context())
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -447,15 +447,15 @@ func (api *API) skylinkDeleteHandler(w http.ResponseWriter, req *http.Request, p
 		api.WriteError(w, err, http.StatusBadRequest)
 		return
 	}
+	if err != nil {
+		api.WriteError(w, err, http.StatusInternalServerError)
+		return
+	}
 	_, err = api.staticDB.UnpinUploads(req.Context(), *skylink, *u)
 	if err != nil {
 		api.WriteError(w, err, http.StatusInternalServerError)
 		return
 	}
-	//_, remaining, err := api.staticDB.UploadsBySkylink(req.Context(), *skylink, 0, 1)
-	//if remaining == 0 {
-	//// TODO call siad to unpin this
-	//}
 	api.WriteSuccess(w)
 }
 
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -1,6 +1,7 @@
 package api
 
 import (
+	"context"
 	"encoding/json"
 	"io/ioutil"
 	"net/http"
@@ -98,19 +99,9 @@ func (api *API) userHandler(w http.ResponseWriter, req *http.Request, _ httprout
 // userLimitsHandler returns the speed limits which apply for this user.
 func (api *API) userLimitsHandler(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
 	u := api.userFromRequest(req)
-	if u == nil {
+	if u == nil || u.QuotaExceeded {
 		api.WriteJSON(w, database.UserLimits[database.TierAnonymous])
 		return
-	}
-	// Check the user's data usage and if it exceeds the limits of their tier,
-	// return TierAnonymous speed limits.
-	us, err := api.staticDB.UserStats(req.Context(), *u)
-	if err != nil {
-		api.WriteError(w, err, http.StatusInternalServerError)
-		return
-	}
-	if us.TotalUploadsSize > database.UserLimits[u.Tier].Storage || us.NumUploads > database.UserLimits[u.Tier].MaxNumberUploads {
-		api.WriteJSON(w, database.UserLimits[database.TierAnonymous])
 	}
 	api.WriteJSON(w, database.UserLimits[u.Tier])
 }
@@ -351,6 +342,11 @@ func (api *API) trackUploadHandler(w http.ResponseWriter, req *http.Request, ps 
 		}()
 	}
 	api.WriteSuccess(w)
+	// Now that we've returned results to the caller, we can take care of some
+	// administrative details, such as user's quotas check.
+	// Nota that this call is not affected by the request's context, so we use
+	// a separate one.
+	api.checkUserQuotas(context.Background(), u)
 }
 
 // trackDownloadHandler registers a new download in the system.
@@ -492,6 +488,24 @@ func (api *API) skylinkDeleteHandler(w http.ResponseWriter, req *http.Request, p
 		return
 	}
 	api.WriteSuccess(w)
+}
+
+// checkUserQuotas compares the resources consumed by the user to their quotas
+// and sets the QuotaExceeded flag on their account if they exceed any.
+func (api *API) checkUserQuotas(ctx context.Context, u *database.User) {
+	us, err := api.staticDB.UserStats(ctx, *u)
+	if err != nil {
+		api.staticLogger.Infof("Failed to fetch user's stats. UID: %s, err: %s", u.ID.Hex(), err.Error())
+		return
+	}
+	q := database.UserLimits[u.Tier]
+	if us.NumUploads > q.MaxNumberUploads || us.TotalUploadsSize > q.Storage {
+		u.QuotaExceeded = true
+		err = api.staticDB.UserSave(ctx, u)
+		if err != nil {
+			api.staticLogger.Infof("Failed to save user. User: %+v, err: %s", u, err.Error())
+		}
+	}
 }
 
 // fetchOffset extracts the offset from the params and validates its value.
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -264,9 +264,9 @@ func (api *API) userUploadDeleteHandler(w http.ResponseWriter, req *http.Request
 	api.WriteSuccess(w)
 	// Now that we've returned results to the caller, we can take care of some
 	// administrative details, such as user's quotas check.
-	// Nota that this call is not affected by the request's context, so we use
+	// Note that this call is not affected by the request's context, so we use
 	// a separate one.
-	api.checkUserQuotas(context.Background(), u)
+	go api.checkUserQuotas(context.Background(), u)
 }
 
 // userDownloadsHandler returns all downloads made by the current user.
@@ -348,9 +348,9 @@ func (api *API) trackUploadHandler(w http.ResponseWriter, req *http.Request, ps 
 	api.WriteSuccess(w)
 	// Now that we've returned results to the caller, we can take care of some
 	// administrative details, such as user's quotas check.
-	// Nota that this call is not affected by the request's context, so we use
+	// Note that this call is not affected by the request's context, so we use
 	// a separate one.
-	api.checkUserQuotas(context.Background(), u)
+	go api.checkUserQuotas(context.Background(), u)
 }
 
 // trackDownloadHandler registers a new download in the system.
@@ -494,9 +494,9 @@ func (api *API) skylinkDeleteHandler(w http.ResponseWriter, req *http.Request, p
 	api.WriteSuccess(w)
 	// Now that we've returned results to the caller, we can take care of some
 	// administrative details, such as user's quotas check.
-	// Nota that this call is not affected by the request's context, so we use
+	// Note that this call is not affected by the request's context, so we use
 	// a separate one.
-	api.checkUserQuotas(context.Background(), u)
+	go api.checkUserQuotas(context.Background(), u)
 }
 
 // checkUserQuotas compares the resources consumed by the user to their quotas
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -451,7 +451,7 @@ func (api *API) skylinkDeleteHandler(w http.ResponseWriter, req *http.Request, p
 		api.WriteError(w, err, http.StatusInternalServerError)
 		return
 	}
-	_, err = api.staticDB.UnpinUploads(req.Context(), *skylink, *u)
+	_, err = api.staticDB.UnpinUpload(req.Context(), *skylink, *u)
 	if err != nil {
 		api.WriteError(w, err, http.StatusInternalServerError)
 		return
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -262,6 +262,11 @@ func (api *API) userUploadDeleteHandler(w http.ResponseWriter, req *http.Request
 		return
 	}
 	api.WriteSuccess(w)
+	// Now that we've returned results to the caller, we can take care of some
+	// administrative details, such as user's quotas check.
+	// Nota that this call is not affected by the request's context, so we use
+	// a separate one.
+	api.checkUserQuotas(context.Background(), u)
 }
 
 // userDownloadsHandler returns all downloads made by the current user.
@@ -488,6 +493,11 @@ func (api *API) skylinkDeleteHandler(w http.ResponseWriter, req *http.Request, p
 		return
 	}
 	api.WriteSuccess(w)
+	// Now that we've returned results to the caller, we can take care of some
+	// administrative details, such as user's quotas check.
+	// Nota that this call is not affected by the request's context, so we use
+	// a separate one.
+	api.checkUserQuotas(context.Background(), u)
 }
 
 // checkUserQuotas compares the resources consumed by the user to their quotas
@@ -499,8 +509,9 @@ func (api *API) checkUserQuotas(ctx context.Context, u *database.User) {
 		return
 	}
 	q := database.UserLimits[u.Tier]
-	if us.NumUploads > q.MaxNumberUploads || us.TotalUploadsSize > q.Storage {
-		u.QuotaExceeded = true
+	quotaExceeded := us.NumUploads > q.MaxNumberUploads || us.TotalUploadsSize > q.Storage
+	if quotaExceeded != u.QuotaExceeded {
+		u.QuotaExceeded = quotaExceeded
 		err = api.staticDB.UserSave(ctx, u)
 		if err != nil {
 			api.staticLogger.Infof("Failed to save user. User: %+v, err: %s", u, err.Error())
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -412,6 +412,43 @@ func (api *API) trackRegistryWriteHandler(w http.ResponseWriter, req *http.Reque
 	api.WriteSuccess(w)
 }
 
+// skylinkDeleteHandler unpins a skylink uploaded by the user.
+func (api *API) skylinkDeleteHandler(w http.ResponseWriter, req *http.Request, ps httprouter.Params) {
+	sub, _, _, err := jwt.TokenFromContext(req.Context())
+	if err != nil {
+		api.WriteError(w, err, http.StatusUnauthorized)
+		return
+	}
+	u, err := api.staticDB.UserBySub(req.Context(), sub, false)
+	if errors.Contains(err, database.ErrUserNotFound) {
+		api.WriteError(w, err, http.StatusNotFound)
+		return
+	}
+	if err != nil {
+		api.WriteError(w, err, http.StatusInternalServerError)
+		return
+	}
+	sl := ps.ByName("skylink")
+	if !database.ValidSkylinkHash(sl) {
+		api.WriteError(w, errors.New("invalid skylink"), http.StatusBadRequest)
+	}
+	skylink, err := api.staticDB.Skylink(req.Context(), sl)
+	if errors.Contains(err, database.ErrInvalidSkylink) {
+		api.WriteError(w, err, http.StatusBadRequest)
+		return
+	}
+	_, err = api.staticDB.UnpinUploads(req.Context(), *skylink, *u)
+	if err != nil {
+		api.WriteError(w, err, http.StatusInternalServerError)
+		return
+	}
+	//_, remaining, err := api.staticDB.UploadsBySkylink(req.Context(), *skylink, 0, 1)
+	//if remaining == 0 {
+	//// TODO call siad to unpin this
+	//}
+	api.WriteSuccess(w)
+}
+
 // fetchOffset extracts the offset from the params and validates its value.
 func fetchOffset(form url.Values) (int, error) {
 	offset, _ := strconv.Atoi(form.Get("offset"))
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -341,8 +341,7 @@ func (api *API) trackUploadHandler(w http.ResponseWriter, req *http.Request, ps 
 		// Queue the skylink to have its meta data fetched and updated in the DB.
 		go func() {
 			api.staticMF.Queue <- metafetcher.Message{
-				UploaderID: u.ID,
-				SkylinkID:  skylink.ID,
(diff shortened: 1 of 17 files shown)


api/handlers.go
```diff
@@ -412,6 +422,43 @@ func (api *API) trackRegistryWriteHandler(w http.ResponseWriter, req *http.Reque
...
+	if err != nil {
+		api.WriteError(w, err, http.StatusInternalServerError)
+		return
+	}
+	sl := ps.ByName("skylink")
+	if !database.ValidSkylinkHash(sl) {
+		api.WriteError(w, errors.New("invalid skylink"), http.StatusBadRequest)
+	}
+	skylink, err := api.staticDB.Skylink(req.Context(), sl)
+	if errors.Contains(err, database.ErrInvalidSkylink) {
```
m-cat, line 446: We should leave a comment if we're intentionally ignoring other errors

api/routes.go
```diff
@@ -76,3 +82,28 @@ func tokenFromRequest(r *http.Request) (string, error) {
...
+	token, err := jwt.ValidateToken(api.staticLogger, t)
+	if err != nil {
+		return nil
+	}
+	if reflect.ValueOf(token.Claims).Kind() != reflect.ValueOf(jwt2.MapClaims{}).Kind() {
+		return nil
+	}
+	claims := token.Claims.(jwt2.MapClaims)
+	if reflect.ValueOf(claims["sub"]).Kind() != reflect.String {
+		err = errors.New("the token does not contain the sub we expect")
```
m-cat, line 102: This error is ignored?

database/upload.go
```diff
@@ -80,10 +82,64 @@ func (db *DB) UploadsBySkylink(ctx context.Context, skylink Skylink, offset, pag
...
-	matchStage := bson.D{{"$match", bson.D{{"skylink_id", skylink.ID}}}}
+	matchStage := bson.D{{"$match", bson.D{
+		{"skylink_id", skylink.ID},
+		{"unpinned", false},
+	}}}
 	return db.uploadsBy(ctx, matchStage, offset, pageSize)
 }
 
+// UploadsBySkylinkAndUser fetches all uploads of this skylink by this user.
+func (db *DB) UploadsBySkylinkAndUser(ctx context.Context, skylink Skylink, user User) ([]Upload, error) {
```
m-cat, line 93: Is this used anywhere?

database/user.go
```diff
@@ -30,13 +31,44 @@ const (
 	TierPremium80
 	// TierMaxReserved is a guard value that helps us validate tier values.
 	TierMaxReserved
+
+	// bpsToBytesPerSecond is a multiplier to get from 1 bps to X bytes per sec.
+	bpsToBytesPerSecond = 1024 * 1024 / 8
```
m-cat, line 36: What's bps? If it's bits per second then this doesn't seem right.

database/user.go
```diff
@@ -222,7 +271,7 @@ func (db *DB) UserSetStripeId(ctx context.Context, u *User, stripeId string) err
 
 // UserSetTier sets the user's tier to the given value.
 func (db *DB) UserSetTier(ctx context.Context, u *User, t int) error {
-	if t <= TierMinReserved || t >= TierMaxReserved {
+	if t <= TierAnonymous || t >= TierMaxReserved {
```
m-cat, line 274: Should this have changed?

test/upload_test.go
```diff
@@ -70,6 +70,31 @@ func TestUpload_UploadsByUser(t *testing.T) {
 		t.Fatalf("Expected storage used of %d (%d MiB), got %d (%d MiB).",
 			storageUsed, storageUsed/skynet.MiB, stats.StorageUsed, stats.StorageUsed/skynet.MiB)
 	}
+	// Delete the last upload. Expect zero uploads and zero storage used after.
+	unpinned, err := db.UnpinUploads(ctx, *sl, *u)
```
m-cat, line 74: What happens if the user uploads the same skylink again after having unpinned it? Should add a test for that

skynet/bandwidth.go
```diff
@@ -52,10 +66,23 @@ func BandwidthDownloadCost(size int64) int64 {
 	return PriceBandwidthDownloadBase + chunks*PriceBandwidthDownloadIncrement
 }
 
+// RawStorageUsed calculates how much storage an upload with a given size
+// actually uses. This method returns the total underlying storage used and not
+// the adjusted number users see. Users see adjusted numbers in order to
+// shield them from the complexity of base/chunk redundancy.
+func RawStorageUsed(uploadSize int64) int64 {
+	baseSectorStorage := int64(PriceStorageUploadBase * RedundancyBaseSector)
```
m-cat, line 74: Why are we multiplying a storage price with a redundancy to get a storage amount?

skynet/bandwidth.go
```diff
@@ -52,10 +66,23 @@ func BandwidthDownloadCost(size int64) int64 {
...
+	chunkStorage := numChunks(uploadSize) * PriceStorageUploadIncrement * RedundancyChunk
```
m-cat, line 75: See above, I don't like that a price is part of a storage equation.

skynet/bandwidth.go
```diff
@@ -52,10 +66,23 @@ func BandwidthDownloadCost(size int64) int64 {
...
+
 // StorageUsed calculates how much storage an upload with a given size actually
-// uses.
+// uses. This method returns user-facing values. For the raw storage value
+// use RawStorageUsed().
 func StorageUsed(uploadSize int64) int64 {
-	return PriceStorageUploadBase + numChunks(uploadSize)*PriceStorageUploadIncrement
+	baseSectorStorage := int64(PriceStorageUploadBase * RedundancyBaseSector)
+	chunkStorage := numChunks(uploadSize) * PriceStorageUploadIncrement * RedundancyChunk
+	return (baseSectorStorage + chunkStorage) / RedundancyAPIDivisor
```
m-cat, line 85: Can this whole function just be `RawStorageUsed(uploadSize) / RedundancyAPIDivisor`?

skynet/bandwidth_test.go
```diff
@@ -24,18 +33,40 @@ func TestNumChunks(t *testing.T) {
...
+	tests := []struct {
+		size   int64
+		result int64
+	}{
+		{size: 0, result: baseSectorTotalSize},
+		{size: 1 * MiB, result: baseSectorTotalSize},
+		{size: 4 * MiB, result: baseSectorTotalSize},
+		{size: 5 * MiB, result: baseSectorTotalSize + chunkTotalSize},
+		{size: 50 * MiB, result: baseSectorTotalSize + 2*chunkTotalSize},
+		{size: 500 * MiB, result: baseSectorTotalSize + 13*chunkTotalSize},
```
m-cat, line 47: There should be a comment explaining why this is 13

m-cat (changes requested):
I hereby request changes.

skynet/bandwidth_test.go
```diff
@@ -24,18 +33,40 @@ func TestNumChunks(t *testing.T) {
...
-		{size: 0, result: SizeBaseSector},
-		{size: 1 * MiB, result: SizeBaseSector},
-		{size: 4 * MiB, result: SizeBaseSector},
-		{size: 5 * MiB, result: SizeBaseSector + SizeChunk},
-		{size: 50 * MiB, result: SizeBaseSector + 2*SizeChunk},
-		{size: 500 * MiB, result: SizeBaseSector + 13*SizeChunk},
+		{size: 0, result: baseSectorTotalSize / RedundancyAPIDivisor},
+		{size: 1 * MiB, result: baseSectorTotalSize / RedundancyAPIDivisor},
+		{size: 4 * MiB, result: baseSectorTotalSize / RedundancyAPIDivisor},
+		{size: 5 * MiB, result: (baseSectorTotalSize + chunkTotalSize) / RedundancyAPIDivisor},
```
m-cat, line 67: Is it correct that we divide by `RedundancyAPIDivisor` (3) for both redundancies (3 and 10)?

skynet/bandwidth.go
```diff
@@ -11,6 +11,20 @@ const (
 	// SizeChunk is the size of a chunk.
 	SizeChunk = 40 * MiB
```
m-cat, line 12: Why is a chunk 40mb?

Danger was confused that a 4.37 MB file is reported at 55.9 MB, since the redundancy for large files should be less than for small ones. I'm guessing it's because chunks are so large, but that really should be explained somewhere on the site, it's quite confusing (even for me).

Danger's tests:

1) 3.88 MB file > 14 MB
2) 4.37 MB file > 55.9 MB

api/handlers.go
ro-tex: Added a check for that error, it wasn't omitted on purpose.

api/routes.go
ro-tex: By accident - it should be followed by a `return`. Fixed.

database/upload.go
ro-tex: Hm, it doesn't seem so. I guess it was and then I forgot to remove it. I'll clean it up now.

database/upload.go
```diff
@@ -80,10 +82,64 @@ func (db *DB) UploadsBySkylink(ctx context.Context, skylink Skylink, offset, pag
...
+	var uploads []Upload
+	err = c.All(ctx, &uploads)
+	if err != nil {
+		return nil, err
+	}
+	return uploads, nil
+}
+
+// UnpinUploads unpins all uploads of this skylink by this user. Returns the
+// number of unpinned uploads.
```
ro-tex, line 123: I wonder if I should limit this method to unpinning a single upload.

User story:  
As a user I upload a given skyfile twice by accident. The file is big and takes a long time to upload. I don't want it to take twice its size from my storage quota, so I unpin the second upload. I am surprised and frustrated when I notice (or even worse - not notice) that both my uploads are unpinned because due to the content-based hashing they have identical skylinks.

OTOH, it might be confusing to have to unpin the same file several times?

I'm not sure. I'm inclined to go with the first approach, as it makes more sense to me but I would like to get some feedback.

database/user.go
ro-tex: Bad naming. It's meant to be megabits per second.

I renamed it to `mbpsToBytesPerSecond` and updated the comment.

database/user.go
ro-tex: The initial idea was to have `TierMinReserved` be the low guard value but then we needed to introduce a level for anonymous users, so I repurposed the value. The alternative was updating all values in the code and in the database but I didn't see a need to do that. We can do it in the future in case we need to use the low guard value for some reason.

test/upload_test.go
ro-tex: Test added.

skynet/bandwidth.go
ro-tex: Hm, this might be more of a naming issue than a real problem.

The basic premise here is that we charge everything in bytes, so a registry access "costs" 1MiB for a read and 5MiB for a write operation. That's why all of these are prefixed by "Price" - it's more to group them together logically than to add markup in any way.

The goal of this logic is to calculate how much actual storage a given upload takes up.  
So, `PriceStorageUploadBase` is the amount of data a base sector on its own takes up - 4MiB. `PriceStorageUploadIncrement` is how much a data chunk takes up - 40MiB.

I'm not sure this clear things up enough but I'm not sure how to approach this in a comment. Maybe we can try it real-time (voice or text) and figure out a better approach.

skynet/bandwidth_test.go
ro-tex: Added comments.

skynet/bandwidth_test.go
ro-tex: Yes, that was an agreed compromise. We'll be moving to a more transparent (and complicated) representation soon.

skynet/bandwidth.go
ro-tex: We have two data packages we work with - sector (4 MB) and chunk (10 sectors, i.e. 40 MB). I know there were plans for a smaller package called segment (64 byte) but I don't know what happened to it.

Each Sia file takes at least one sector which we call the base sector. That sector contains the basic metadata of the sia file and as much of the content as we can manage. This sector is very important and it's uploaded with 10x redundancy. Any data that won't fit in these 4 MB overflows in one or more chunks. Chunks are 40 MB each and are uploaded with 3x redundancy. The reason they are exactly 40 MB is that each chunk contains 10 sectors and 3x redundancy comes from the Reed-Salomon error correction encoding which translates those 10 sectors into 40 resulting sectors, any 10 of which are sufficient for reconstructing the original chunk data.

Also, we are currently reporting the size of skyfiles as raw size / 3 which is not very intuitive, so we are backtracking on that. But that's the reason you're currently seeing a 3MB file take 40/3 = 13.3MB of space and a 4.1 MB file take (4*10 + 40*3) / 3 = 53MB.

skynet/bandwidth.go
ro-tex: Renamed all Price* constants to Cost* in order to capture the idea that they represent bandwidth/storage costs rather than monetary values.

MSevey:
Not sure if it is applicable for this MR but the skyd unpin functionality is almost ready so we could be able to actively unpin files https://gitlab.com/SkynetHQ/skyd/-/merge_requests/62

api/handlers.go
```diff
@@ -228,6 +239,36 @@ func (api *API) userUploadsHandler(w http.ResponseWriter, req *http.Request, _ h
...
+	if err != nil {
+		api.WriteError(w, err, http.StatusInternalServerError)
+		return
+	}
+	api.WriteSuccess(w)
+	// Now that we've returned results to the caller, we can take care of some
+	// administrative details, such as user's quotas check.
+	// Nota that this call is not affected by the request's context, so we use
+	// a separate one.
+	api.checkUserQuotas(context.Background(), u)
```
peterjan, line 269: Doesn't it make more sense to just fire up a goroutine that handles this? It's a little bit awkward to be doing stuff after you've written success. Also I'm  not sure whether the connection isn't kept open for as long as this function didn't return... Isn't it possible this is keeping alive the client connection, even though we've already written the response headers and status code? Not sure.

api/handlers.go
```diff
@@ -228,6 +239,36 @@ func (api *API) userUploadsHandler(w http.ResponseWriter, req *http.Request, _ h
...
+	uid := ps.ByName("uploadId")
+	_, err = api.staticDB.UnpinUpload(req.Context(), uid, *u)
+	if err != nil {
+		api.WriteError(w, err, http.StatusInternalServerError)
+		return
+	}
+	api.WriteSuccess(w)
+	// Now that we've returned results to the caller, we can take care of some
+	// administrative details, such as user's quotas check.
+	// Nota that this call is not affected by the request's context, so we use
```
peterjan, line 267:
```suggestion
	// Note that this call is not affected by the request's context, so we use
```

api/handlers.go
```diff
@@ -300,12 +341,16 @@ func (api *API) trackUploadHandler(w http.ResponseWriter, req *http.Request, ps
...
-				UploaderID: u.ID,
-				SkylinkID:  skylink.ID,
+				SkylinkID: skylink.ID,
 			}
 		}()
 	}
 	api.WriteSuccess(w)
+	// Now that we've returned results to the caller, we can take care of some
+	// administrative details, such as user's quotas check.
+	// Nota that this call is not affected by the request's context, so we use
```
peterjan, line 351:
```suggestion
	// Note that this call is not affected by the request's context, so we use
```

api/handlers.go
```diff
@@ -300,12 +341,16 @@ func (api *API) trackUploadHandler(w http.ResponseWriter, req *http.Request, ps
...
+	// a separate one.
+	api.checkUserQuotas(context.Background(), u)
```
peterjan, line 353: same remark

api/handlers.go
```diff
@@ -412,6 +457,67 @@ func (api *API) trackRegistryWriteHandler(w http.ResponseWriter, req *http.Reque
...
+	}
+	_, err = api.staticDB.UnpinUploads(req.Context(), *skylink, *u)
+	if err != nil {
+		api.WriteError(w, err, http.StatusInternalServerError)
+		return
+	}
+	api.WriteSuccess(w)
+	// Now that we've returned results to the caller, we can take care of some
+	// administrative details, such as user's quotas check.
+	// Nota that this call is not affected by the request's context, so we use
```
peterjan, line 497:
```suggestion
	// Note that this call is not affected by the request's context, so we use
```

api/handlers.go
```diff
@@ -412,6 +457,67 @@ func (api *API) trackRegistryWriteHandler(w http.ResponseWriter, req *http.Reque
...
+	// a separate one.
+	api.checkUserQuotas(context.Background(), u)
```
peterjan, line 499: same remark

api/handlers.go
```diff
@@ -412,6 +457,67 @@ func (api *API) trackRegistryWriteHandler(w http.ResponseWriter, req *http.Reque
...
+}
+
+// checkUserQuotas compares the resources consumed by the user to their quotas
+// and sets the QuotaExceeded flag on their account if they exceed any.
+func (api *API) checkUserQuotas(ctx context.Context, u *database.User) {
```
peterjan, line 504: We're calling this on delete/unpin but are we also calling it on upload?

api/routes.go
```diff
@@ -76,3 +100,28 @@ func tokenFromRequest(r *http.Request) (string, error) {
...
```
peterjan, line 120: conversation resolved but code unchanged - forgot to push? - resolved by mistake?

api/stripe.go
```diff
@@ -194,34 +208,35 @@ func (api *API) processStripeSub(ctx context.Context, s *stripe.Subscription) er
...
-			latestCreated = created
-			latestSub = subsc.ID
-			// It seems weird that the Plan.ID is actually a price id but this
-			// is what we get from Stripe.
-			u.Tier = stripePrices()[subsc.Plan.ID]
-			u.SubscribedUntil = time.Unix(subsc.CurrentPeriodEnd, 0).UTC()
-			u.SubscriptionStatus = string(subsc.Status)
-			u.SubscriptionCancelAt = time.Unix(subsc.CancelAt, 0)
-			u.SubscriptionCancelAtPeriodEnd = subsc.CancelAtPeriodEnd
+		if mostRecentSub == nil || mostRecentSub.Created < subsc.Created {
```
peterjan, line 220: (nit) `if mostRecentSub == nil || subsc.Created > mostRecentSub.Created {` is more logical

database/user_test.go
```diff
@@ -0,0 +1,42 @@
+package database
+
+import (
+	"testing"
+	"time"
+)
+
+//TestMonthStart ensures we calculate the start of the subscription month
```
peterjan, line 8:
```suggestion
// TestMonthStart ensures we calculate the start of the subscription month
```

database/user_test.go
```diff
@@ -0,0 +1,42 @@
...
+			subUntil:     time.Date(2020, 1, now.Day(), 3, 4, 5, 6, time.UTC),
+			startOfMonth: time.Date(now.Year(), now.Month(), now.Day(), 0, 0, 0, 0, time.UTC),
+		},
+		{
+			// If the monthly sub cycle resets tomorrow, expect the start of
+			// month to be tomorrows date but a month back.
+			subUntil:     time.Date(2020, 1, now.Day()+1, 3, 4, 5, 6, time.UTC),
+			startOfMonth: time.Date(now.Year(), now.Month()-1, now.Day()+1, 0, 0, 0, 0, time.UTC),
+		},
+	}
```
peterjan, line 34: why are you hardcoding 2020 in the `subUntil`s?

skynet/bandwidth.go
```diff
@@ -5,41 +5,59 @@ const (
...
-	// PriceBandwidthRegistryRead the bandwidth cost of a single registry read
-	PriceBandwidthRegistryRead = MiB
+	// RedundancyBaseSector describes the base sector redundancy the portal is
+	// using. This is not freely configurable because we need database
+	// consistency.
+	RedundancyBaseSector = 10
+	// RedundancyBaseSector describes the redundancy of regular chunks the
+	// portal is using. This is not freely configurable because we need database
+	// consistency.
+	RedundancyChunk = 3
```
peterjan, line 25: update docstring

skynet/bandwidth.go
```diff
@@ -5,41 +5,59 @@ const (
...
+	// RedundancyAPIDivisor is the value by which we divide the raw used storage
```
peterjan, line 26: Does this mean we're not making a difference between base sector redundancy vs fanout redundancy when it comes to signalling the storage amounts to the user? Just asking out of curiosity.

peterjan (approved):

ro-tex:
> Not sure if it is applicable for this MR but the skyd unpin functionality is almost ready so we could be able to actively unpin files https://gitlab.com/SkynetHQ/skyd/-/merge_requests/62

Not for this MR but definitely worth a follow up. I added #45 for that.

api/handlers.go
ro-tex: TBH, I'm not sure either. My initial implementation used a goroutine for exactly those reasons but then I decided that it's only complicating things. I'll switch to a goroutine in a F/U, it's a single `go` keyword anyway.

database/user_test.go
ro-tex: The idea was to have something in the past, a subscription that's been running for a while. I can also use `now.Year()-1` if you prefer.

skynet/bandwidth.go
ro-tex: 1. Yes, this is exactly what it meant. The  idea was to give the user a fizzy value that's in the close vicinity of the real one but without all the complications of reality.
2. This is now obsolete and unused, as it caused a lot of trouble. We switched to counting the actual on-disk size of the file towards the user's quota and reporting only that. I forgot to clean up this thing, which I will do now.

api/handlers.go
ro-tex: Absolutely, check out `trackUploadHandler`.

api/routes.go
ro-tex: Good catch! This was happening at two places and I fixed one of them and somehow didn't fix the other.
Fixed now.

api/stripe.go
ro-tex: Sure, makes sense.

database/upload.go
ro-tex: We ended up with two endpoints:  
`DELETE /user/upload/:uploadId` which unpins a specific single upload  
`DELETE /skylink/:skylink` which unpins all.

peterjan (approved):

api/handlers.go
```diff
@@ -264,9 +264,9 @@ func (api *API) userUploadDeleteHandler(w http.ResponseWriter, req *http.Request
 	api.WriteSuccess(w)
 	// Now that we've returned results to the caller, we can take care of some
 	// administrative details, such as user's quotas check.
-	// Nota that this call is not affected by the request's context, so we use
+	// Note that this call is not affected by the request's context, so we use
 	// a separate one.
-	api.checkUserQuotas(context.Background(), u)
+	go api.checkUserQuotas(context.Background(), u)
```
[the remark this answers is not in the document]
ro-tex: Will be done as F/U in #47

database/upload.go
```diff
@@ -80,10 +82,61 @@ func (db *DB) UploadsBySkylink(ctx context.Context, skylink Skylink, offset, pag
...
+	matchStage := bson.D{{"$match", bson.D{
+		{"skylink_id", skylink.ID},
+		{"unpinned", false},
+	}}}
 	return db.uploadsBy(ctx, matchStage, offset, pageSize)
 }
 
+// UnpinUpload unpins a single upload by this user. Returns the number of
+// unpinned uploads.
+func (db *DB) UnpinUpload(ctx context.Context, uploadId string, user User) (int64, error) {
```
m-cat, line 94: q: I don't really understand the difference between this and `unpinUploads`. One unpins by an upload ID and the other by a skylink? Shouldn't they amount to the same operation?

database/user_test.go
```diff
@@ -0,0 +1,42 @@
...
+func TestMonthStart(t *testing.T) {
+	now := time.Now()
+	tests := []struct {
+		subUntil     time.Time
+		startOfMonth time.Time
+	}{
+		{
+			// If the monthly sub cycle reset yesterday, expect the start of
+			// month to also be yesterday.
+			subUntil:     time.Date(2020, 1, now.Day()-1, 3, 4, 5, 6, time.UTC),
```
m-cat, line 19: f/u: Test with different months and years apart from just January and 2020

m-cat: Maybe worth adding an edge case for end of month, start of month, and leap years?

database/user.go
```diff
@@ -30,13 +31,67 @@ const (
...
 	TierMaxReserved
+
+	// mbpsToBytesPerSecond is a multiplier to get from megabits per second to
+	// bytes per second.
+	mbpsToBytesPerSecond = 1024 * 1024 / 8
+
+	// filesAllowedPerTB defines a limit of number of uploaded files we impose
+	// on users. While we define it per TB, we impose it based on their entire
+	// quota, so an Extreme user will be able to upload up to 400_000 files
+	// before being hit with a speed limit.
```
m-cat, line 42: Should this be...?
```suggestion
	// before being hit with a storage limit.
```

database/user.go
```diff
@@ -30,13 +31,67 @@ const (
...
+	filesAllowedPerTB = 25_000
 )
 
 var (
 	// True is a helper for when we need to pass a *bool to MongoDB.
 	True = true
 	// False is a helper for when we need to pass a *bool to MongoDB.
 	False = false
+	// UserLimits defines the speed limits for each tier.
```
m-cat, line 51:
```suggestion
	// UserLimits defines the limits for each tier.
```

m-cat (approved):

database/upload.go
ro-tex: One unpins a specific upload while the other unpins all uploads of the given skylink by the given user. The difference appears when the user uploads the exact same file multiple times. Those uploads will result in the same skylink.

database/user_test.go
ro-tex: GitHub is rather unhelpful with follow ups - I need to link it manually. F/U: #48

database/user.go
ro-tex: This is intended. Once you exceed your quota you get hit with a *speed* limit, i.e. all of your speed limitations become the same as an anonymous account.

database/user.go
ro-tex: Fixed.

database/upload.go
m-cat: how do you unpin a specific upload? I thought you could only unpin by skylink, hence why I thought both operations would be the same.

ro-tex: When a user uploads a file that results in a skyfile stored on Sia and an upload record in our database. The skyfile is posted on the network by the portal's Sia node. The upload records are not directly connected to the uploaded skyfiles, they are just a record we keep on which users are paying for the upkeep of the given skyfile. While there is at least one user paying for that upkeep, the portal's node will not unpin the file from the network. The portal's mechanism for monitoring this are the upload records in the DB. 

When the user chooses to unpin a single upload, we delete a single upload record they have for that skyfile. Or, rather, mark it as `unpinned`. When they choose to unpin all, we mark all of the uploads they have made for that skyfile as `unpinned`. If no other user is pinning this skyfile, the portal's node will unpin it from the network on its next cleanup cycle. If, however, other users are still pinning the skyfile the portal will keep it pinned.

The unpinning of a specific upload by its id was implemented mostly for the purposes of the front end, so a user clicking `unpin` on their second upload of a given skyfile will see that specific second upload as unpinned. This allows for smooth and predictable UX.

PR #40, opened 2021-03-23
Change license.

ro-tex:


README.md
```diff
@@ -32,6 +32,17 @@ KRATOS_ADDR=localhost:4433
 OATHKEEPER_ADDR=localhost:4456
 ```
 
+## License
+
+Skynet uses a custom [License](./LICENSE.md). The Skynet License is a source
```
MSevey, line 37:
```suggestion
Skynet Accounts uses a custom [License](./LICENSE.md). The Skynet License is a source
```

MSevey (dismissed):

kwypchlo (approved):

README.md
```diff
@@ -32,6 +32,17 @@ KRATOS_ADDR=localhost:4433
...
+## License
+
+Skynet Accounts uses a custom [License](./LICENSE.md). The Skynet License is a source
+code license that allows you to use, modify and distribute the software, but
+you must preserve the payment mechanism in the software.
+
+For the purposes of complying with our code license, you can use the
+following Siacoin address:
+
+`fb6c9320bc7e01fbb9cd8d8c3caaa371386928793c736837832e634aaaa484650a3177d6714a`
```
m-cat, line 44: Should this be in the license itself?

m-cat (approved):

README.md
MSevey: no, because we would have to version the license every time we change the address. It also allows us to use different addresses for different products. Even though the license is called `Skynet License` it is actually generic in that anyone could use the Skynet License to protect their open source project that had payments.

MSevey (approved):

PR #52, opened 2021-05-04
Add a simple health check.

ro-tex:
* The health check always returns 200 when the service is alive.
* The body of the health check contains a JSON with the status of the database which can be `true` or `false` (if the DB is down)

I don't mind changing the way this works to return an error code when the DB is down instead of a 200 - both would make sense to me.

### DIFF
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -17,6 +17,18 @@ import (
 	"gitlab.com/NebulousLabs/errors"
 )
 
+// healthHandler returns the status of the service
+func (api *API) healthHandler(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
+	status := struct {
+		DBAlive bool `json:"dbAlive"`
+	}{}
+	_, err := api.staticDB.UserBySub(req.Context(), "", false)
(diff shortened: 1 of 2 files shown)


m-cat (approved):

ChrisSchinnerl (approved):

api/handlers.go
```diff
@@ -17,6 +17,18 @@ import (
 	"gitlab.com/NebulousLabs/errors"
 )
 
+// healthHandler returns the status of the service
+func (api *API) healthHandler(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
+	status := struct {
+		DBAlive bool `json:"dbAlive"`
+	}{}
+	_, err := api.staticDB.UserBySub(req.Context(), "", false)
```
peterjan, line 25: I believe mongo clients have a `Ping` method you could use to accomplish this. This will work as well though so I'll approve the PR. Just saying that `Ping` is probably a good option as well for a health check.

ro-tex: Good point. F/U: https://github.com/SkynetLabs/skynet-accounts/pull/53

PR #53, opened 2021-05-05
Improved healthcheck

ro-tex:
Switch to using Mongo's Ping method.

### DIFF
--- a/database/user.go
+++ b/database/user.go
@@ -16,6 +16,7 @@ import (
 	"go.mongodb.org/mongo-driver/bson/primitive"
 	"go.mongodb.org/mongo-driver/mongo"
 	"go.mongodb.org/mongo-driver/mongo/options"
+	"go.mongodb.org/mongo-driver/mongo/readpref"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


ChrisSchinnerl (approved):

database/user.go
```diff
@@ -319,6 +320,14 @@ func (db *DB) UserStats(ctx context.Context, user User) (*UserStats, error) {
 	return db.userStats(ctx, user)
 }
 
+// Ping sends a ping command to verify that the client can connect to the DB and
+// specifically to the primary.
+func (db *DB) Ping(ctx context.Context) error {
+	ctx2, cancel := context.WithTimeout(ctx, 5*time.Second)
```
peterjan, line 326: Just overwrite `ctx` instead of doing `ctx2`. I'm not even sure why we need to pass in the request context here, is that really necessary?

ro-tex: The caller should have control over how long the ping should take. If the context you're running within is about to expire in 1 second you don't want to start an operation that can take longer than that, so the context chain must be preserved.

ro-tex: About overwriting - I generally don't like overwriting parameters. In longer functions it becomes confusing to follow which params were overwritten and which - not, simply because the code highlighting nudges you towards the assumption that this value is a parameter and comes from outside. So, I would prefer to have it in a different variable just to stay consistent on that. I agree that the name is not great and I can change it if you'd like.

m-cat (approved):

PR #57, opened 2021-07-07
Public limits endpoint

ro-tex:
Exposes an endpoint which requires no authentication and lists all speed limits of the portal. Currently, those are just the limits different tiers of user accounts have. In the future we might add others as well.

### DIFF
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -108,8 +108,14 @@ func (api *API) userHandler(w http.ResponseWriter, req *http.Request, _ httprout
 
 // userLimitsHandler returns the speed limits which apply for this user.
 func (api *API) userLimitsHandler(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
-	u := api.userFromRequest(req)
-	if u == nil || u.QuotaExceeded {
+	sub, err := api.subFromRequest(req)
+	if err != nil {
+		api.WriteJSON(w, database.UserLimits[database.TierAnonymous])
+		return
+	}
+	// TODO: If we want to add caching to the /limits endpoint, this is the spot.
+	u, err := api.staticDB.UserBySub(req.Context(), sub, false)
+	if err != nil || u == nil || u.QuotaExceeded {
 		api.WriteJSON(w, database.UserLimits[database.TierAnonymous])
 		return
 	}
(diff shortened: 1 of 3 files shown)


m-cat (dismissed):

kwypchlo (approved):

database/user.go
```diff
@@ -130,12 +135,24 @@ type (
...
+		UploadBandwidth   int    `json:"upload"`        // bytes per second
+		DownloadBandwidth int    `json:"download"`      // bytes per second
+		MaxUploadSize     int64  `json:"maxUploadSize"` // the max size of a single upload in bytes
+		MaxNumberUploads  int    `json:"-"`
+		RegistryDelay     int    `json:"registry"` // ms delay
+		Storage           int64  `json:"-"`
+	}
+	// TierLimitsPublic is a DTO specifically designed to inform the public
+	// about the different limits of each account tier.
+	TierLimitsPublic struct {
```
ChrisSchinnerl, line 148: If this is just for the API I'd move it to the API's package in `handler.go`

api/handlers.go
```diff
@@ -27,6 +27,28 @@ func (api *API) healthHandler(w http.ResponseWriter, req *http.Request, _ httpro
...
+			UploadBandwidth:   t.UploadBandwidth * 8,   // convert from bytes
+			DownloadBandwidth: t.DownloadBandwidth * 8, // convert from bytes
+			MaxUploadSize:     t.MaxUploadSize,
+			MaxNumberUploads:  t.MaxNumberUploads,
+			RegistryDelay:     t.RegistryDelay,
+			Storage:           t.Storage,
+		}
+	}
+	resp := struct {
+		UserLimits map[int]database.TierLimitsPublic `json:"userLimits"`
```
ChrisSchinnerl, line 45: I'd make this a slice instead of a map with integers as keys.

ChrisSchinnerl (changes requested):

database/user.go
ro-tex: I moved it to handler.go but I'd like to keep the name different. It makes it clear what the purpose of the type is without reading the comments.

kwypchlo (approved):

ChrisSchinnerl (approved):

PR #59, opened 2021-08-30
Rename handlers to match their verbs.

ro-tex:
This is a preparatory PR for implementing user registration.


ChrisSchinnerl (approved):

peterjan (approved):
`nit` I like `handleGetHealth` more than I like `healthGET`

PR #61, opened 2021-08-30
Add JWT/JWKS support

ro-tex:
This PR introduces the generation, signing, and verification of JWTs.  
It also adds a new endpoint which exposes a public version of our JWKS, so others can use it in order to verify the JWTs signed by us.


Makefile
```diff
@@ -99,7 +99,7 @@ check:
 
 test:
 	go test -short -tags='debug testing netgo' -timeout=5s $(pkgs) -run=. -count=$(count)
-test-long: clean fmt vet lint-ci
+test-long: clean fmt vet lint-ci test
```
MSevey, line 102: You shouldn't need to include `test` here. That is just going to run the short tests twice.
Long tests are skipped unless it is a long test run, but short test shouldn't have any skip code in them

hash/hash.go
```diff
@@ -0,0 +1,119 @@
...
+	"fmt"
+	"strings"
+
+	"gitlab.com/NebulousLabs/errors"
+	"gitlab.com/NebulousLabs/fastrand"
+	"golang.org/x/crypto/argon2"
+)
+
+var (
+	ErrInvalidHash               = errors.New("the encoded hash is not in the correct format")
```
MSevey, line 16: Follow Up: there must be some linting that this go repo doesn't have. skyd complains if exported items don't have comments.

hash/hash.go
```diff
@@ -0,0 +1,119 @@
...
+	ErrIncompatibleVersion       = errors.New("incompatible version of argon2")
+	ErrMismatchedHashAndPassword = errors.New("passwords do not match")
+
+	// This is the configuration of the argon2id hasher. For the moment we'll
+	// use the setup we had with Kratos. Later on we can move this to a config.
+
```
MSevey, line 22: nit: extra newline

hash/hash.go
```diff
@@ -0,0 +1,119 @@
...
+	config = argon2Config{
+		SaltLength:  16,
+		Iterations:  2,
+		Memory:      131072,
+		Parallelism: 1,
+		KeyLength:   16,
+	}
+)
+
+type argon2Config struct {
```
MSevey, line 32: Follow Up: Docstring

hash/hash.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+	return ErrMismatchedHashAndPassword
+}
+
+// decodeHash is a helper method which extracts the configuration from the
+// encoded hash string and returns its parts.
+//
+// Example password record:
+// {"hashed_password":"$argon2id$v=19$m=131072,t=2,p=1$dwr95pEjaa7emZOu9bDAWw$eDQwOMoSyRmzyvpD/wwGBg"}
+func decodeHash(encodedHash string) (ac *argon2Config, salt, hash []byte, err error) {
```
MSevey, line 85: Follow Up: This method could use inline comments.

hash/hash.go
```diff
@@ -0,0 +1,119 @@
...
+type argon2Config struct {
+	SaltLength  uint32
+	Iterations  uint32
+	Memory      uint32
+	Parallelism uint8
+	KeyLength   uint32
+}
+
+// Generate hashes the given password with the default configuration.
+func Generate(password []byte) ([]byte, error) {
```
MSevey, line 41: REQ: needs testing of this file.

main.go
```diff
@@ -26,6 +26,9 @@ const (
 )
 
 var (
+	// envAccountsJWKSFile holds the name of the environment variable which
+	// holds the path to the JWKS file we need to use. Optional.
+	envAccountsJWKSFile = "ACCOUNTS_JWKS_FILE"
```
MSevey, line 31: Follow Up: We need to make sure that this gets follow up all the way through our documentation. This will need to be added to the `.env` file right?

ro-tex: It defaults to `/accounts/conf/jwks.json`. So, if you don't set anything in your `.env`, it will look for the file there.
I'll add some explanation to the README.

Makefile
ro-tex: Oh, you're right. :)

api/handlers.go
```diff
@@ -80,13 +79,12 @@ func (api *API) loginPOST(w http.ResponseWriter, req *http.Request, _ httprouter
 		api.WriteError(w, err, http.StatusUnauthorized)
 		return
 	}
-	exp, err := jwt.TokenExpiration(token)
-	if err != nil {
-		api.staticLogger.Traceln("Error checking token expiration:", err)
-		api.WriteError(w, err, http.StatusUnauthorized)
+	exp := token.Expiration()
```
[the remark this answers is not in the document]
ro-tex: Done.

hash/hash.go
ro-tex: F/U: https://github.com/SkynetLabs/skynet-accounts/issues/64

hash/hash.go
ro-tex: I am not sure what I can add in those inline comments. The code is extremely straightforward and self-documenting.
It extracts the configuration for the `argon2` hashing (explained in the docstring) by extracting each bit of it in a separate block.

Example:  
```go
	salt, err = base64.RawStdEncoding.DecodeString(parts[4])
	if err != nil {
		return nil, nil, nil, err
	}
	ac.SaltLength = uint32(len(salt))
```
I can add a comment to this code which says "extracts the password salt" but that seems redundant to a level at which it reduces the redability of the code rather than enchancing it.

If you feel strongly about those comments I will add them but in my opinion they will not be an improvement.

main.go
MSevey: 👍

README.md
```diff
@@ -12,33 +12,76 @@ All local secrets are loaded from a `.env` file in the root directory of the pro
...
 STRIPE_API_KEY="put-your-key-here"
 STRIPE_WEBHOOK_SECRET="put-your-secret-here"
 ```
 
 There are some optional ones, as well:
 
 ```.env
-SKYNET_ACCOUNTS_LOG_LEVEL=trace
-KRATOS_ADDR=localhost:4433
+ACCOUNTS_JWKS_FILE="/accounts/conf/jwks.json"
```
MSevey, line 29: cc @firyx 
we will need to update ansible/lastpass for the server set up scripts. We can wait until it is merged but wanted to call your attention to it.

hash/hash.go
MSevey: it is redundant if you leave out the why 😄 
Tell me the story, take me on this journey of decoding the hash.

imo inline comments should explain why we are doing each step as well as what each step is. I agree that each individual step is pretty straightforward in terms of what it is doing, but reviewing the code, or coming back to it to potentially edit it in 6 months, I can't pull the context of why each step is needed or if there are additionally order of operation concerns. 

Inline comments should save me the mental energy of having to read the code to understand the why.  

Example:
```
if version != argon2.Version {
		return nil, nil, nil, ErrIncompatibleVersion
}
```
I know this is checking for the correct version. But are there other considerations to consider with different versions? Do we need to offer compatibility code in the future if the version changes?

MSevey: Being able to understand the why through inline comments also helps reviewers provide better feedback or potential alternatives because it is clear what the developer's thought process was when writing the code.

ro-tex: > Tell me the story, take me on this journey of decoding the hash.

Good point, you're absolutely right. Comments coming right up.

jwt/jwt.go
```diff
@@ -227,43 +293,75 @@ func UserDetailsFromJWT(ctx context.Context) (firstName, lastName, email string,
...
+		case "testing":
+			b, err = ioutil.ReadFile("fixtures/jwks.json")
+		default:
+			b, err = ioutil.ReadFile(AccountsJWKSFile)
+		}
+		if err != nil {
+			logger.Warningln("ERROR while reading accounts JWKS", err)
+			return nil, err
+		}
+		set := jwk.NewSet()
```
ChrisSchinnerl, line 347: Is this necessary? Does 
```go
var set jwk.Set
err = json.Unmarshal(b, &set)
```
not work?

jwt/jwt.go
```diff
@@ -2,55 +2,121 @@ package jwt
...
 	}
-	return int64(claims["exp"].(float64)), nil
+	if sigAlgo == "" {
+		return nil, nil, errors.New("failed to determine signature algorithm")
+	}
+	bytes, err := jwt.Sign(t, sigAlgo, key)
+	if err != nil {
+		return nil, nil, errors.New("failed to sign token")
+	}
+	tk, err := jwt.Parse(bytes)
```
ChrisSchinnerl, line 115: Seems like you never need the parsed result so I wouldn't bother parsing and returning it

ChrisSchinnerl (approved):

hash/hash.go
```diff
@@ -0,0 +1,127 @@
...
+
+	// config is the configuration of the argon2id hasher.
+	// For the moment we'll use the setup we had with Kratos. Later on we can
+	// move this to a configuration file, if needed.
+	config = argon2Config{
+		SaltLength:  16,
+		Iterations:  2,
+		Memory:      131072,
+		Parallelism: 1,
+		KeyLength:   16,
```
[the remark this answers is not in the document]
ro-tex: We could. These are the default values used by Kratos.

The main advantage of the argon2 algo is that the hash size is not the only determining factor in determining the complexity of the hashing process. We want slow hashing, so people won't be able to brute force it but not so slow that it would be bad for us as well. This is controlled by the other 3 parameters.

16 bytes is the recommended size by argon2's authors but I don't see a problem with doubling that.

Links:
 - https://www.ory.sh/choose-recommended-argon2-parameters-password-hashing/
 - https://password-hashing.net/argon2-specs.pdf (see "Recommended parameters")

If we choose to change the length, all existing passwords won't be affected.

api/handlers.go
```diff
@@ -80,13 +81,17 @@ func (api *API) loginPOST(w http.ResponseWriter, req *http.Request, _ httprouter
...
 		return
 	}
-	exp, err := jwt.TokenExpiration(token)
-	if err != nil {
-		api.staticLogger.Traceln("Error checking token expiration:", err)
-		api.WriteError(w, err, http.StatusUnauthorized)
+	// We fetch the expiration time of the token, so we can set the expiration
+	// time of the cookie to match it.
+	exp := token.Expiration()
+	if time.Now().UTC().After(exp) {
```
peterjan, line 87: I think you're off here with the timezones because 4 lines below you do `exp.UTC()`.
So it should probably be 
```
if exp.UTC().before(time.Now().UTC()) {
    // handle expired token
}
```
is that  possible?

hash/hash.go
```diff
@@ -0,0 +1,127 @@
...
+		base64.RawStdEncoding.EncodeToString(hash),
+	); err != nil {
+		return nil, errors.AddContext(err, "failed to generate password hash")
+	}
+
+	return b.Bytes(), nil
+}
+
+// Compare verifies whether the given password matches the given hash.
+func Compare(password []byte, hash []byte) error {
```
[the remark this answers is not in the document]
peterjan: maybe something like `func(us UserSecret) Equals(other UserSecret) bool`?

hash/hash.go
```diff
@@ -0,0 +1,127 @@
...
+}
+
+// decodeHash is a helper method which extracts the configuration from the
+// encoded hash string and returns its parts.
+//
+// Example of a password hash record:
+// "$argon2id$v=19$m=131072,t=2,p=1$dwr95pEjaa7emZOu9bDAWw$eDQwOMoSyRmzyvpD/wwGBg"
+func decodeHash(encodedHash string) (ac *argon2Config, salt, hash []byte, err error) {
+	parts := strings.Split(encodedHash, "$")
+	if len(parts) != 6 {
```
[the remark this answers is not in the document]
ro-tex: I agree about the magic number, I'll add a descriptive comment explaining what each part of the string is.

hash/hash_test.go
```diff
@@ -0,0 +1,47 @@
...
+	"crypto/subtle"
+	"encoding/base64"
+	"testing"
+
+	"gitlab.com/NebulousLabs/fastrand"
+)
+
+// TestDecodeHash is a short sanity check that ensures decodeHash works as
+// expected.
+func TestDecodeHash(t *testing.T) {
```
peterjan, line 13: This unit test is not covering any of the exit flows that aren't the happy flow.

peterjan: You definitely want to cover `ErrInvalidHash` and `ErrIncompatibleVersion` at least right?

jwt/jwt.go
```diff
@@ -227,43 +293,75 @@ func UserDetailsFromJWT(ctx context.Context) (firstName, lastName, email string,
...
-//  "sub": "695725d4-a345-4e68-919a-7395cb68484c"
-//}
-func ValidateToken(logger *logrus.Logger, t string) (*jwt.Token, error) {
-	keyForTokenWithLogger := func(token *jwt.Token) (interface{}, error) {
-		return keyForToken(logger, token)
-	}
-	token, err := jwt.Parse(t, keyForTokenWithLogger)
+// }
+func ValidateToken(logger *logrus.Logger, t string) (jwt.Token, error) {
+	_, err := accountsKeySet(logger)
```
peterjan, line 299: does this mean we are reading this file on every validation? Seems like something we might want to cache right?

jwt/jwt.go
```diff
@@ -227,43 +293,75 @@ func UserDetailsFromJWT(ctx context.Context) (firstName, lastName, email string,
...
+			b, err = ioutil.ReadFile(AccountsJWKSFile)
+		}
+		if err != nil {
+			logger.Warningln("ERROR while reading accounts JWKS", err)
+			return nil, err
+		}
+		set := jwk.NewSet()
+		err = json.Unmarshal(b, set)
+		if err != nil {
+			logger.Warningln("ERROR while parsing accounts JWKS", err)
```
peterjan, line 350: Is this log level severe enough? Seems like this is a pretty critical issue.

jwt/jwt.go
ro-tex: I couldn't make it work. The `jwk.NewSet()` is this:
```go
func NewSet() Set {
	return &set{}
}
```
So, while the `Set` interface is public, the returned type is unexported. There might be a way to do this without this function call but I'm not sure how.

jwt/jwt.go
peterjan: Oh I just realised that it is cached, I'm wondering though why we're not reading the key set on init. Something like that shouldn't be lazy loaded imo but rather prevent the server from coming up even. I guess this also goes hand in hand with my comment on the severity of the warning that we print when the file is not there. Shouldn't we simply exit if our accounts server doesn't have a keys file?

jwt/jwt.go
```diff
@@ -227,43 +293,75 @@ func UserDetailsFromJWT(ctx context.Context) (firstName, lastName, email string,
...
 	}
-	keys := keySet.LookupKeyID(token.Header["kid"].(string))
-	if len(keys) == 0 {
-		return nil, errors.New("no suitable keys found")
+	return token, nil
+}
+
+// accountsKeySet checks whether we have the needed key set cached and if we
+// don't it fetches it and caches it for us. The method returns the first key in
+// the set or ann error.
```
peterjan, line 323:
```suggestion
// the set or an error.
```

jwt/jwt.go
ro-tex: Oh, this is an artifact of splitting the code into several PRs. This function is not actually used in this PR and should have been in the next one where you can find this line where we use the token to grab its expiration time:
https://github.com/SkynetLabs/skynet-accounts/pull/62/files#diff-855b4003387cd9e2b8e2c353c747fd1817eccf1d223b4066f0d7a1db0b4b4b4eR109

build/debug_off.go
```diff
@@ -1,3 +1,4 @@
+//go:build !debug
 // +build !debug
```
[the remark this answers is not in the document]
ro-tex: I am not sure. I know that `gofmt` adds the first one (`//go:build !debug`) if the second one is there and the second one is the old one which was used with go versions below 1.17.

So, unless we want to limit our users to go 1.17, I would prefer to keep both lines.

hash/hash.go
```diff
@@ -0,0 +1,127 @@
...
+
+type argon2Config struct {
+	SaltLength  uint32
+	Iterations  uint32
+	Memory      uint32
+	Parallelism uint8
+	KeyLength   uint32
+}
+
+// Generate hashes the given password with the default configuration.
```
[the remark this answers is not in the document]
ro-tex: Docstring updated and comments added.

hash/hash.go
```diff
@@ -0,0 +1,127 @@
...
+func Generate(password []byte) ([]byte, error) {
+	salt := fastrand.Bytes(int(config.SaltLength))
+	hash := argon2.IDKey(password, salt, config.Iterations, config.Memory, config.Parallelism, config.KeyLength)
```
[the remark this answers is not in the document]
ro-tex: I think it would be very confusing for everyone reading this who's not really into the matter. Password hashing is a familiar concept to most people and throwing the word "key" at them would only create confusion. I, myself, don't know how argon2 works under the hood and I haven't looked into how this "key" is generated and used.

hash/hash.go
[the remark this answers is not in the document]
ro-tex: Wait... These two things are not two of a kind. One of them is a plaintext secret coming from the user and the other is the hashed version of a secret that we store in the database. That's why I've chosen the variable names there.

I was going to respond to https://github.com/SkynetLabs/skynet-accounts/pull/61#discussion_r700884476 directly but this is a very good example of the problem that would create. `UserSecret` does not tell us what we have in there - whether it's a plaintext, a ciphertext, or a hash. Or a "key" in the case of argon2, which functions like a hash for our purposes. I agree that neither does `[]byte` but that at least doesn't add to the confusion. 

I propose to keep referring to argon2 keys as hashes because we use them as such and it will add less confusion.  
I will add a type called `Argon2HashRecord` for the `[]byte` returned by argon2.  
I will change the `Compare` function to accept a string and a Argon2HashRecord types, so it's clearer what should go where.

Does that sound like a good solution?

hash/hash.go
```diff
@@ -0,0 +1,127 @@
...
+		&b,
+		"$argon2id$v=%d$m=%d,t=%d,p=%d$%s$%s",
+		argon2.Version, config.Memory, config.Iterations, config.Parallelism,
+		base64.RawStdEncoding.EncodeToString(salt),
+		base64.RawStdEncoding.EncodeToString(hash),
+	); err != nil {
+		return nil, errors.AddContext(err, "failed to generate password hash")
+	}
+
+	return b.Bytes(), nil
```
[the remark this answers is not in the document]
ro-tex: Added type `Argon2HashRecord`.

api/handlers.go
ro-tex: Yes, well spotted! Thanks!

hash/hash_test.go
ro-tex: I added this quickly just to ensure the method does what I expect it to. And then I forgot to expand it. :)
More cases added.

jwt/jwt.go
ro-tex: Yes, that makes sense. While writing this I was thinking about the mixed case where we're running with both accounts and Kratos signing JWTs but it looks more and more likely that once we deliver this we will immediately axe Kratos. I will make it required.

jwt/jwt.go
ChrisSchinnerl: ohhh it's an interface. nevermind then

MSevey (approved):

ChrisSchinnerl (approved):

PR #62, opened 2021-08-30
User creation

ro-tex:
This PR introduces:
* user creation in MongoDB
* user login with email and password

Part of #63


api/handlers.go
```diff
@@ -10,9 +10,11 @@ import (
 	"time"
 
 	"github.com/NebulousLabs/skynet-accounts/database"
+	"github.com/NebulousLabs/skynet-accounts/hash"
 	"github.com/NebulousLabs/skynet-accounts/jwt"
 	"github.com/NebulousLabs/skynet-accounts/metafetcher"
 	"github.com/julienschmidt/httprouter"
+	"github.com/sirupsen/logrus"
```
ChrisSchinnerl, line 17: why not use our own lib?

api/handlers.go
```diff
@@ -67,6 +69,55 @@ func (api *API) limitsGET(w http.ResponseWriter, _ *http.Request, _ httprouter.P
 
 // loginPOST starts a user session by issuing a cookie
 func (api *API) loginPOST(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
+	err := req.ParseForm()
```
ChrisSchinnerl, line 72: You don't need this. `PostFormValue` should implicitly call it.

api/handlers.go
```diff
@@ -182,6 +233,38 @@ func (api *API) userStatsGET(w http.ResponseWriter, req *http.Request, _ httprou
...
+		return
+	}
+	email := req.PostFormValue("email")
+	// TODO Validation
+	if email == "" {
+		api.WriteError(w, errors.New("email is required"), http.StatusBadRequest)
+		return
+	}
+	pw := req.PostFormValue("password")
+	// TODO Validation (basic complexity)
```
ChrisSchinnerl, line 250: I don't think we want to enforce any rules on the pw anymore.

api/handlers.go
```diff
@@ -182,6 +233,38 @@ func (api *API) userStatsGET(w http.ResponseWriter, req *http.Request, _ httprou
...
 
+// userPOST creates a new user.
+func (api *API) userPOST(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
+	err := req.ParseForm()
+	if err != nil {
+		api.WriteError(w, err, http.StatusInternalServerError)
+		return
+	}
+	email := req.PostFormValue("email")
+	// TODO Validation
```
ChrisSchinnerl, line 244: I'd add this right away. Just need to do `address, err := mail.Parse(email)`
It's in the stdlib.

Makefile
```diff
@@ -109,5 +109,6 @@ test-long: clean fmt vet lint-ci
 test-int: test-long start-mongo
 	GORACE='$(racevars)' go test -race -v -tags='testing debug netgo' -timeout=300s $(integration-pkgs) -run=. -count=$(count) ; \
 	make stop-mongo
+	go mod tidy
```
MSevey, line 112: nit: should this be added to lint instead?

api/handlers.go
```diff
@@ -67,6 +69,55 @@ func (api *API) limitsGET(w http.ResponseWriter, _ *http.Request, _ httprouter.P
...
+	if err != nil {
+		api.staticLogger.Log(logrus.WarnLevel, "Error parsing POST form:", err)
+	}
+	email := req.PostFormValue("email")
+	pw := req.PostFormValue("password")
+	if email != "" && pw != "" {
+		api.loginPOSTCredentials(w, req, email, pw)
+		return
+	}
+	api.loginPOSTToken(w, req)
```
MSevey, line 82: Comments please :-)

api/handlers.go
```diff
@@ -67,6 +69,55 @@ func (api *API) limitsGET(w http.ResponseWriter, _ *http.Request, _ httprouter.P
...
+func (api *API) loginPOSTCredentials(w http.ResponseWriter, req *http.Request, email, password string) {
+	// Fetch the user with that email, if they exist.
+	u, err := api.staticDB.UserByEmail(req.Context(), email, false)
+	if err != nil {
+		api.staticLogger.Tracef("Error fetching a user with email '%s': %+v\n", email, err)
+		api.WriteError(w, err, http.StatusUnauthorized)
+		return
+	}
+	// Check if the password matches.
+	err = hash.Compare(password, []byte(u.PasswordHash))
```
MSevey, line 95: nit: should we do this first to avoid a db query if the password is wrong?

database/user.go
```diff
@@ -151,7 +153,36 @@ func (db *DB) UserBySub(ctx context.Context, sub string, create bool) (*User, er
 	users, err := db.managedUsersByField(ctx, "sub", sub)
 	if create && errors.Contains(err, ErrUserNotFound) {
 		var u *User
-		u, err = db.UserCreate(ctx, sub, TierFree)
+		_, email, err := jwt.UserDetailsFromJWT(ctx)
+		if err != nil {
+			// Log the error but don't do anything differently.
```
MSevey, line 158: If `err != nil` will the `email` be blank and will that cause an issue with `UserCreate`? 

Is there a test that hits this case?

database/user.go
```diff
@@ -225,25 +256,40 @@ func (db *DB) UserByStripeID(ctx context.Context, id string) (*User, error) {
...
+	// Check for an existing user with this sub.
+	if sub != "" {
+		users, err := db.managedUsersByField(ctx, "sub", sub)
+		if err != nil && !errors.Contains(err, ErrUserNotFound) {
+			return nil, errors.AddContext(err, "failed to query DB")
+		}
+		if len(users) > 0 {
+			return nil, ErrUserAlreadyExists
+		}
+	} else {
```
MSevey, line 277: You can invert this can't you?
```
if sub == "" {
  sub = generateSub()
}
users, err := db.managedUsersByField(ctx, "sub", sub)
if err != nil && !errors.Contains(err, ErrUserNotFound) {
	return nil, errors.AddContext(err, "failed to query DB")
}
if len(users) > 0 {
	return nil, ErrUserAlreadyExists
}
```

database/user.go
```diff
@@ -225,25 +256,40 @@ func (db *DB) UserByStripeID(ctx context.Context, id string) (*User, error) {
...
+		sub = generateSub()
+	}
+	var passHash []byte
+	if pass != "" {
```
MSevey, line 281: Is no password accepted? Shouldn't that be an error condition?

MSevey: Comment explaining why not?

database/user.go
```diff
@@ -631,3 +677,22 @@ func monthStart(subscribedUntil time.Time) time.Time {
 	d := now.AddDate(0, monthsDelta, daysDelta)
 	return time.Date(d.Year(), d.Month(), d.Day(), 0, 0, 0, 0, time.UTC)
 }
+
+// generateSub is a helper method that generates a UUID and encodes it in hex.
+func generateSub() string {
+	uid, err := uuid.New()
+	// the only way to get an error here is for uuid to be unable to read from
+	// the RNG reader, which is highly unlikely. We'll log and retry.
```
MSevey, line 685: Highly unlikely is not impossible :-)
What is the harm in returning the error?
We should at least `build.Critical` and return and empty string instead of just logging the error. That is more in line with how we handle similar cases.

database/user.go
```diff
@@ -151,7 +153,36 @@ func (db *DB) UserBySub(ctx context.Context, sub string, create bool) (*User, er
...
+	}
+	if err != nil {
+		return nil, err
+	}
+	return users[0], nil
+}
+
+// UserByEmail returns the user with the given username. If `create` is `true`
+// it will create the user if it doesn't exist.
+func (db *DB) UserByEmail(ctx context.Context, email string, create bool) (*User, error) {
```
MSevey, line 180: REQ: I don't see this tested.

api/handlers.go
```diff
@@ -182,6 +233,38 @@ func (api *API) userStatsGET(w http.ResponseWriter, req *http.Request, _ httprou
 	api.WriteJSON(w, us)
 }
 
+// userPOST creates a new user.
+func (api *API) userPOST(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
```
MSevey, line 237: REQ: Where are the tests for the methods in this file? I don't see updates for the new methods.

api/handlers.go
ro-tex: I like the standard severity levels. Currently `accounts` is running with `trace` level logging but I want to be able to bring that up to `info` and only switch to `debug` and `trace` when I need to.

api/handlers.go
ro-tex: That was my impression as well but I wanted to confirm. I'll only ensure it's not empty.

api/handlers.go
ro-tex: Thanks, this is what I was looking for. :)

api/handlers.go
ro-tex: Ha ha, fair enough! :)

api/handlers.go
ro-tex: We are comparing the password to the hash we get from the DB. If we can do that without fetching the data from the DB first we're gonna revolutionize the entire computer world. :P

database/user.go
ro-tex: TL;DR: No test. It's not a problem. I have a plan. :)

No, there's no test that hits this use case. This entire project could use some more love on the testing side.

If `err!=nil` the `email` will be empty but that won't cause any issues with `UserCreate` outside of being unable to send a confirmation email and the user will be left unable to reset their password. This was implemented to work with Kratos, which took care of all of those aspects. That's why the identity of the user (i.e. their email) was directly taken from the JWT where Kratos put it.

Now that we're removing Kratos we'll need to change our flows a little and have nginx explicitly call our `POST /user` endpoint instead of creating users implicitly by providing their JWTs. That won't be much of a change for `skynet-webportal` because that call is already being made, it's just being made to Kratos.

database/user.go
ro-tex: Yes, I can. I am not sure I really need to, though. Hear me out:

If there is a `sub` passed with the request then there's a good chance that the user is trying to log in or create a user based on an existing identity. That identity might already be registered in the DB, so we want to make sure it's not there before we create a new account with it.

Alternatively, if we're generating the `sub`, the chances for a collision between a new UUID and an existing one is close to zero. Close enough, actually, that we don't need to check - that's the entire premise of UUIDs. That's why I'm skipping the second trip to the DB.

Having in mind that this request doesn't need to be fast, I don't mind checking for a `sub` collision even when we generate it, so I'll just add it.

database/user.go
ro-tex: User management was handled in Kratos until now, so there were several places where we create an `accounts` user based on the passed JWT token, which only contains an email and a sub.

For the moment I am going with the safer assumption which is that we'll be running kratos and the new accounts system together, so it would still be possible for people to create accounts with kratos and get them registered with `accounts`. The bolder assumption would be that Kratos will be entirely gone and accounts will be the only thing managing users. But having in mind that we deploy as a rolling wave, we'll have a heterogeneous system at first, so assuming Kratos still exists seems like the logical thing to do. We'll clean up these parts in a subsequent PR.

I added a comment explaining that.

database/user.go
ro-tex: Hm. Initially I was very reluctant to use `build.Critical` because that's something we use in order to keep an eye on systems we control. `crypto/rand` is definitely not such a system. OTOH, if we ever fail to read from `crypto/rand`'s reader that's probably a non-recoverable error that we might want to `log.Fatal` on, as our node will be unable to create new IDs.

In any case, I'll add the `build.Critical`, although it seems like a bad fit for the case because it's not in a module we control.

database/user.go
ro-tex: It seems like a good moment to overhaul the integration test suite and add tests for the uncovered methods. Tests added.

api/handlers.go
ro-tex: We don't have any tests for the handlers. I'm not sure what would be a good approach to testing them. I don't want to expose them and I don't see how I can test them without the full integration setup. Maybe I can create exported test methods for them within this package and then have those exit right away if the testing type is not integration. And then I can call them from a test in the integration suite.

While this sounds like a good idea, I'm not sure it's what we should be doing right now. Do you mind if I  punt this into a F/U?

api/handlers.go
ro-tex: Switching the logging library might be a good idea, actually. I would just not go for the lib we wrote for `skyd`. Instead I would go for a structured logging library, such as [zap](https://github.com/uber-go/zap) or [zerolog](https://github.com/rs/zerolog).

api/handlers.go
MSevey: oh ha sorry missed that 🙃

api/handlers.go
MSevey: Follow up is fine. We should definitely figure out how to extend the testing so that it hits mongo. I'm sure there are plenty of solutions.

ro-tex: Yes, definitely. We already have a set of integration tests that run against a clean Mongo in a docker container. We just need another set which acts as an end-to-end. Should be simple enough to set up, it will just take some time.

F/U: #69

MSevey (approved):

test/util.go
```diff
@@ -0,0 +1,85 @@
...
+	return RegisterTestUpload(ctx, db, user, skylink)
+}
+
+// RandomSkylink generates a random skylink
+func RandomSkylink() string {
+	sb := strings.Builder{}
+	for i := 0; i < skylinkLen; i++ {
+		_ = sb.WriteByte(skylinkCharset[fastrand.Intn(len(skylinkCharset))])
+	}
+	return sb.String()
```
ChrisSchinnerl, line 68: You could use the following to get a random skylink with valid encoding. Then you also don't need the hacky charset.

```go
var h crypto.Hash
fastrand.Read(h[:])
sl, _ := skymodules.NewSkylinkV1(h, 0, 0)
return sl.String()
```

database/user.go
```diff
@@ -224,26 +229,75 @@ func (db *DB) UserByStripeID(ctx context.Context, id string) (*User, error) {
...
+	}
+	if sub == "" {
+		sub, err = generateSub()
+		if err != nil {
+			return nil, errors.AddContext(err, "failed to generate user sub")
+		}
+	}
 	// Check for an existing user with this sub.
-	users, err := db.managedUsersByField(ctx, "sub", sub)
+	users, err = db.managedUsersByField(ctx, "sub", sub)
```
ChrisSchinnerl, line 277: I think you call this quite a lot. Might warrant a new `managedUsersBySub(ctx, sub)` to reduce the number of hardcoded `"sub"`s

database/user.go
```diff
@@ -631,3 +685,13 @@ func monthStart(subscribedUntil time.Time) time.Time {
...
 }
+
+// generateSub is a helper method that generates a UUID and encodes it in hex.
+func generateSub() (string, error) {
+	uid, err := uuid.New()
+	if err != nil {
+		build.Critical("Error during UUID creation:", err)
+		return "", err
+	}
+	return hex.EncodeToString(uid[:]), nil
```
ChrisSchinnerl, line 696: To avoid the error handling here you could do

Do you need a valid UUID here? Otherwise the following would work and not return error.
```go
var sub [16]byte
fastrand.Read(sub[:])
return hex.EncodeToString(sub[:])
```

test/util.go
```diff
@@ -0,0 +1,85 @@
+package test
```
ChrisSchinnerl, line 1: Seems like the helpers in this file are only used within `test/database/user_test.go`?

You could just move it to `test/database/utils_test.go` in that case. That way it's also not part of the production binary.

database/user.go
```diff
@@ -224,26 +229,75 @@ func (db *DB) UserByStripeID(ctx context.Context, id string) (*User, error) {
 	return &u, nil
 }
 
+// UserBySub returns the user with the given sub. If `create` is `true` it will
+// create the user if it doesn't exist. The sub is the Kratos id of that user.
+func (db *DB) UserBySub(ctx context.Context, sub string, create bool) (*User, error) {
```
ChrisSchinnerl, line 234: Is there a reason why your getter methods come with a `create` bool? This feels pretty strange tbh. 
I'd expect a getter that returns a user and maybe a bool and error to indicates whether the user exists or not and a setter which creates a user and returns an error.

database/user.go
```diff
@@ -224,26 +229,75 @@ func (db *DB) UserByStripeID(ctx context.Context, id string) (*User, error) {
...
+		_, email, err := jwt.UserDetailsFromJWT(ctx)
+		if err != nil {
+			// Log the error but don't do anything differently.
+			db.staticLogger.Debugf("We failed to extract the expected user infotmation from the JWT token. Error: %s", err.Error())
+		}
+		u, err := db.UserCreate(ctx, email, "", sub, TierFree)
+		// If we're successful or hit any error, other than a duplicate key we
+		// want to just return. Hitting a duplicate key error means we ran into
+		// a race condition and we can easily recover from that.
+		if err == nil || !strings.Contains(err.Error(), "E11000 duplicate key error collection") {
```
ChrisSchinnerl, line 246: Are keys for new users not randomly generated? When would we run into a duplicate key error?

database/user.go
ro-tex: Sure, done.

database/user.go
ro-tex: I prefer having an actual UUID.

test/util.go
ro-tex: I didn't want to do that because `utils_test.go` looks like something that's testing a `utils.go` file instead of providing helpers. But I don't want those methods in the bin for no reason.

database/user.go
ro-tex: The reason for that is that `accounts` wasn't responsible for user registration. So, the workflow was (still is):
- user registers with Kratos
- `webportal` maybe tries to call `accounts` in a post-registration hook and tell it about the new user but that might fail and we might end up with the user not in accounts. `webportal` cannot store that anywhere since it doesn't have a DB, so it cannot keep retrying the user registration.
- the user performs an action (upload, download, registry access, etc.) while logged in. Nginx calls `accounts` on the respective `track` endpoint and provides the user's JWT token which contains everything we need in order to register them. Note that nginx cannot try to call `track`, call a `create user` endpoint on error and then retry the `track` request - it can only issue a static set of API calls.
- `accounts` receives the `track` request, looks for the user in the DB and doesn't find it. It decides that it's a new user that it just hasn't been informed about and registers them.
- `accounts` uses a boolean flag in order to be able to use the same methods in situations in which the user we're trying to fetch shouldn't be auto-created.

Once we have removed Kratos and migrated all the data into accounts's MongoDB, we can remove that flag and stop creating users implicitly.

database/user.go
ro-tex: This error is handled in two functions - `UserBySub` and `UserByEmail`.

For `UserBySub` this duplicate key error is possible because the user might have registered with Kratos and then might use the same JWT to make simultaneous request to two different servers. Those can cause a data race resulting in this duplicate key error. We recover from that by just fetching the user from the DB.

For `UserByEmail` the situation is a little different. The `email` field isn't set to be unique right now. I just noticed that because of your comment. We should make that unique because we use it as username. I cannot make it unique right now because we have plenty of users with empty emails in the DB. First we need to do the full data migration and then make emails unique. And after we make them unique we will be in the very same situation where we can get a dup and we can just fetch from the DB.

F/U: #70

ro-tex: F/U: #71

database/user.go
ro-tex: F/U: #71

ChrisSchinnerl (approved):

MSevey (approved):

Issue #63, opened 2021-08-30
Replace Kratos


ro-tex:
**Is your feature request related to a problem? Please describe.**

We would like to stop using Kratos and Oathkeeper.

**Describe the solution you'd like**

Accounts will start offering all necessary features:
- user creation (with passwords stored in MongoDB)
- user login with credentials (email and password)
- issuing JWTs tokens
- supporting JWTs tokens issued by Kratos, so users won't be logged out
- supporting user logins with credentials *stored in Kratos*, so users won't need to re-set their passwords
- account confirmation over email
- password reset over email

**Describe alternatives you've considered**

**Additional context**

#### Data migration

After all the code has been written and merged we need to transition some data from Kratos's CockroachDB into accounts's MongoDB:
- [x] copy over the email and password hash for each user based on their `sub`
- [x] when copying the password, extract the `hashed_password` value instead of copying the entire record
- [x] scrub unused fields from MongoDB: first and last name, etc.

#### Configuration changes

After these changes `accounts` will need two more environment variables:
 - `ACCOUNTS_EMAIL_URI` will give it the necessary configuration to send emails. This should be set on all servers - `accounts` will handle sending each email just once on its own.
 - `ACCOUNTS_JWKS_FILE` will point to the `id_token.jwks.json` file holding the JWKS we want to use to sign and verify JWTs. This file should be identical to the one currently used by Oathkeeper, so we don't invalidate any existing tokens. This var defaults to `/accounts/conf/jwks.json`, so in that case it's optional. The existence of the JWKS file at that location is not.

ro-tex:
This is not done, yet.

ro-tex:
Apparently, lining a PR to an issue closes the issue when you close the PR. Even if there are other PRs linked to the issue which are not closed. That's why I unlinked the PRs that are still in progress and just mentioned this issue in their descriptions. A little pedestrian but it will do.

PR #65, opened 2021-09-01
Sending emails

ro-tex:
This PR introduces the ability to send emails to user. Those will be used to confirm that the user owns the email with which they registered and to request a password reset (only after confirming their email).

Depends on #62 

Part of #63


ro-tex:
This PR was closed because it grew too big to properly review. The changes from it are split into several PRs - #72 and others that are to come.

PR #72, opened 2021-09-09
Sending emails, part 1

ro-tex:
# PULL REQUEST

## Overview

This PR implements sending email and it's a first in a series that implements email communication with users.

## Checklist
 - [x] Changelog file created


database/email.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+	// Fetch up to batchSize messages already locked with lockID.
+	opts := options.Find()
+	opts.SetLimit(batchSize)
+	c, err := db.staticEmails.Find(ctx, filter, opts)
+	if err != nil {
+		return nil, errors.AddContext(err, "failed to fetch emails")
+	}
+	defer func() {
+		if errDef := c.Close(ctx); errDef != nil {
```
MSevey, line 85: Should this error be combined with the normal return error?

database/email.go
```diff
@@ -0,0 +1,200 @@
...
+		"sent_at":   nil,
+	}
+	opts := options.Find()
+	opts.SetLimit(num)
+	c, err := db.staticEmails.Find(ctx, filter, opts)
+	if err != nil {
+		return nil, errors.AddContext(err, "failed to fetch ids")
+	}
+	defer func() {
+		if errDef := c.Close(ctx); errDef != nil {
```
MSevey, line 115: same question on combining errors. Otherwise a comment explaining why not

database/email.go
```diff
@@ -0,0 +1,200 @@
...
+			db.staticLogger.Traceln("Error on closing DB cursor.", errDef)
+		}
+	}()
+	var ids []primitive.ObjectID
+	for c.Next(ctx) {
```
MSevey, line 120: This could be refactored to a method that returns both the ids and the messages.

email/sender.go
```diff
@@ -0,0 +1,103 @@
...
+
+const (
+	// sleepBetweenScans defines how long the sender should sleep between its
+	// sweeps of the DB.
+	sleepBetweenScans = 30 * time.Second
+)
+
+var (
+	// ServerDomain holds the name of the name of this particular server.
+	ServerDomain = "siasky.net"
```
MSevey, line 22: Can we reference the value in the env file for consistency?

email/mailer.go
```diff
@@ -0,0 +1,190 @@
...
+unlock it, so it can be retried later. If a message fails to get sent more than
+`maxAttemptsToSend` times it is marked as `Failed`.
+*/
+
+const (
+	// emailBatchSize defines the largest batch of emails we will try to send.
+	emailBatchSize = 10
+)
+
+var (
```
MSevey, line 43: Should these be in the env file? Thinking about how community portal operators should be updating these.

database/email.go
ro-tex: I don't think so. If we do that we won't send the messages we've already locked. The error is somewhat important to us as the team responsible for the infrastructure but it's not important for the process of sending emails. That's why it's logged but otherwise ignored.

database/email.go
ro-tex: I can add a comment.

ro-tex: Combined.

database/email.go
ro-tex: The reason for the current approach is that I don't have a way to lock a limited number of records, so I want to lock them by their ids. Which means that another server doing the same thing at the same time might decide to update the same records. If that happens and `fetchUnlockedMessageIDs` returns the records instead of the ids, those emails will get sent twice. With the current setup the messages updated by both servers will be sent just once and one of the servers will send fewer emails in its current batch, which is not an issue.

There might be something I'm missing here. It will be easier to discuss that on Discord.

email/sender.go
ro-tex: It's already there but I haven't mentioned it in the docstring. I guess I added it to the env file later.
Updating the docsting.

email/mailer.go
ro-tex: They are. I forgot to mention it in the docstrings.
Good catch, thanks. :)

ro-tex:
The PR is now updated to target `main`.

database/email.go
```diff
@@ -0,0 +1,199 @@
...
+}
+
+// EmailLockAndFetch locks up to batchSize records with the given lockId and
+// returns up to batchSize locked entries. Some of the returned entries might
+// not have been locked during the current execution.
+func (db *DB) EmailLockAndFetch(ctx context.Context, lockID string, batchSize int64) (msgs []EmailMessage, err error) {
+	// Find out how many entries are already locked by this id. Maybe we don't
+	// need to lock any additional ones.
+	filter := bson.M{
+		"locked_by": lockID,
```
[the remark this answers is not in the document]
ChrisSchinnerl: I used this in skyd btw. Saved me a lot of work https://github.com/square/mongo-lock

database/email.go
MSevey: You have the same for loop that iterates over `for c.Next(ctx)` twice https://github.com/SkynetLabs/skynet-accounts/pull/72/files/7f099e6c21a8226a5ccc9f0155e67b09d5b46e06#diff-23fbde68a3027d369c3bf5fcd43e6d5cce2830e5108303066a7d9fc14e0c1358R90

Both for loops decode the EmailMessage. One loop creates a slice of EmailMessages, and one loop creates a slice of ObjectIDs. 

You could just have one helper method that returns both. The looping of the `c.Next(ctx)` feels like something that might happen for a number of reasons so might be worth having pulled out. 

Not blocking, just a thought.

email/sender.go
```diff
@@ -0,0 +1,217 @@
...
+func (s Sender) Start() {
+	go func() {
+		for {
+			select {
+			case <-s.staticCtx.Done():
+				return
+			default:
+			}
+			s.scanAndSend()
+			time.Sleep(sleepBetweenScans)
```
MSevey, line 99: This can just be added to the select statement

```
s.scanAndSend()
select {
case <-s.staticCtx.Done():
case <-time.After(sleepBetweenScans):
}
```

MSevey: Or we could use a `time.AfterFunc` here

MSevey: nit / follow up

MSevey (approved):

database/email.go
ro-tex: I think I misunderstood the original idea. Moving this to a helper makes sense. Will add it in the next PR.

email/sender.go
ro-tex: Sure thing. Changed in the upcoming PR
```go
		for {
			select {
			case <-s.staticCtx.Done():
				return
			case <-time.After(sleepBetweenScans):
				s.scanAndSend()
			}
		}
```

database/email.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+)
+
+// EmailCreate creates an email message in the DB which is waiting to be sent.
+func (db *DB) EmailCreate(ctx context.Context, m EmailMessage) error {
+	fields, err := bson.Marshal(m)
+	if err != nil {
+		return err
+	}
+	_, err = db.staticEmails.InsertOne(ctx, fields)
```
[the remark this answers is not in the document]
ro-tex: Changed.

database/email.go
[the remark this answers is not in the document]
ro-tex: Added.

database/email.go
```diff
@@ -0,0 +1,199 @@
...
+	if err != nil {
+		return errors.AddContext(err, "failed to Insert")
+	}
+	return nil
+}
+
+// EmailLockAndFetch locks up to batchSize records with the given lockId and
+// returns up to batchSize locked entries. Some of the returned entries might
+// not have been locked during the current execution.
+func (db *DB) EmailLockAndFetch(ctx context.Context, lockID string, batchSize int64) (msgs []EmailMessage, err error) {
```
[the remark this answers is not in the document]
ro-tex: Done.

MSevey (dismissed):
As a follow up, is there a chance the `abortTxn` flow can be moved to a defer statement?

ro-tex:
> As a follow up, is there a chance the `abortTxn` flow can be moved to a defer statement?

Yes! This was my right-before-falling-asleep thought last night as well, ha ha.

api/handlers.go
```diff
@@ -662,6 +798,36 @@ func (api *API) wellKnownJwksGET(w http.ResponseWriter, _ *http.Request, _ httpr
...
+	if err != nil {
+		api.staticLogger.Warningln("Failed to commit transaction. Error:", err.Error())
+	}
+}
+
+// startTxn is a helper method which starts a Mongo Session and then starts a
+// new transaction with it.
+func startTxn(ctx context.Context) (mongo.SessionContext, error) {
+	sess := mongo.SessionFromContext(ctx)
+	sctx := mongo.NewSessionContext(ctx, sess)
```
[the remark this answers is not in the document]
ChrisSchinnerl: This is also very easy to unit test since you can just add a dummy handler which you wrap with `WithDBSession` that only writes a single thing to the db and then returns an error one time and a success the other time.
So you just check whether the txn happened or not and then you can trust that as long as you pass the right context to the db, all modifications within a handler happen atomically.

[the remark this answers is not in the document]
ro-tex: Added.

database/database.go
```diff
@@ -45,7 +52,7 @@ var (
 	// mongoReadPreference defines the DB's read preference. The options are:
 	// primary, primaryPreferred, secondary, secondaryPreferred, nearest.
 	// See https://docs.mongodb.com/manual/core/read-preference/
-	mongoReadPreference = "nearest"
+	mongoReadPreference = "primary"
```
ro-tex, line 55: This change is required for transactions. It's unfortunate but if we want transactions, we need to read from the primary.

api/mongowriter.go
```diff
@@ -0,0 +1,51 @@
...
+
+// Write implements the ResponseWriter interface.
+func (mw *MongoWriter) Write(bytes []byte) (int, error) {
+	return mw.w.Write(bytes)
+}
+
+// WriteHeader writes the header and finalises the transaction.
+func (mw *MongoWriter) WriteHeader(statusCode int) {
+	mw.w.WriteHeader(statusCode)
+	if statusCode > 299 {
```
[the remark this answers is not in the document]
ro-tex: Yes. :)

email/sender.go
```diff
@@ -0,0 +1,263 @@
...
+	_, msgsToLock, err := s.staticDB.FindEmails(s.staticCtx, filter, opts)
+	// Lock them.
+	var msgs []database.EmailMessage
+	ld := lock.LockDetails{
+		Owner: ServerLockID,
+		TTL:   lockTTL,
+	}
+	for _, m := range msgsToLock {
+		err = s.staticDB.LockClient.XLock(s.staticCtx, m.ID.Hex(), ServerLockID, ld)
+		if err != nil {
```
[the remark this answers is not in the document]
ro-tex: Makes sense.

database/email.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+	return ids, msgs, nil
+}
+
+// MarkAsSent unlocks all given messages and marks them as sent.
+func (db *DB) MarkAsSent(ctx context.Context, ids []primitive.ObjectID) error {
+	if len(ids) == 0 {
+		return nil
+	}
+	filter := bson.M{"_id": bson.M{"$in": ids}}
```
peterjan, line 73: should we ensure we don't overwrite `sent_at` by adding `sent_at: null` in the predicate?

database/email.go
```diff
@@ -0,0 +1,128 @@
...
+		return nil
+	}
+	ids := make([]primitive.ObjectID, len(msgs))
+	var failed []primitive.ObjectID
+	var tryAgain []primitive.ObjectID
+	for i, m := range msgs {
+		ids[i] = m.ID
+		// the messages that are about to reach or exceed the limit will be
+		// marked as failed and won't be retries again.
+		if m.FailedAttemptsToSend >= maxAttemptsToSend-1 {
```
peterjan, line 100: this is a pretty weird check right, shouldn't it be `m.FailedAttemptsToSend == maxAttempts`

email/sender.go
```diff
@@ -0,0 +1,267 @@
...
+
+// scanAndSend scans the database for email messages waiting to be sent and
+// sends them.
+//
+// We lock the messages before sending them and update their SentAt field after
+// sending them. We also don't lock more than batchSize messages.
+func (s Sender) scanAndSend() {
+	// Fetch a batch of email messages, waiting to be sent.
+	filter := bson.M{
+		"failed":  bson.M{"$ne": true},
```
peterjan, line 132: needs an index here as well

email/sender.go
```diff
@@ -0,0 +1,267 @@
...
+	m := mail.NewMessage()
+	m.SetHeader("From", from)
+	m.SetHeader("To", to)
+	m.SetHeader("Subject", subject)
+	m.SetBody(bodyMime, body)
+
+	return s.sendMultiple(m)
+}
+
+// send one or more email messages.
```
peterjan, line 214: `sendMultiple` docstring mismatch

test/fixtures/mongo_keyfile
```diff
@@ -0,0 +1,16 @@
+PUQGGGC+g1sxoYzKKnX7531pos0WvDCYlx1Ak5ZupX3KvyuU2iVcC9puJJSG06VR
```
peterjan, line 1: this is safe to commit right?

email/sender.go
```diff
@@ -0,0 +1,267 @@
...
+	// batchSize defines the largest batch of emails we will try to send.
+	batchSize = 10
+
+	// lockTTL defines the TTL (in seconds) of the lock on the emails waiting to
+	// be sent.
+	lockTTL = 30
+
+	// sleepBetweenScans defines how long the sender should sleep between its
+	// sweeps of the DB.
+	sleepBetweenScans = 10 * time.Second
```
peterjan, line 33: seems high

database/email.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+// MarkAsFailed increments the FailedAttemptsToSend counter on each message and
+// marks the message as Failed if that counter exceeds the maxAttemptsToSend.
+// It also unlocks all given messages.
+func (db *DB) MarkAsFailed(ctx context.Context, msgs []*EmailMessage) error {
+	if len(msgs) == 0 {
+		return nil
+	}
+	ids := make([]primitive.ObjectID, len(msgs))
```
peterjan, line 93: why are you collecting `ids`?

database/email.go
```diff
@@ -0,0 +1,128 @@
...
+const (
+	// maxAttemptsToSend defines the maximum number of attempts we will make to
+	// send a given email message before giving up.
+	maxAttemptsToSend = 3
+)
+
+type (
+	// EmailMessage represents an email message waiting to be sent
+	EmailMessage struct {
+		ID                   primitive.ObjectID `bson:"_id,omitempty"`
```
peterjan, line 22: why are some fields `omitempty`?

database/email.go
```diff
@@ -0,0 +1,128 @@
...
+
+	// Mark all messages that exceeded the limit on failures as permanently
+	// failed.
+	var errFailed error
+	if len(failed) > 0 {
+		db.staticLogger.Warningf("%d email messages failed to be sent more than %d times and won't be tried anymore.", len(failed), maxAttemptsToSend)
+
+		filter = bson.M{"_id": bson.M{"$in": failed}}
+		update = bson.M{
+			"$inc": bson.M{"failed_attempts_to_send": 1},
```
peterjan, line 122: could've avoided having two fields by having `failed_attempts` and init it at zero, and then do `failed: {$lt: maxFailedAttempts}` in a predicate that fetches emails that need to be retried

peterjan (changes requested):
only requesting changes for the missing index, otherwise it LGTM

Makefile
```diff
@@ -66,13 +66,17 @@ start-mongo:
...
-     -p 127.0.0.1:17017:27017 \
+     -p 17017:17017 \
      -e MONGO_INITDB_ROOT_USERNAME=admin \
      -e MONGO_INITDB_ROOT_PASSWORD=aO4tV5tC1oU3oQ7u \
-     mongo
+     -v $(shell pwd)/test/fixtures/mongo_keyfile:/data/mgkey \
+	mongo:4.4.1 mongod --port=17017 --replSet=skynet --keyFile=/data/mgkey
+	sleep 3 # wait for mongo to start before we try to configure it
+	# Initialise a single node replica set.
+	docker exec skynet-accounts-mongo-test-db mongo -u admin -p aO4tV5tC1oU3oQ7u --port 17017 --eval "rs.initiate({_id: \"skynet\", members: [{ _id: 0, host: \"localhost:17017\" }]})"
```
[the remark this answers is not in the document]
ro-tex: Yes. This thing is spinning up a test container that's exposing its port on the host system. If the host system is running another mongo container (as mine is 90% of the time) this will fail for a silly and mundane reason. So, I switched it to a custom port in order to avoid that.

api/cookie.go
```diff
@@ -27,10 +28,15 @@ const (
 var (
 	secureCookie = func() *securecookie.SecureCookie {
 		_ = godotenv.Load()
+		hashKeyStr := os.Getenv(envCookieHashKey)
+		encKeyStr := os.Getenv(envCookieEncKey)
+		if len(hashKeyStr) < 32 || len(encKeyStr) < 32 {
+			panic(fmt.Sprintf("Both %s and %s are required environment variables and need to contain at least 32 bytes of hex-encoded entropy. %s, %s", envCookieHashKey, envCookieEncKey, hashKeyStr, encKeyStr))
+		}
 		// These keys need to be *exactly* 16 or 32 bytes long.
```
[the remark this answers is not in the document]
ro-tex: We can but it doesn't really matter. We'll take what we need even if there is more data there. This makes it OK for you to test by just holding the `A` key down for a bit without counting characters.

database/email.go
ro-tex: If the message id reaches this line it means it was just sent. In the worst case we will update a `sent_at` value that's milliseconds before the current one. The real issue here would be the fact that an email was sent multiple times. And even that is not a real problem. We definitely want to avoid it but even if it happens, it's not an issue.

database/email.go
ro-tex: The reason for this check is the fact that we increment the field in the next DB call. If I increment in one call and then check the value and mark the message as failed in another, we would have what you describe. But if we don't want to update records twice (and either re-fetch them from the DB or scan the entire collection) we need to do it this way.

email/sender.go
ro-tex: Added.

test/fixtures/mongo_keyfile
ro-tex: Yes. This is not our key file, it's the key file we use to connect to the test mongodb replica set which always starts empty.

email/sender.go
ro-tex: What would be a more adequate value? 1 second?

database/email.go
ro-tex: Oh, nice catch! Dead code. I was using them for locking before.

database/email.go
ro-tex: If you don't set the `_id` filed to `omitempty` you need to generate the ID yourself, otherwise the record gets inserted in the DB with a `00000000000000` ID which is not what we want. The same is true for `sent_at` where we won't be able to check if it's `nil` in the query but we'll need to compare it to a specific zero value.

database/email.go
ro-tex: Good idea.

peterjan (approved):
LGTM 👍

ChrisSchinnerl (approved):

PR #78, opened 2021-09-29
Handler tests

ro-tex:
# PULL REQUEST

## Overview

This PR sets up some infrastructure for testing handlers and also add a base set for integration tests for the existing handlers. Those tests are not meant to be fully exhaustive but more a sane base line on which we can build in the future.

## Checklist
 - [ ] Changelog file created


test/tester.go
```diff
@@ -0,0 +1,206 @@
...
+	return processResponse(r)
+}
+
+// Put executes a PUT request against the test service.
+func (at *AccountsTester) Put(endpoint string, params map[string]string, putParams map[string]string) (r *http.Response, body []byte, err error) {
+	return at.executeRequest(http.MethodPut, endpoint, params, putParams)
+}
+
+// Shutdown performs a graceful shutdown of the AccountsTester service.
+func (at *AccountsTester) Shutdown() {
```
ChrisSchinnerl, line 159: This is not blocking for anything to actually shut down so I'd call it `Close() error` to implement the `io.Closer` interface and make people aware that this needs to be closed. It's easier to oversee a `Shutdown` method and it's also very unique to things like the `http.Server` which need to do some more work for a shutdown.

test/tester.go
```diff
@@ -0,0 +1,206 @@
...
+	}, nil
+}
+
+// Get executes a GET request against the test service.
+func (at *AccountsTester) Get(endpoint string, params map[string]string) (r *http.Response, body []byte, err error) {
+	return at.executeRequest(http.MethodGet, endpoint, params, nil)
+}
+
+// Delete executes a DELETE request against the test service.
+func (at *AccountsTester) Delete(endpoint string, params map[string]string) (r *http.Response, body []byte, err error) {
```
ChrisSchinnerl, line 126: I'm not yet sure why you'd need to expose low-level endpoints like `Delete`, `Post` and `Put` instead of creating helpers which talk to a specific endpoint and are named accordingly but either way I'd just replace `params` with a `url.Values` if you are going to encode them anyway.

test/api/handlers_test.go
```diff
@@ -0,0 +1,813 @@
...
+
+const (
+	// Helper consts for checking the returned errors.
+	badRequest   = "400 Bad Request"
+	unauthorized = "401 Unauthorized"
+)
+
+type subtest struct {
+	Name string
+	Test func(t *testing.T, at *test.AccountsTester)
```
ChrisSchinnerl, line 32: Any reason why these fields are exposed?

test/api/handlers_test.go
```diff
@@ -0,0 +1,813 @@
...
+	at, err := test.NewAccountsTester()
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer at.Shutdown()
+
+	// Specify subtests to run
+	tests := []subtest{
+		// GET /health
+		{Name: "health", Test: testHandlerHealthGET},
```
ChrisSchinnerl, line 47: We usually use uppercase names in skyd. I'd stick to that convention.

test/api/handlers_test.go
```diff
@@ -0,0 +1,813 @@
...
+		t.Fatal("DB down.")
+	}
+}
+
+// testHandlerUserPOST tests user creation and login.
+func testHandlerUserPOST(t *testing.T, at *test.AccountsTester) {
+	// Use the test's name as an email-compatible identifier.
+	name := strings.ReplaceAll(t.Name(), "/", "_")
+	email := name + "@siasky.net"
+	params := map[string]string{
```
ChrisSchinnerl, line 101: I'd build the map inside of `Post` and have a `at.CreateUserPost` method which accepts an email and password.

test/api/handlers_test.go
```diff
@@ -0,0 +1,813 @@
...
+		"email":    email,
+		"password": hex.EncodeToString(fastrand.Bytes(16)),
+	}
+	// Try to create a user without passing params.
+	_, _, err := at.Post("/user", nil, nil)
```
ChrisSchinnerl, line 106: I'd also test the empy string email and password.

ChrisSchinnerl (changes requested):

Makefile
```diff
@@ -122,7 +122,17 @@ test-long: clean fmt vet lint lint-ci
 test-int: export COOKIE_HASH_KEY="7eb32cfab5014d14394648dae1cf4e606727eee2267f6a50213cd842e61c5bce"
 test-int: export COOKIE_ENC_KEY="65d31d12b80fc57df16d84c02a9bb62e2bc3b633388b05e49ef8abfdf0d35cf3"
 test-int: test-long start-mongo
-	GORACE='$(racevars)' go test -race -v -tags='testing debug netgo' -timeout=300s $(integration-pkgs) -run=. -count=$(count) ; \
-	make stop-mongo
-
-.PHONY: all fmt install release clean check test test-int test-long stop-mongo
+	GORACE='$(racevars)' go test -race -v -tags='testing debug netgo' -timeout=300s $(integration-pkgs) -run=. -count=$(count)
+	# Run the email_sender tests separately, without parallelism (`-p 1`).
+	CONTENDING_SENDERS='run' GORACE='$(racevars)' go test -p 1 -race -v -tags='testing debug netgo' -timeout=300s ./test/email -run=TestContendingSenders -count=$(count)
```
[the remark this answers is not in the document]
ro-tex: It does, I'm afraid.
`The 'go test' command may run tests for different packages in parallel as well, according to the setting of the -p flag (see 'go help build').`

Source: https://pkg.go.dev/cmd/go/internal/test
Article: https://medium.com/@xcoulon/how-to-avoid-parallel-execution-of-tests-in-golang-763d32d88eec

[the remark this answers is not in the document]
ro-tex: Re: running tests in separate databases: At first this sounded a little over the top but then... I like it! Implemented. Works great. :D

test/utils.go
```diff
@@ -19,3 +36,115 @@ func DBTestCredentials() database.DBCredentials {
...
+	_, _, err := at.Post("/user", nil, params)
+	if err != nil {
+		return nil, nil, errors.AddContext(err, "user creation failed")
+	}
+	// Fetch the user from the DB, so we can delete it later.
+	u, err := at.DB.UserByEmail(at.Ctx, email, false)
+	if err != nil {
+		return nil, nil, errors.AddContext(err, "failed to fetch user from the DB")
+	}
+	cleanup := func(user *database.User) {
```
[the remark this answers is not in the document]
ro-tex: Old habits die hard, I guess.

This approach resolves the issue that forced me to pass the `*testing.T` param in https://github.com/SkynetLabs/skynet-accounts/pull/78#discussion_r721423178, so that's great.

test/utils.go
```diff
@@ -19,3 +36,115 @@ func DBTestCredentials() database.DBCredentials {
 		Port:     "17017",
 	}
 }
+
+// CreateUser is a helper method which simplifies the creation of test users
+func CreateUser(t *testing.T, at *AccountsTester, customEmail, customPassword string) (*database.User, func(user *database.User), error) {
```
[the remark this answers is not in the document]
ro-tex: Done.

test/tester.go
```diff
@@ -0,0 +1,206 @@
...
+		// If set, this cookie will be attached to all requests.
+		Cookie *http.Cookie
+
+		cancel context.CancelFunc
+	}
+)
+
+// ExtractCookie is a helper method which extracts the login cookie from a
+// response, so we can use it with future requests while testing.
+func ExtractCookie(r *http.Response) *http.Cookie {
```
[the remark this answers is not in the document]
ro-tex: Yes, I use this extensively in `handlers_test.go`. The idea is to freely choose when you have a cookie set and when you don't.

test/api/handlers_test.go
ro-tex: Not really.

test/tester.go
ro-tex: Low-level access: I can have a small and really useful tester instead of dozens and dozens of repetitive helpers. I really prefer this approach.

`url.Values`: I chose `map[string]string` for its flexibility and the ability to quickly inline it within the method calls. But I don't mind using `url.Values`. A lot more typing for no real gain, IMO.

test/tester.go
ro-tex: I don't see how anyone will notice that this is an `io.Closer` without looking through the methods. And if they look through the methods and don't call the `Shutdown` one then I'm not sure a `Close` one would help.

But sure. It wouldn't hurt to try.

test/utils.go
```diff
@@ -19,3 +36,115 @@ func DBTestCredentials() database.DBCredentials {
...
+	c := ExtractCookie(r)
+	if c == nil {
+		return nil, nil, nil, err
+	}
+	return u, c, cleanup, nil
+}
+
+// CreateTestUpload creates a new skyfile and uploads it under the given user's
+// account. Returns the skylink, the upload's id and error.
+func CreateTestUpload(ctx context.Context, db *database.DB, user *database.User, size int64) (*database.Skylink, primitive.ObjectID, error) {
```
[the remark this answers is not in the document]
ro-tex: I'm using `CreateTestUpload` from various places, not necessarily with with the new `testUser` struct. I would need to specifically construct that in order to call it. This is more confusing to me than having a simple helper method that has an obvious way to be called.

test/tester.go
```diff
@@ -0,0 +1,206 @@
...
+	testPortalAddr = "http://127.0.0.1"
+	testPortalPort = "6000"
+	pathToJWKSFile = "../../jwt/fixtures/jwks.json"
+)
+
+type (
+	// AccountsTester is a simple testing kit for accounts. It starts a testing
+	// instance of the service and provides simplified ways to call the handlers.
+	AccountsTester struct {
+		Ctx    context.Context
```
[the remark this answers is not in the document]
ro-tex: Unexporting a field and locking it behind an exported getter is a standard practice, yes. The reason for that is that often we want to limit or validate the way users interact with otherwise sensitive internal fields. Having in mind that this is a simple tester, I don't think the overhead of adding extra methods here is worth it. I would rather keep it as simple as possible, unless there are validations that I keep doing. If that's the case I'd put those in a getter at that point.

ChrisSchinnerl (approved):

PR #81, opened 2021-10-04
Add caching to the /user/limits endpoint.

ro-tex:
# PULL REQUEST

## Overview

We want to cache the `/users/limits` endpoint in order to minimise the number of database hits caused by it. It's a very frequently hit endpoint, so it has an outsized impact on the system/database load and service performance.

## Checklist
 - [ ] Changelog file created

### DIFF
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -19,6 +19,20 @@ import (
 	"gitlab.com/NebulousLabs/errors"
 )
 
+const (
+	// userLimitCacheTTL is the TTL of the entries in the userLimitCache.
+	userLimitCacheTTL = time.Hour
+)
+
+var (
+	// userLimitCacheEntry is an in-mem cache that maps from a user's sub to
+	// the data we care about.
+	userLimitCache = map[string]userLimitCacheEntry{}
+	// userLimits is a caching entry to avoid doing the same static operation
+	// over and over.
+	userLimits []TierLimitsPublic
+)
+
 type (
 	// LimitsPublic provides public information of the various limits this
 	// portal has.
@@ -35,6 +49,13 @@ type (
 		MaxNumberUploads  int    `json:"maxNumberUploads"`
 		RegistryDelay     int    `json:"registryDelay"` // ms
 		Storage           int64  `json:"storageLimit"`
+	}
+
+	// userLimitCacheEntry allows us to cache some basic information about the
+	// user, so we don't need to hit the DB to fetch data that rarely changes.
+	userLimitCacheEntry struct {
+		Tier       int
+		LastUpdate time.Time
 	}
 
 	// userUpdateData defines the fields of the User record that can be changed
@@ -57,20 +78,22 @@ func (api *API) healthGET(w http.ResponseWriter, req *http.Request, _ httprouter
 
 // limitsGET returns the speed limits of this portal.
 func (api *API) limitsGET(w http.ResponseWriter, _ *http.Request, _ httprouter.Params) {
-	ul := make([]TierLimitsPublic, len(database.UserLimits))
-	for i, t := range database.UserLimits {
-		ul[i] = TierLimitsPublic{
-			TierName:          t.TierName,
-			UploadBandwidth:   t.UploadBandwidth * 8,   // convert from bytes
-			DownloadBandwidth: t.DownloadBandwidth * 8, // convert from bytes
-			MaxUploadSize:     t.MaxUploadSize,
-			MaxNumberUploads:  t.MaxNumberUploads,
-			RegistryDelay:     t.RegistryDelay,
-			Storage:           t.Storage,
+	if userLimits == nil {
+		userLimits = make([]TierLimitsPublic, len(database.UserLimits))
+		for i, t := range database.UserLimits {
+			userLimits[i] = TierLimitsPublic{
+				TierName:          t.TierName,
+				UploadBandwidth:   t.UploadBandwidth * 8,   // convert from bytes
+				DownloadBandwidth: t.DownloadBandwidth * 8, // convert from bytes
+				MaxUploadSize:     t.MaxUploadSize,
+				MaxNumberUploads:  t.MaxNumberUploads,
+				RegistryDelay:     t.RegistryDelay,
+				Storage:           t.Storage,
+			}
 		}
 	}
 	resp := LimitsPublic{
-		UserLimits: ul,
+		UserLimits: userLimits,
 	}
 	api.WriteJSON(w, resp)
 }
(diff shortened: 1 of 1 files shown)


api/handlers.go
```diff
@@ -212,12 +235,45 @@ func (api *API) userGET(w http.ResponseWriter, req *http.Request, _ httprouter.P
...
+	s, exists := tk.Get("sub")
+	if !exists {
+		api.staticLogger.Tracef("Token without a sub.")
+		api.WriteJSON(w, database.TierAnonymous)
+		return
+	}
+	sub := s.(string)
+	// If the user is not cached, or they were cached too long ago we'll fetch
+	// their data from the DB.
+	if c, exists := userLimitCache[sub]; !exists || c.LastUpdate.Before(time.Now().Add(-userLimitCacheTTL).UTC()) {
```
MSevey, line 257:
```suggestion
	if c, exists := userLimitCache[sub]; !exists || c.LastUpdate.Since(time.Now()) > userLimitCacheTTL {
```

MSevey:
No testing makes me sad 😞

api/handlers.go
ro-tex: It turns out it's `c.LastUpdate.Sub(time.Now()) > userLimitCacheTTL`

api/handlers.go
```diff
@@ -19,6 +19,20 @@ import (
...
 
+const (
+	// userLimitCacheTTL is the TTL of the entries in the userLimitCache.
+	userLimitCacheTTL = time.Hour
+)
+
+var (
+	// userLimitCacheEntry is an in-mem cache that maps from a user's sub to
+	// the data we care about.
+	userLimitCache = map[string]userLimitCacheEntry{}
```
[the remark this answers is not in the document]
ro-tex: Mutex added.

ro-tex:
> No testing makes me sad 😞

Yes, me too. But I have a large PR going through review right now that's introducing handler testing and I don't want to come up with a different way for doing it while waiting for it. Maybe I should wait for that PR to be merged, so I can use it here.

api/handlers.go
```diff
@@ -57,20 +80,22 @@ func (api *API) healthGET(w http.ResponseWriter, req *http.Request, _ httprouter
...
-	for i, t := range database.UserLimits {
-		ul[i] = TierLimitsPublic{
-			TierName:          t.TierName,
-			UploadBandwidth:   t.UploadBandwidth * 8,   // convert from bytes
-			DownloadBandwidth: t.DownloadBandwidth * 8, // convert from bytes
-			MaxUploadSize:     t.MaxUploadSize,
-			MaxNumberUploads:  t.MaxNumberUploads,
-			RegistryDelay:     t.RegistryDelay,
-			Storage:           t.Storage,
+	if userLimits == nil {
```
MSevey, line 83: REQ: should be handled under lock

MSevey (dismissed):

api/handlers.go
```diff
@@ -19,6 +20,22 @@ import (
...
+)
+
+var (
+	// userLimitCacheEntry is an in-mem cache that maps from a user's sub to
+	// the data we care about.
+	userLimitCache   = map[string]userLimitCacheEntry{}
+	userLimitCacheMu sync.Mutex
+	// userLimits is a caching entry to avoid doing the same static operation
+	// over and over.
+	userLimits   []TierLimitsPublic
```
ChrisSchinnerl, line 35: This seems like something that could just be a `staticUserLimits` field in the `API` which is pulled from the db when you call `New(...) (*API, error)`

api/handlers.go
```diff
@@ -19,6 +20,22 @@ import (
...
+const (
+	// userLimitCacheTTL is the TTL of the entries in the userLimitCache.
+	userLimitCacheTTL = time.Hour
+)
+
+var (
+	// userLimitCacheEntry is an in-mem cache that maps from a user's sub to
+	// the data we care about.
+	userLimitCache   = map[string]userLimitCacheEntry{}
+	userLimitCacheMu sync.Mutex
```
ChrisSchinnerl, line 32: Similar thing as below. You are caching to be able to faster respond to api requests so this should be in the `API`. 
It's extremely rare that you ever need to put something in the global namespace except for constants or vars which are constants but can't be set as constants for some reason.

api/handlers.go
```diff
@@ -212,12 +240,50 @@ func (api *API) userGET(w http.ResponseWriter, req *http.Request, _ httprouter.P
...
+		return
+	}
+	sub := s.(string)
+	// If the user is not cached, or they were cached too long ago we'll fetch
+	// their data from the DB.
+	userLimitCacheMu.Lock()
+	defer userLimitCacheMu.Unlock()
+	if c, exists := userLimitCache[sub]; !exists || c.LastUpdate.Sub(time.Now()) > userLimitCacheTTL {
+		// we don't want to hold the lock during a DB operation
+		userLimitCacheMu.Unlock()
```
ChrisSchinnerl, line 266: Pls don't unlock a mutex that is set to be unlocked in a `defer`.  That really hurts readability.

In this case it's probably best to just do the old fashioned

```go
userLimitCacheMu.Lock()
c, exists := userLimitCache[sub]
userLimitCacheMu.Unlock()

if !exists || c.LastUpdate.Sub(time.Now()) > userLimitCacheTTL {
```

Although in this case updating the cache already requires enough logic to warrant a refactor into a 
```go
type userLimitCache struct {
  // some map 
  // some mutex
}

func (c *userLimitCache) Get(...) {}
func (c *userLimitCache) Update(...) {}
```

api/handlers.go
```diff
@@ -212,12 +240,50 @@ func (api *API) userGET(w http.ResponseWriter, req *http.Request, _ httprouter.P
...
+		userLimitCacheMu.Lock()
+		if err != nil {
+			api.staticLogger.Debugf("Failed to fetch user from DB for sub '%s'. Error: %s", sub, err.Error())
+			api.WriteJSON(w, database.TierAnonymous)
+			return
+		}
+		if u == nil || u.QuotaExceeded {
+			userLimitCache[sub] = userLimitCacheEntry{
+				Tier:       database.TierAnonymous,
+				LastUpdate: time.Now().UTC(),
```
ChrisSchinnerl, line 277: My spider-sense tells me that there is some danger here.

I think you need to make sure that you set `LastUpdate` in a way that the cache invalidation aligns with a potential quota refill at the end of the month. Otherwise the user might end up getting the Anonymous tier for 1 hour after they should already be back to the tier they pay for.

ChrisSchinnerl:
I'd hold off on merging this one until you can get the testing in.

ChrisSchinnerl (changes requested):

api/cookie.go
```diff
@@ -30,6 +32,14 @@ var (
 		_ = godotenv.Load()
 		hashKeyStr := os.Getenv(envCookieHashKey)
 		encKeyStr := os.Getenv(envCookieEncKey)
+		if build.Release == "testing" {
+			if len(hashKeyStr) < 32 {
+				hashKeyStr = string(fastrand.Bytes(32))
+			}
+			if len(encKeyStr) < 32 {
+				encKeyStr = string(fastrand.Bytes(32))
```
ChrisSchinnerl, line 40: I'd use `crypto.HashSize` or some other constant here instead of 32.

api/stripe.go
```diff
@@ -260,6 +260,8 @@ func (api *API) processStripeSub(ctx context.Context, s *stripe.Subscription) er
 	if err == nil {
 		api.staticLogger.Tracef("Subscribed user id %s, tier %d, until %s.", u.ID, u.Tier, u.SubscribedUntil.String())
 	}
+	// Re-set the tier cache for this user, in case their tier changed.
+	api.staticUserTierCache.Set(u)
```
ChrisSchinnerl, line 264: we should have an integration test for this. To make sure if you used up your tier but the next month starts, you are back to your paid tier instead of the anonymous one.

api/cache.go
```diff
@@ -0,0 +1,83 @@
...
+		Tier      int
+		ExpiresAt time.Time
+	}
+)
+
+// NewUserTierCache creates a new userTierCache.
+func NewUserTierCache() *userTierCache {
+	return &userTierCache{
+		cache: make(map[string]userTierCacheEntry),
+		mu:    sync.Mutex{},
```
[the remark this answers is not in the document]
MSevey: Tests will tell :-)
But no I don't think it is problematic, just more of an fyi

PR #83, opened 2021-10-13
Implement challenge-response endpoints

ro-tex:
# PULL REQUEST

## Overview

This PR implemented the basic endpoints needed for challenge-response registration and login.
The endpoints for associating a MySky identity with an already existing account, as well as the recovery endpoints will come in a later PR.

## Checklist
 - [x] Changelog file created

### DIFF
--- a/database/user.go
+++ b/database/user.go
@@ -440,7 +440,26 @@ func (db *DB) UserDelete(ctx context.Context, u *User) error {
 	if u.ID.IsZero() {
 		return errors.AddContext(ErrUserNotFound, "user struct not fully initialised")
 	}
-	filter := bson.D{{"_id", u.ID}}
+	// Delete all data associated with this user.
+	filter := bson.D{{"user_id", u.ID}}
+	_, err := db.staticDownloads.DeleteMany(ctx, filter)
+	if err != nil {
+		return errors.AddContext(err, "failed to delete user downloads")
+	}
+	_, err = db.staticUploads.DeleteMany(ctx, filter)
+	if err != nil {
+		return errors.AddContext(err, "failed to delete user uploads")
+	}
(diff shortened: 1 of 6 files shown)


database/user.go
```diff
@@ -368,6 +383,76 @@ func (db *DB) UserCreate(ctx context.Context, emailAddr, pass, sub string, tier
...
+	if len(users) > 0 {
+		return nil, ErrUserAlreadyExists
+	}
+	// Generate a password hash, if a password is provided. A password might not
+	// be provided if the user intends to only use pubkey authentication.
+	var passHash []byte
+	if pass != "" {
+		passHash, err = hash.Generate(pass)
+		if err != nil {
+			return nil, errors.AddContext(ErrGeneralInternalFailure, "failed to generate password")
```
m-cat, line 426:
```suggestion
			return nil, errors.AddContext(ErrGeneralInternalFailure, "failed to generate password hash")
```

m-cat (approved):

m-cat (dismissed):

DavidVorick:
I'm guessing this: https://hackmd.io/7f0rlCtERFaYjJ4mEUN5sw?view

+ using different keys for different portals is a good idea
+ if possible mysky should cache the derivative seeds (probably in local storage?) so we don't have use the seed too often
+ looks underspecified, mentions a 'type+recipient' in the response signature, but doesn't say what those fields are or where they come from

ro-tex:
> I'm guessing this: https://hackmd.io/7f0rlCtERFaYjJ4mEUN5sw?view
> 
>     * using different keys for different portals is a good idea
> 
>     * if possible mysky should cache the derivative seeds (probably in local storage?) so we don't have use the seed too often

@m-cat will be in better position to comment on those and/or expand the spec.

PR #85, opened 2021-10-15
Add a fallback to CockroachDB for user login.

ro-tex:
# PULL REQUEST

## Overview

Our goal is to move all user data from CockroachDB to MongoDB but since we're doing that on a live system, people will be registering while we do that. Which means that passwords will be missing from MongoDB. This fallback allows us to dynamically move accounts from CockroachDB to MongoDB on login.

The connection to Cockroach is left optional, dependent on the presence of configuration. This will allows us to stop using it by simply removing the config from the `.env` file.

**Caveat**: If someone changes their password between the transfer of data from CockroachDB to MongoDB then we won't be able to handle that. This will only be a problem if it's done from a server that is running the old portal code which is talking to Kratos and not to accounts.

## Checklist
 - [ ] Changelog file created


api/handlers.go
```diff
@@ -87,11 +88,56 @@ func (api *API) loginPOST(w http.ResponseWriter, req *http.Request, _ httprouter
...
+		copy(u.ID[:], fastrand.Bytes(len(u.ID)))
+		err = api.staticDB.UserSave(req.Context(), u)
+		if err != nil {
+			api.staticLogger.Warnf("Failed to move user from CockroachDB: %s", err)
+			api.WriteError(w, database.ErrUserNotFound, http.StatusUnauthorized)
+			return
+		}
+		api.staticLogger.Debugf("User with email '%s' moved from CockroachDB.", u.Email)
+	}
+	// Check if we somehow have an empty password hash and try to fill that.
```
MSevey, line 124: What is the edge case that would cause this? 

Looking at the code it shouldn't be because it was blank in cockroach, as you just pulled it so pulling it again would also be blank.

If it is blank from mongo, why wouldn't that just be an error?

PR #87, opened 2021-10-19
Add an endpoint for deleting current user.

ro-tex:
# PULL REQUEST

## Overview

This PR implements `DELETE /user` which we will use in order to allow users to delete their accounts and all associated data.

## Checklist
 - [x] Changelog file created

### DIFF
--- a/test/api/handlers_test.go
+++ b/test/api/handlers_test.go
@@ -134,7 +134,7 @@ func testHandlerUserPOST(t *testing.T, at *test.AccountsTester) {
 		t.Fatal("User creation failed. Error ", err.Error())
 	}
 	// Make sure the user exists in the DB.
-	u, err := at.DB.UserByEmail(at.Ctx, email)
+	u, err := at.DB.UserByEmail(at.Ctx, emailAddr)
 	if err != nil {
 		t.Fatal("Error while fetching the user from the DB. Error ", err.Error())
 	}
@@ -284,7 +284,7 @@ func testUserPUT(t *testing.T, at *test.AccountsTester) {
 	// Fetch the user from the DB because we want to be sure that their email
 	// is marked as unconfirmed which is not reflected in the JSON
 	// representation of the object.
-	u3, err := at.DB.UserByEmail(at.Ctx, email)
+	u3, err := at.DB.UserByEmail(at.Ctx, emailAddr)
 	if err != nil {
 		t.Fatal(err)
 	}
@@ -318,9 +318,8 @@ func testUserDELETE(t *testing.T, at *test.AccountsTester) {
 	if err != nil || r.StatusCode != http.StatusNoContent {
 		t.Fatalf("Expected %d success, got %d '%s'", http.StatusNoContent, r.StatusCode, err)
 	}
-	// Make sure the user doesn't exist anymore.
-	_, err = at.DB.
-  (at.Ctx, u.Email, false)
+	// Make sure the use doesn't exist anymore.
+	_, err = at.DB.UserByEmail(at.Ctx, u.Email)
 	if !errors.Contains(err, database.ErrUserNotFound) {
 		t.Fatalf("Expected error '%s', got '%s'.", database.ErrUserNotFound, err)
 	}


API.md
```diff
@@ -101,6 +101,17 @@ Conflict.
...
+### DELETE `/user`
+
+Deletes the user and all of their data.
+
+* Requires valid JWT: `true`
+* Returns:
+  - 204
+  - 401 (missing JWT)
+  - 404 (when there is no such user)
+  - 500 (on any other error)
```
MSevey, line 113: nit or follow up: we could probably move this to the top of the API docs right? I assume 401, 404 are the same for all endpoints where applicable, and then 500 is the default error. So at the top we have a default response codes section and then each endpoint simply states if they have a non default response.

api/handlers.go
```diff
@@ -266,6 +266,35 @@ func (api *API) userStatsGET(w http.ResponseWriter, req *http.Request, _ httprou
...
+func (api *API) userDELETE(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
+	ctx := req.Context()
+	sub, _, _, err := jwt.TokenFromContext(ctx)
+	if err != nil {
+		api.WriteError(w, err, http.StatusUnauthorized)
+		return
+	}
+	u, err := api.staticDB.UserBySub(ctx, sub, false)
+	if errors.Contains(err, database.ErrUserNotFound) {
+		api.WriteError(w, database.ErrUserNotFound, http.StatusNotFound)
```
MSevey, line 279: Follow Up: We should return the original error in case there is additional context added currently or in the future.

api/handlers.go
```diff
@@ -266,6 +266,35 @@ func (api *API) userStatsGET(w http.ResponseWriter, req *http.Request, _ httprou
...
+		return
+	}
+	if err != nil {
+		api.WriteError(w, err, http.StatusInternalServerError)
+		return
+	}
+	err = api.staticDB.UserDelete(ctx, u)
+	if errors.Contains(err, database.ErrUserNotFound) {
+		api.WriteError(w, database.ErrUserNotFound, http.StatusNotFound)
```
MSevey, line 288: same follow up comment

database/user.go
```diff
@@ -373,7 +373,26 @@ func (db *DB) UserDelete(ctx context.Context, u *User) error {
 	if u.ID.IsZero() {
 		return errors.AddContext(ErrUserNotFound, "user struct not fully initialised")
 	}
-	filter := bson.D{{"_id", u.ID}}
+	// Delete all data associated with this user.
+	filter := bson.D{{"user_id", u.ID}}
+	_, err := db.staticDownloads.DeleteMany(ctx, filter)
```
MSevey, line 378: Does `DeleteMany` return an error if there is no user data? If so, we should verify that the error is not due to no data.

test/api/handlers_test.go
```diff
@@ -302,6 +304,59 @@ func testUserPUT(t *testing.T, at *test.AccountsTester) {
 	}
 }
 
+// testUserDELETE tests the DELETE /user endpoint.
+func testUserDELETE(t *testing.T, at *test.AccountsTester) {
+	u, c, err := test.CreateUserAndLogin(at, t.Name())
+	if err != nil {
+		t.Fatal("Failed to create a user and log in:", err)
+	}
+
```
MSevey, line 313: Building from my other comment, we should add a test case here to delete the user before adding test data.

test/api/handlers_test.go
```diff
@@ -302,6 +304,70 @@ func testUserPUT(t *testing.T, at *test.AccountsTester) {
...
+	// Delete the user.
+	at.Cookie = c
+	defer func() { at.Cookie = nil }()
+	r, _, err = at.Delete("/user", nil)
+	if err != nil || r.StatusCode != http.StatusNoContent {
+		t.Fatalf("Expected %d success, got %d '%s'", http.StatusNoContent, r.StatusCode, err)
+	}
+	// Make sure that the data is gone.
+	stats, err := at.DB.UserStats(at.Ctx, *u.User)
+	if err != nil {
```
MSevey, line 357: Shouldn't this return error not found?

test/api/handlers_test.go
```diff
@@ -302,6 +304,70 @@ func testUserPUT(t *testing.T, at *test.AccountsTester) {
...
+func testUserDELETE(t *testing.T, at *test.AccountsTester) {
+	u, c, err := test.CreateUserAndLogin(at, t.Name())
+	if err != nil {
+		t.Fatal("Failed to create a user and log in:", err)
+	}
+	// Delete the user.
+	at.Cookie = c
+	defer func() { at.Cookie = nil }()
+	r, _, err := at.Delete("/user", nil)
+	if err != nil || r.StatusCode != http.StatusNoContent {
```
MSevey, line 317: I don't think we should be getting `StatusNoContent`. If a user exists, but doesn't have content, they should just be deleted regardless of if there is content of not. 

Should add verification that the user is in fact gone.

test/api/handlers_test.go
ro-tex: It's DB-level method, not a handler-level method. It's not its job to check that.

test/api/handlers_test.go
ro-tex: 1. Status No Content means that the operation was performed successfully and there is no data that's supposed to be returned. This is what our `WriteSuccess` method does both in accounts and in skyd.
2. Explicit verification that the user is gone would be redundant, as we try to create a user with the exact same data 3 lines later.

test/api/handlers_test.go
```diff
@@ -302,6 +305,80 @@ func testUserPUT(t *testing.T, at *test.AccountsTester) {
...
+		t.Fatalf("Expected %d success, got %d '%s'", http.StatusNoContent, r.StatusCode, err)
+	}
+	// Make sure the use doesn't exist anymore.
```
MSevey, line 321:
```suggestion
	// Make sure the user doesn't exist anymore.
```

test/api/handlers_test.go
```diff
@@ -302,6 +305,80 @@ func testUserPUT(t *testing.T, at *test.AccountsTester) {
...
+		t.Fatalf("Expected %d, got %d", http.StatusUnauthorized, r.StatusCode)
+	}
+	// Delete the user.
+	at.Cookie = c
+	defer func() { at.Cookie = nil }()
+	r, _, err = at.Delete("/user", nil)
+	if err != nil || r.StatusCode != http.StatusNoContent {
+		t.Fatalf("Expected %d success, got %d '%s'", http.StatusNoContent, r.StatusCode, err)
+	}
+	// Make sure the use doesn't exist anymore.
```
MSevey, line 361:
```suggestion
	// Make sure the user doesn't exist anymore.
```

MSevey (approved):

PR #89, opened 2021-10-22
Kratos migration script

ro-tex:
# PULL REQUEST

## Overview

This script will be run after all of our servers are running the new version of the portal and accounts which do not depend on Kratos for login and registrations.

**This PR will probably never be merged but it allows us to review the migration script.**

## Checklist
 - [ ] Changelog file created


peterjan (approved):
bit rough around the edges but the logic seems fine 👍

go.sum
```diff
@@ -6,18 +6,53 @@ cloud.google.com/go v0.44.1/go.mod h1:iSa0KzasP4Uvy3f1mN/7PiObzGgflwredwwASm/v6A
 cloud.google.com/go v0.44.2/go.mod h1:60680Gw3Yr4ikxnPRS/oxxkBccT6SA1yMk63TGekxKY=
 cloud.google.com/go v0.45.1/go.mod h1:RpBamKRgapWJb87xiFSdk4g1CME7QZg3uwTez+TSTjc=
 cloud.google.com/go v0.46.3/go.mod h1:a6bKKbmY7er1mI7TEI4lsAkts/mkhTSZK8w33B4RAg0=
+cloud.google.com/go v0.50.0/go.mod h1:r9sluTvynVuxRIOHXQEHMFffphuXHOMZMycpNR5e6To=
```
peterjan, line 9: Q: wondering where these come from, probably `go mod tiny` would get rid of these?

ChrisSchinnerl (approved):

go.sum
ro-tex: Nope, `go mod tidy` added them. It's interesting because when I run `go mod why` for those it says that the main module doesn't need them. But if I run `go mod graph` then I see that some `golang.org/x/tools` packages depend on them.

PR #90, opened 2021-10-22
Allow attaching MySky identities to accounts users

ro-tex:
# PULL REQUEST

## Overview

This PR allows for updating the user's pubKey via the `PUT /user` endpoint.

It depends on #83 and should be reviewed after that one is merged. The PR is currently pointing to that PR's branch, so the differences can be easily seen.

## Checklist
 - [x] Changelog file created


api/routes.go
```diff
@@ -34,6 +34,8 @@ func (api *API) buildHTTPRoutes() {
 	api.staticRouter.DELETE("/user", api.WithDBSession(api.validate(api.userDELETE)))
 	api.staticRouter.GET("/user/limits", api.noValidate(api.userLimitsGET))
 	api.staticRouter.GET("/user/stats", api.validate(api.userStatsGET))
+	api.staticRouter.GET("/user/addpubkey", api.validate(api.userAddPubKeyGET))
+	api.staticRouter.POST("/user/addpubkey", api.validate(api.userAddPubKeyPOST))
```
ro-tex, line 38: I don't really like the endpoint but `GET /user/pubkey/add` felt wrong. Also `GET /user/pubkey` would be very confusing if it doesn't actually give you the pubkey of the user and returns a challenge instead. So, I'm open to suggestions. Keeping the current name is also fine.

Makefile
```diff
@@ -75,7 +75,7 @@ start-mongo:
      -v $(shell pwd)/test/fixtures/mongo_keyfile:/data/mgkey \
 	mongo:4.4.1 mongod --port=17017 --replSet=skynet --keyFile=/data/mgkey 1>/dev/null 2>&1
 	# wait for mongo to start before we try to configure it
-	sleep 3
+	sleep 4
```
ChrisSchinnerl, line 78: Is there no better / more deterministic way to achieve this? e.g. by waiting in a loop for 60 seconds and trying to talk to mongo every second or so?

api/handlers.go
```diff
@@ -546,21 +548,125 @@ func (api *API) userPUT(w http.ResponseWriter, req *http.Request, _ httprouter.P
...
+		api.staticLogger.Warnf("Potential attempt to modify another user's pubKey. Sub of challenge requester '%s', sub of response submitter '%s'", uu.Sub, sub)
+		api.WriteError(w, errors.New("user's sub doesn't match update sub"), http.StatusBadRequest)
+		return
+	}
+	u, err := api.staticDB.UserBySub(ctx, sub, false)
+	if err != nil {
+		api.WriteError(w, err, http.StatusInternalServerError)
+		return
+	}
+	u.PubKeys = append(u.PubKeys, pk)
```
ChrisSchinnerl, line 656: Do we want to allow multiple pubkeys per user?

api/routes.go
ChrisSchinnerl: Why not `/user/pubkeys/register` ? Or just the singular?
Doesn't feel that wrong from a filesystem hierarchy standpoint.

test/database/challenge_test.go
```diff
@@ -112,3 +122,52 @@ func TestValidateChallengeResponse(t *testing.T) {
 		t.Fatal("Failed to create second challenge", err)
 	}
 }
+
+// TestUnconfirmedUserUpdate ensures the entire flow for unconfirmed user
+// updates works as expected.
+func TestUnconfirmedUserUpdate(t *testing.T) {
+	ctx := context.Background()
+	dbName := strings.ReplaceAll(t.Name(), "/", "_")
```
ChrisSchinnerl, line 130: This should be a helper. e.g. `dbNameForTest(t.Name())`

api/handlers.go
ro-tex: Yes, that was a specifically requested feature.

Makefile
ro-tex: I was putting this off because I knew it would take some time but... yeah, it had to be done. :)

m-cat (approved):

ChrisSchinnerl (approved):

PR #94, opened 2021-10-29
Raise the limits on upload file size.

ro-tex:
# PULL REQUEST

## Overview

This PR raises the limits on upload size based on the recent improvements made by Chris.

**Note:** Please note that this PR is targeted at `portal-latest` and not `main`, so it can be immediately deployed when we're ready to release these new limits.

## Checklist
 - [ ] Changelog file created


MSevey:
Can we set this against main? It isn't a big deal to cherry pick, and then it keeps us from having to merge portal-latest into main to get the update in the future.

ro-tex:
> Can we set this against main? It isn't a big deal to cherry pick, and then it keeps us from having to merge portal-latest into main to get the update in the future.

I specifically set it against `portal-latest`, so we can safely deploy it. We have a lot of new stuff in `main` that I would rather not deploy to prod, yet.

Cherry picking is fine in a pinch but it is an inherently dangerous strategy because we don't know what new interactions it might create. I prefer using the Gitflow hotfix approach where possible because it makes things so much easier to reason about in the long run.

MSevey (approved):

ChrisSchinnerl (approved):

PR #97, opened 2021-11-10
Change user's password

ro-tex:
# PULL REQUEST

## Overview

This PR makes it possible to change the user's password via the `PUT /user` endpoint.

## Checklist
 - [ ] Changelog file created


api/handlers.go
```diff
@@ -493,6 +494,15 @@ func (api *API) userPUT(w http.ResponseWriter, req *http.Request, _ httprouter.P
 		return
 	}
 
+	if payload.Password != "" {
```
MSevey, line 497: I wish I could comment outside the diff. 

But if the password isn't passed in, or if any individual fields isn't passed in, is this line then overwritten with empty fields https://github.com/SkynetLabs/skynet-accounts/pull/97/files#diff-855b4003387cd9e2b8e2c353c747fd1817eccf1d223b4066f0d7a1db0b4b4b4eR561

test/api/handlers_test.go
```diff
@@ -280,14 +280,53 @@ func testUserPUT(t *testing.T, at *test.AccountsTester) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	// Fetch the user from the DB again and make sure their password hash hasn't
+	// changed.
+	uSamePassHash, err := at.DB.UserByID(at.Ctx, u.ID)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if uSamePassHash.PasswordHash != u.PasswordHash {
```
MSevey, line 300: do we validate that the initial password is not empty?  So that `u.PasswordHash` isn't just a hash of an empty string, which in fact would be what was just set.

MSevey (changes requested):

api/handlers.go
```diff
@@ -476,7 +477,7 @@ func (api *API) userPUT(w http.ResponseWriter, req *http.Request, _ httprouter.P
 		api.WriteError(w, err, http.StatusBadRequest)
 		return
 	}
-	if payload.Email == "" && payload.StripeID == "" {
+	if payload.Email == "" && payload.StripeID == "" && payload.Password == "" {
```
ChrisSchinnerl, line 480: Can't you compare `payload == (userUpdateData{})` ? Seems like there are no reference types there which should make that comparison possible.

Now that I'm thinking about it doesn't it feel like we should return an error here to avoid dev mistakes? Why would you ever want this to be a no-op as a client?

api/handlers.go
ro-tex: I'm not sure which line you have in mind but nothing is overwritten unless explicitly passed.
What we do here is the following:
 - fetch the user from the DB
 - update the fields of the struct that we have validated values for
 - save the struct back to the DB with a `replace` statement

Any changes that we explicitly do here are preserved in the DB but no fields get reset to their default values.

api/handlers.go
ro-tex: Both good points. Yes, an empty update seems like a dev mistake.

test/api/handlers_test.go
ro-tex: Even if the old password was an empty string and the new one is also an empty string, their hashes would still be different because of the salt that's generated for each of them. So, if the password hash is the same we are sure that we didn't update the password, even if the old and new passwords are identical.

test/api/handlers_test.go
```diff
@@ -280,14 +280,53 @@ func testUserPUT(t *testing.T, at *test.AccountsTester) {
 		t.Fatalf("Expected the user to have StripeID %s, got %s", stripeID, u2.StripeID)
 	}
 	// Try to update the StripeID again. Expect this to fail.
-	r, _, err := at.UserPUT("", stripeID)
+	r, _, err := at.UserPUT("", "", stripeID)
 	if err == nil || !strings.Contains(err.Error(), "409 Conflict") || r.StatusCode != http.StatusConflict {
 		t.Fatalf("Expected to get error '409 Conflict' and status 409, got '%s' and %d", err, r.StatusCode)
 	}
 
+	// Update the user's password with an empty one. Expect this to succeed but
```
peterjan, line 288: does the comment match the code here? we expect a 400 right?

peterjan (approved):

ChrisSchinnerl (approved):

test/api/handlers_test.go
ro-tex: You're right. I'll F/U in https://github.com/SkynetLabs/skynet-accounts/pull/100

PR #100, opened 2021-11-15
Adjust API to portal expectations

ro-tex:
# PULL REQUEST

## Overview

This PR is a product of adjusting the API of `accounts` to the expectations and needs of the portal. Most of it was done in a rapid iteration flow while working with Karol. If the PR is hard to read because of the seemingly disconnected changes I can split it up.

List of changes:
 - Convert all non-GET handlers to pull their parameters from the request's body and not from the form.
 - Change GET /user/recover into POST /user/recover/request.
 - All links we send to users for the purposes of account management (email confirmation, account recovery, etc.) will now point to https://account.siasky.net and not to https://siasky.net.
 - Expose an `emailConfirmed` boolean field on the user object.

## Checklist
 - [ ] Changelog file created


test/tester.go
```diff
@@ -136,19 +135,28 @@ func (at *AccountsTester) Delete(endpoint string, params url.Values) (r *http.Re
 }
 
 // Post executes a POST request against the test service.
-func (at *AccountsTester) Post(endpoint string, params url.Values, postParams url.Values) (r *http.Response, body []byte, err error) {
+// TODO Remove the url.Values in favour of a simple map.
```
ro-tex, line 138: This is handled in #101

API.md
```diff
@@ -211,13 +211,13 @@ Requests another confirmation email sent to the account's email address.
  - 401
  - 500
 
-### GET `/user/recover`
+### POST `/user/recover/request`
```
peterjan, line 214: not sure why this is a `POST` and not just a `GET` request with email as a query string param?

api/handlers.go
```diff
@@ -39,16 +40,31 @@ type (
 		Storage           int64  `json:"storageLimit"`
 	}
 
-	// userUpdateData defines the fields of the User record that can be changed
+	// accountRecoveryDTO defines the payload we expect when a user is trying to
+	// change their password.
+	accountRecoveryDTO struct {
+		Token           string `json:"token"`
+		Password        string `json:"password"`
+		ConfirmPassword string `json:"confirmPassword"`
```
peterjan, line 48: don't we usually lower case these json prop names?

api/handlers.go
```diff
@@ -39,16 +40,31 @@ type (
...
-	userUpdateData struct {
+	userUpdateDTO struct {
 		Email    string `json:"email,omitempty"`
 		StripeID string `json:"stripeCustomerId,omitempty"`
 	}
 )
 
 // healthGET returns the status of the service
 func (api *API) healthGET(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
+	defer api.closeBody(req)
```
peterjan, line 67: This feels like it should be a middleware on the API that calls this at the end of each handler? Is that a possibility?

api/handlers.go
```diff
@@ -92,22 +110,30 @@ func (api *API) loginGET(w http.ResponseWriter, req *http.Request, _ httprouter.
...
+	err = json.Unmarshal(body, &payload)
+	if err == nil && payload.Email != "" && payload.Password != "" {
+		api.loginPOSTCredentials(w, req, payload.Email, payload.Password)
 		return
 	}
 
 	// Check for a challenge response in the request.
 	var chr database.ChallengeResponse
-	err := chr.LoadFromRequest(req)
+	err = chr.LoadFromRequest(bytes.NewBuffer(body))
```
peterjan, line 136: why is it called `LoadFromRequest` if you pass it a reader?

api/handlers.go
```diff
@@ -92,22 +110,30 @@ func (api *API) loginGET(w http.ResponseWriter, req *http.Request, _ httprouter.
 
 // loginPOST starts a user session by issuing a cookie
 func (api *API) loginPOST(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
+	defer api.closeBody(req)
+	// Get the body, we might need to use it several times.
```
peterjan, line 114: you're not using it several times right? I think this is redundant as you're more or less doing:
```
- body := ioutil.ReadAll(req.Body)
- buffer := bytes.NewBuffer(body)
- body := ioutil.ReadAll(buffer)
```

api/handlers.go
```diff
@@ -92,22 +110,30 @@ func (api *API) loginGET(w http.ResponseWriter, req *http.Request, _ httprouter.
...
+	body, err := ioutil.ReadAll(req.Body)
+	if err != nil {
+		api.WriteError(w, errors.AddContext(err, "empty request body"), http.StatusBadRequest)
```
peterjan, line 117: `nit` context might be misleading, it might not be empty right?

api/handlers.go
```diff
@@ -92,22 +110,30 @@ func (api *API) loginGET(w http.ResponseWriter, req *http.Request, _ httprouter.
...
-	if email != "" && pw != "" {
-		api.loginPOSTCredentials(w, req, email, pw)
+	var payload credentialsDTO
+	err = json.Unmarshal(body, &payload)
+	if err == nil && payload.Email != "" && payload.Password != "" {
+		api.loginPOSTCredentials(w, req, payload.Email, payload.Password)
 		return
 	}
 
 	// Check for a challenge response in the request.
```
peterjan, line 134: does this mean you're overloading a route with essentially 2 different types of flows depending on what's in the request body? Wouldn't it be better if it were a separate route?

api/handlers.go
```diff
@@ -260,18 +295,23 @@ func (api *API) registerPOST(w http.ResponseWriter, req *http.Request, _ httprou
...
+	var payload credentialsDTO
+	err = json.Unmarshal(body, &payload)
+	if err != nil {
+		api.WriteError(w, errors.AddContext(err, "failed to parse request body"), http.StatusBadRequest)
+		return
+	}
 	// Validate the email address.
-	email := req.PostFormValue("email")
-	e, err := mail.ParseAddress(email)
+	e, err := mail.ParseAddress(payload.Email)
```
peterjan, line 306: `nit` can we use `email`?

api/handlers.go
```diff
@@ -260,18 +295,23 @@ func (api *API) registerPOST(w http.ResponseWriter, req *http.Request, _ httprou
...
 	if err != nil {
 		api.WriteError(w, errors.New("invalid email provided"), http.StatusBadRequest)
 		return
 	}
 	// Strip any names from the email and leave just the address.
```
peterjan, line 311: should we remove or move this comment?

api/handlers.go
```diff
@@ -260,18 +295,23 @@ func (api *API) registerPOST(w http.ResponseWriter, req *http.Request, _ httprou
...
-	email = e.Address
 	// The password is optional.
-	pass := req.PostFormValue("password")
-	u, err := api.staticDB.UserCreatePK(ctx, email, pass, "", pk, database.TierFree)
+	payload.Email = e.Address
```
peterjan, line 313: should we update the comment here? it mentions the password, but adjust the email

api/handlers.go
```diff
@@ -321,6 +362,7 @@ func (api *API) userGET(w http.ResponseWriter, req *http.Request, _ httprouter.P
 
 // userLimitsGET returns the speed limits which apply to this user.
 func (api *API) userLimitsGET(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
+	defer api.closeBody(req)
```
peterjan, line 365: Do we have to call `closeBody` on all of these? it's easy to forget I think, would be cool if it could be applied automatically. But also, do we have a request body on all of these requests?

api/handlers.go
```diff
@@ -412,17 +456,28 @@ func (api *API) userDELETE(w http.ResponseWriter, req *http.Request, _ httproute
...
+	// Parse the request's body.
+	var payload credentialsDTO
+	b, err := ioutil.ReadAll(req.Body)
+	if err != nil {
+		api.WriteError(w, errors.AddContext(err, "failed to parse request body"), http.StatusBadRequest)
+		return
+	}
+	err = json.Unmarshal(b, &payload)
+	if err != nil {
+		api.WriteError(w, errors.AddContext(err, "failed to parse request body"), http.StatusBadRequest)
```
peterjan, line 469: update error message

api/handlers.go
```diff
@@ -39,16 +40,31 @@ type (
 		Storage           int64  `json:"storageLimit"`
 	}
 
-	// userUpdateData defines the fields of the User record that can be changed
+	// accountRecoveryDTO defines the payload we expect when a user is trying to
+	// change their password.
+	accountRecoveryDTO struct {
```
ChrisSchinnerl, line 45: What does `DTO` stand for? In skyd we would just call this `accountRecoveryPOST`.

api/handlers.go
```diff
@@ -794,26 +858,38 @@ func (api *API) userReconfirmPOST(w http.ResponseWriter, req *http.Request, _ ht
...
+		err = errors.AddContext(err, "failed to parse request body")
+		api.WriteError(w, err, http.StatusBadRequest)
+		return
+	}
+	if payload.Email == "" {
 		api.WriteError(w, errors.New("missing required parameter 'email'"), http.StatusBadRequest)
 		return
 	}
-	u, err := api.staticDB.UserByEmail(req.Context(), email)
+	u, err := api.staticDB.UserByEmail(req.Context(), payload.Email)
```
peterjan, line 887: here you're passing in the `Email` you received from the user directly without parsing the address. I think that's annoying UX right because on user creation or update you allow to pass value X (which internally gets transformed into X') but then when he wants to recover he would send X again, and you wouldn't find the user by email because you're not parsing it here. Right?

api/handlers.go
```diff
@@ -865,27 +941,33 @@ func (api *API) userRecoverGET(w http.ResponseWriter, req *http.Request, _ httpr
...
+	// Parse the request's body.
+	var payload accountRecoveryDTO
+	b, err := ioutil.ReadAll(req.Body)
+	if err != nil {
+		api.WriteError(w, errors.AddContext(err, "failed to parse request body"), http.StatusBadRequest)
+		return
+	}
+	err = json.Unmarshal(b, &payload)
+	if err != nil {
+		api.WriteError(w, errors.AddContext(err, "failed to parse request body"), http.StatusBadRequest)
```
peterjan, line 954: update error message

api/handlers.go
```diff
@@ -865,27 +941,33 @@ func (api *API) userRecoverGET(w http.ResponseWriter, req *http.Request, _ httpr
...
 		return
 	}
-	u, err := api.staticDB.UserByRecoveryToken(req.Context(), recoveryToken)
+	u, err := api.staticDB.UserByRecoveryToken(req.Context(), payload.Token)
 	if err != nil {
 		api.WriteError(w, errors.New("no such user"), http.StatusBadRequest)
 		return
 	}
-	passHash, err := hash.Generate(password)
+	passHash, err := hash.Generate(payload.Password)
```
peterjan, line 970: shouldn't we assert the `payload.Password` is not an empty string here? We do it on update as well right

api/handlers.go
ChrisSchinnerl, line 48: Is this the "repeat your password" field you usually have in forms? Hard to tell from the diff without docstrings but if it is that feels like something that should be handled by some JS on the website.

database/challenge.go
```diff
@@ -184,15 +198,25 @@ func (db *DB) DeleteUnconfirmedUserUpdate(ctx context.Context, chID primitive.Ob
 
 // LoadFromRequest loads a ChallengeResponse by extracting its string form from
 // http.Request.
-func (cr *ChallengeResponse) LoadFromRequest(req *http.Request) error {
-	resp, err := hex.DecodeString(req.PostFormValue("response"))
+func (cr *ChallengeResponse) LoadFromRequest(body io.Reader) error {
```
peterjan, line 201: I think if you pass in the body we should also update the name of the method.

database/database.go
```diff
@@ -61,7 +61,7 @@ var (
 	mongoWriteConcern = "majority"
 	// mongoWriteConcernTimeout specifies a time limit, in milliseconds, for
 	// the write concern to be satisfied.
-	mongoWriteConcernTimeout = "1000"
+	mongoWriteConcernTimeout = "5000"
```
peterjan, line 64: you increased this to 5s, is it worth documenting why it's 5s? probably you ran into timeouts at 1?

peterjan (changes requested):

database/user.go
```diff
@@ -112,6 +112,7 @@ type (
 		// its ID.Hex() form.
 		ID                               primitive.ObjectID `bson:"_id,omitempty" json:"-"`
 		Email                            string             `bson:"email" json:"email"`
+		EmailConfirmed                   bool               `bson:"email_confirmed" json:"emailConfirmed"`
```
ChrisSchinnerl, line 115: If `EmailConfirmed` is only true if the corresponding token is empty and vice versa, do we need this field? Can't we just return `EmailConfirmationToken == ""` ?

API.md
ro-tex: It's a POST because it creates a record in the DB rather than reading one. Therefore according to the REST approach it needs to be a POST and not a GET.

api/handlers.go
ro-tex: No. We use snake_case in MongoDB field names and camelCase in JSON fields.

api/handlers.go
ro-tex: It is possible, yes. I would strongly advise against it, though. The reason for that is that it's very much non-standard and developers won't expect it, meaning that they will go an close their request bodies and then they'll be surprised by the double-close panics. And all such a middleware would achieve is save us one line per handler. I would rather go for simple, explicit and obvious code.

I don't know how other projects deal with this problem. In `skyd` I believe we just leak those all over the place. It might be a good idea to examine and fix that, if it's really the case.

api/handlers.go
ro-tex: Because I forgot to rename it. :) It used to read from the request.
I'll fix it, thanks.

api/handlers.go
ro-tex: That's true. Fixing.

api/handlers.go
ro-tex: That's true. Fixing.

api/handlers.go
ro-tex: Yes, this is what's happening.

Currently we have three ways of logging in:
* challenge-response
* username and password
* JWT token (either in a cookie or in an `Authorization` header)

We can make a separate endpoint for each of the three and then make the webportal call the right one. But it seems we'll be removing the username and password before long, so it might no make much sense to do that.

api/handlers.go
ro-tex: `e` is a temporary variable that lives for 3 lines. I can change it to `emailAddr` but I would rather not use `email` as that overlaps with a package name (which is the original reason for me to make this change, as seen in this diff).

api/handlers.go
ro-tex: Moved.

api/handlers.go
ro-tex: Moved the comment down a bit and expanded to fully explain why it's there - we mention that the password is optional because we're not checking it in any way.

api/handlers.go
ro-tex: Yes, we need to always close the body of a request that has a body. Here's a discussion about it:  https://stackoverflow.com/questions/33238518/what-could-happen-if-i-dont-close-response-body

If we don't close it's not a terrible thing but it's a resource leak and, more importantly, it will prevent the HTTP connection from being reused for the next API call the user wants to make. That means another handshake, which is slow. Closing this everywhere does not cost us anything.

api/handlers.go
ro-tex: This one is fine. But the one at 464 should say "failed to read request body". Fixed.

api/handlers.go
ro-tex: Oh, nice catch! Yes, you are absolutely right. I'll add a helper method for cleaning the email and I'll use it everywhere.

api/handlers.go
ro-tex: Same as above, changed the message of the preceding error.

api/handlers.go
ro-tex: We do that on line 950 where we have
```go
	if payload.Password == "" || payload.ConfirmPassword == "" || payload.Token == "" {
		api.WriteError(w, errors.New("missing required parameter"), http.StatusBadRequest)
		return
	}
```

database/challenge.go
ro-tex: Name updated.

api/handlers.go
ro-tex: It's a standard abbreviation: https://en.wikipedia.org/wiki/Data_transfer_object

api/handlers.go
ro-tex: Yes, this is something that I though as well, and Karol mentioned, too. The main reason it's here is to make sure we don't forget that this part of the process needs to be handled. Karol and I agreed that it should be handled by the frontend but left it like this with "we'll fix this later".

I'd rather F/U on this than address it here and break the portal code we have. It's on my radar.

changelog/items/key-updates/json_payloads.md
```diff
@@ -0,0 +1 @@
+- BREAKING: All non-GET handlers now read their parameters from the request's body JSON instead of the form.
```
MSevey, line 1: Calling out as a reminder we should bump the major version then

database/challenge_test.go
```diff
@@ -14,41 +16,80 @@ import (
 	"golang.org/x/crypto/ed25519"
 )
 
-// TestChallengeResponse_LoadFromRequest tests the LoadFromRequest method of
+// TestChallengeResponse_LoadFromRequest tests the LoadFromReader method of
 // ChallengeResponse.
 func TestChallengeResponse_LoadFromRequest(t *testing.T) {
```
MSevey, line 21: nit: same update, comment too

changelog/items/key-updates/json_payloads.md
ro-tex: Can do. But we can also not do it - the major version is zero, which implies no compatibility guarantees. 

But it might be time this service got its recognition with an official version. :)

email/sender.go
```diff
@@ -32,6 +32,9 @@ var (
 	// comes from the PORTAL_DOMAIN environment variable, preceded by the
 	// appropriate schema.
 	PortalAddress = "https://siasky.net"
+	// PortalAddressAccounts defines the URI where we can access the accounts
+	// sub-site.
+	PortalAddressAccounts = "https://account.siasky.net"
```
MSevey, line 37: Is there a reason this and `PortalAddress` aren't just loaded from the `.env` file?

MSevey: They probably should, otherwise community portals are going to be sending emails with `siasky.net` links in them instead of their own links.

api/handlers.go
```diff
@@ -260,18 +294,21 @@ func (api *API) registerPOST(w http.ResponseWriter, req *http.Request, _ httprou
...
-	e, err := mail.ParseAddress(email)
+	// Parse the request's body.
+	var payload credentialsDTO
+	err = json.Unmarshal(body, &payload)
+	if err != nil {
+		api.WriteError(w, errors.AddContext(err, "failed to parse request body"), http.StatusBadRequest)
+		return
+	}
+	// Validate and normalize the email address.
+	payload.Email, err = lib.NormalizeEmail(payload.Email)
```
MSevey, line 305: nit: probably worth a `UnmarshalPayload` method that combines these two actions so that callers don't forget one.

api/handlers.go
```diff
@@ -469,7 +520,7 @@ func (api *API) userPUT(w http.ResponseWriter, req *http.Request, _ httprouter.P
 		api.WriteError(w, err, http.StatusBadRequest)
 		return
 	}
-	var payload userUpdateData
+	var payload userUpdateDTO
 	err = json.Unmarshal(bodyBytes, &payload)
```
MSevey, line 524: We should make the `UnmarshalPayload` method so we can use it here since we aren't normalizing it immediately which probably should be updated.

api/handlers.go
```diff
@@ -517,14 +568,12 @@ func (api *API) userPUT(w http.ResponseWriter, req *http.Request, _ httprouter.P
 
 	var changedEmail bool
 	if payload.Email != "" {
-		// Validate the new email.
-		a, err := mail.ParseAddress(payload.Email)
+		// Validate  and normalize the new email.
+		payload.Email, err = lib.NormalizeEmail(payload.Email)
```
MSevey, line 572: Should be part of the unmarshal code.

api/handlers.go
```diff
@@ -794,26 +851,38 @@ func (api *API) userReconfirmPOST(w http.ResponseWriter, req *http.Request, _ ht
...
+		err = errors.AddContext(err, "failed to read request body")
 		api.WriteError(w, err, http.StatusBadRequest)
 		return
 	}
-	email := req.Form.Get("email")
-	if email == "" {
+	var payload struct {
+		Email string `json:"email"`
+	}
+	err = json.Unmarshal(bodyBytes, &payload)
```
MSevey, line 870: Same, missed calling normalize for the email.

api/handlers.go
```diff
@@ -865,27 +934,33 @@ func (api *API) userRecoverGET(w http.ResponseWriter, req *http.Request, _ httpr
...
-		api.WriteError(w, err, http.StatusBadRequest)
+	defer api.closeBody(req)
+	// Parse the request's body.
+	var payload accountRecoveryDTO
+	b, err := ioutil.ReadAll(req.Body)
+	if err != nil {
+		api.WriteError(w, errors.AddContext(err, "failed to read request body"), http.StatusBadRequest)
+		return
+	}
+	err = json.Unmarshal(b, &payload)
```
MSevey, line 945: same

MSevey (dismissed):
Left a few comments which are fine as Follow ups.

database/user.go
ro-tex: Thanks for the reminder. This is another thing I had to add very quickly and I knew I'd have to follow up on.

I will need to keep the field on the struct but I won't store it in the DB anymore. It will be filled when we're marshalling the user struct into JSON. But we can't avoid having the field unless we want to manually augment the payload each time marshal this into JSON.

If you have other ideas, pleas share.

database/challenge_test.go
ro-tex: Thanks. GoLand's rename feature could be better.

email/sender.go
ro-tex: They are. These are just initial values. It seems that they are confusing, though. I'll remove them.

api/handlers.go
ro-tex: Add a new type `EmailField` with a custom unmarshaller. I just need to write its tests and fix all the other tests that broke in the process.

email/sender.go
ro-tex: Actually, I'm going to keep them. The reason for that is that they won't be initialised during testing unless I do it manually which is not great.

We don't need to worry about custom portals - if the `PORTAL_DOMAIN` env var which defines those is not set the service will panic on start with an appropriate error message.

changelog/items/key-updates/json_payloads.md
MSevey: I think it makes sense for the official version as it will align with us pushing community portals

email/sender.go
MSevey: If these are only used in testing I would move them to a test file to make it more clear that these aren't duplicate production constants. 

Or use a build variable :-) 

Fine as a follow up.

MSevey (dismissed):

api/handlers.go
ChrisSchinnerl: Let's pls stick to our skyd code conventions. DTO is very generic and really doesn't give any context about the usage of the object. Suffixes like `GET` or `Request`/`Response` give a bit more context even if they are just DTOs. The latter is also a bit more friendly than an abbrevation.

ChrisSchinnerl (changes requested):

lib/util.go
```diff
@@ -16,3 +18,16 @@ func GenerateUUID() (string, error) {
 	}
 	return hex.EncodeToString(uid[:]), nil
 }
+
+// NormalizeEmail returns the email address and strips all other text from the
+// input, e.g. "Barry Gibbs <bg@example.com>" becomes "bg@example.com".
+func NormalizeEmail(emailAddr string) (string, error) {
```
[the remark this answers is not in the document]
ro-tex: I would expect to get all the fields from a function like that.

types/email.go
```diff
@@ -0,0 +1,31 @@
+package types
+
+import (
+	"encoding/json"
+
+	"github.com/SkynetLabs/skynet-accounts/lib"
```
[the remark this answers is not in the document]
ro-tex: I moved the email method to `types/email.go`.

types/email.go
```diff
@@ -0,0 +1,31 @@
...
+)
+
+// EmailField is a helper type that has custom unmarshalling code which
+// makes sure the email has the correct structure.
+type EmailField string
```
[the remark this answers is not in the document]
ro-tex: Sure.

test/tester.go
```diff
@@ -161,6 +175,8 @@ func (at *AccountsTester) Post(endpoint string, params url.Values, postParams ur
 }
 
 // Put executes a PUT request against the test service.
+//
+// NOTE: The Body of the returned response is already read and closed.
```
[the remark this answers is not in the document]
ro-tex: I don't know what I'll need, so I'd rather have the entire response. Usually it's the status and the headers.

lib/util.go
```diff
@@ -16,3 +18,16 @@ func GenerateUUID() (string, error) {
 	}
 	return hex.EncodeToString(uid[:]), nil
 }
+
+// NormalizeEmail returns the email address and strips all other text from the
+// input, e.g. "Barry Gibbs <bg@example.com>" becomes "bg@example.com".
```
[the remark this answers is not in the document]
ro-tex: The whole thing started form the idea of validating emails. There are a bunch of unreliable regexs out there but there's also this library method. So, I decided to us the library instead.

The other part is that the challenge-response registration will be driven by someone's code. It might be our code in mysky but it might also be something else. So, that code might use the full email formatting shown in the comment. Then the validation would go through (i.e. it won't error out) but the email we pass to the DB and to the mail sender will be the full string. So, the user won't be able to login as expected. So I decided to normalize the email. And Matt requested that it happens on deserialization, so we can't forget it.

What we can do instead is validate the email by parsing it with the same library function and if the input doesn't match the address output we can throw an error. This should take care of the issue in a simpler way. But we'll still have the "how do we make sure we never forget to do this?" issue.

api/handlers.go
ro-tex: I really don't like this because it replaces an industry-standard term with a bunch of in-house ones which are also not very informative. Calling something DTO means that it's a wrapper used only for transportation and the object should not be used for anything else. It also means that the object doesn't have any functionality attached to, aside from formatting helpers.

I can see the benefit of sticking to conventions, though. So, sure, I'll rename all DTOs to <verb>/request/response types.

peterjan: > I can see the benefit of sticking to conventions, though.

This is really the only reason this should be renamed. DTO was the correct term here imo. DTOs are pretty widely adopted, if you've worked with ORMs or frameworks like symfony or react or whatever, DTO as a concept is baked into all of those.

api/handlers.go
peterjan: ok

api/handlers.go
peterjan: I don't mind if you don't update it, but I disagree it adds complexity. On the contrary, it just ensures it can not be forgotten, if a developer runs into a double close he'll simply notice and realize it's being closed for him by the middleware. But I don't feel strongly so you can resolve this comment.

api/handlers.go
peterjan: It's because it's not necessarily a minor performance hit. I can send a very large body and you would be making it easy for me because you would be allocating twice the amount of memory. You mention you need an `io.Reader` in another, are you referring to `LoadFromReader`? Because there the first thing we do is `b, err := ioutil.ReadAll(body)`, perhaps we need `LoadFromBytes` to go alongside it? Or a json decoder is probably best.

api/handlers.go
peterjan: The stacko you linked is something else in my opinion. There the client is closing it, it does not suggest to close it in the HTTP handler, which I think is actually advised against for example here https://stackoverflow.com/questions/42525499/where-to-put-defer-req-body-close

database/database.go
peterjan: Well, we should definitely increase it to something more than 1s. I think 5s is probably a good value. It is indeed the value that you see appear in most examples, that's because it's the value used by the mongo documentation as well. 

I think this is a pretty dangerous setting for us due to our setup. Our replica set currently has 20 nodes, and the primary might be at the other side of the world right. I think this timeout is thrown when the primary doesn't acknowledge the write within the given timeout value. For us this means that we have to reach the primary, and a good chunk (majority) of the secondaries as well I believe, before it'll ack.

What I'm trying to say is that 5s might not give us enough leeway even. We should consider dropping it altogether or perhaps bump it to something like 30s? Which is still a reasonable request timeout value, and gives us even more leeway. It depends on how we treat the timeout, because the timeout doesn't mean the data is not persisted at the primary.

> A write operation that times out waiting for the specified write concern only indicates that the required number of replica set members did not acknowledge the write operation within the wtimeout time period. It does not necessarily indicate that the primary failed to apply the write. The data may exist on a subset of replica set nodes at the time of the write concern error, and can continue replicating until all nodes in the cluster have that data. Applications should take into account the potential availability of written data regardless of the state of write concern acknowledgment.

https://docs.mongodb.com/manual/core/replica-set-write-concern/

api/handlers.go
ro-tex: Hey, you are right! I need to go back to that stackoverflow article and leave a comment.

The comment you found was key, I can't believe I haven't seen it. `The Server will close the request body. The ServeHTTP Handler does not need to.`

Here's some code that proves that the body is closed by the server: https://go.dev/play/p/pBuVoF7SyQN
(you will see that it prints ` >>> CLOSED! <<<` on its own).

I'll remove these `Close()` calls.

database/challenge.go
```diff
@@ -182,17 +196,28 @@ func (db *DB) DeleteUnconfirmedUserUpdate(ctx context.Context, chID primitive.Ob
...
-// http.Request.
-func (cr *ChallengeResponse) LoadFromRequest(req *http.Request) error {
-	resp, err := hex.DecodeString(req.PostFormValue("response"))
+// LoadFromReader loads a ChallengeResponse from the given io.Reader.
+//
+// Typically, this reader will be a request.Body.
+func (cr *ChallengeResponse) LoadFromReader(body io.Reader) error {
+	// Parse the request's body.
+	var payload challengeResponseDTO
+	b, err := ioutil.ReadAll(body)
```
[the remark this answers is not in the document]
ro-tex: The decoder won't save us from that, I'm afraid (I checked). But you are right that this is a problem. I'll use either https://github.com/nanmu42/limitio or my own limited reader (unless we already have one?) in order to protect against that.

email/sender.go
ro-tex: Build variable sounds great. Will do.

database/database.go
ro-tex: I agree with that. I'm not sure I want to go for no limit at all because that will potentially block forever.

The 30s sounds very long but still reasonable, especially having in mind that `skyd` has similar timeouts.

I wouldn't really worry about data corruption because all of our operations are wrapped in transactions. A write concern error which aborts a given flow will cause the transaction to be aborted and rolled back.

api/handlers.go
ro-tex: The `LoadFromBytes` is a good idea, thanks.

A JSON decoder can be helpful if we have a very long stream of JSON objects but it won't protect us from the malicious body. What will protect us is a limit reader. I'll write one and we can discuss whether it works well for the purpose.

api/handlers.go
ro-tex: You are right. The server doesn't need to close the body. Removed.

api/handlers.go
ro-tex: All DTOs renamed.

api/handlers.go
ro-tex: * `LoadFromBytes` implemented
* `io.LimitedReader` user for reading request bodies

api/handlers.go
ro-tex: Is overloading this route acceptable at this point?

changelog/items/key-updates/json_payloads.md
ro-tex: Should this be a follow up for when we're ready to release?

database/challenge.go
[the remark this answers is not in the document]
ro-tex: Added json decoder and limitreader.

types/email.go
```diff
@@ -0,0 +1,31 @@
...
+	// Missing or empty values will be interpreted as an empty string.
+	if b == nil || len(b) == 0 {
+		*e = ""
+		return nil
+	}
+	var emailStr string
+	if err := json.Unmarshal(b, &emailStr); err != nil {
+		return err
+	}
+	normalizedEmail, err := lib.NormalizeEmail(emailStr)
```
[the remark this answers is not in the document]
ro-tex: Resolved by a more standard approach to validating emails.

lib/util.go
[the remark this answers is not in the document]
ro-tex: Resolved by a more standard approach to validating emails.

types/email.go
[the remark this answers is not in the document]
ro-tex: Type removed altogether.

lib/util.go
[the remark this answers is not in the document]
ro-tex: Removed.

test/tester.go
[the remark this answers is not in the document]
ro-tex: I'm not sure what I think about this one.

On one hand, yes, we don't need the entire response because we already have the body and there isn't anything else that's useful in the response besides status code and headers. 

On the other hand, passing a single pointer to the response is a non-issue, as long as the developer realises that they already have the body conveniently returned to them and doesn't try to read it again. Changing the return signature to `(code int, head http.Header, body []byte, err error)` doesn't seem to be improving it much.

To me it looks like busywork (because we'll need to fix literally all tests) with no real impact. But if you find this valuable I'll make the change.

changelog/items/key-updates/json_payloads.md
ro-tex: Version 1.0!

api/handlers.go
ro-tex: https://github.com/SkynetLabs/skynet-accounts/issues/111

test/tester.go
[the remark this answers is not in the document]
ro-tex: https://github.com/SkynetLabs/skynet-accounts/issues/111

database/challenge.go
```diff
@@ -204,6 +232,18 @@ func (cr *ChallengeResponse) LoadFromRequest(req *http.Request) error {
 	return nil
 }
 
+// LoadFromReader loads a ChallengeResponse from the given io.Reader.
+//
+// Typically, this reader will be a request.Body.
+func (cr *ChallengeResponse) LoadFromReader(r io.Reader) error {
+	// Parse the request's body.
+	b, err := ioutil.ReadAll(r)
```
ChrisSchinnerl, line 240: f/u It's not obvious that `ReadAll` is limited by a `LimitReader`. So I'd apply the limit reader in here.

api/handlers.go
```diff
@@ -248,31 +279,39 @@ func (api *API) registerGET(w http.ResponseWriter, req *http.Request, _ httprout
...
-		api.WriteError(w, errors.AddContext(err, "failed to validate challenge response"), http.StatusBadRequest)
+		api.WriteError(w, errors.AddContext(err, "failed to parse request body"), http.StatusBadRequest)
 		return
 	}
-	// Validate the email address.
-	email := req.PostFormValue("email")
-	e, err := mail.ParseAddress(email)
-	if err != nil {
+	parsed, err := mail.ParseAddress(payload.Email)
+	if err != nil || payload.Email != parsed.Address {
```
peterjan, line 303: This is quite odd? I don't think I follow. So you parse the address from the `payload.Email` field and then later on you error out when `payload.Email` is not equal to `parsed.Address`. Doesn't it make sense to alter `ParseAddress` and return an `error` in that case?

api/handlers.go
peterjan: I think this is still an issue?

api/handlers.go
ChrisSchinnerl: f/u I would scan the codebase for `ioutil.ReadAll` and then just get rid of all of them. It's usually not something you want to use except for testing since you can always just pass the limited reader to a `json.Decoder`. That's also more efficient since it will decode the body as it comes in without reading the whole thing first and then decoding it.

ChrisSchinnerl (approved):

database/challenge.go
```diff
@@ -119,8 +133,15 @@ func (db *DB) ValidateChallengeResponse(ctx context.Context, chr ChallengeRespon
 	// Now that we know the challenge type, we can get the recipient as well.
 	recipientOffset := ChallengeSize + len([]byte(cType))
 	recipient := string(resp[recipientOffset:])
-	if recipient != PortalName {
-		return nil, primitive.ObjectID{}, errors.New("invalid recipient " + recipient)
+	// Make sure the recipient has a schema. It usually won't. We check for
+	// `http` only in order to cover both `http://` and `https://`
+	if !strings.HasPrefix(recipient, "http") {
```
peterjan, line 138: `nit` do we care that this doesn't work when the recipient is `httpsomethingsomething`?

peterjan (commented):
almost ready I think, I unresolved one comment (https://github.com/SkynetLabs/skynet-accounts/pull/100/files#r754270960) which I think still needs to be addressed, if not I'll approve it

api/handlers.go
ro-tex: That was my initial approach but it comes with its own problems. This approach guarantees that unless the email is valid and simple (e.g. "joe@example.com" and not "Joe Smith <joe@example.org>") the process will stop with an error.

api/handlers.go
```diff
@@ -248,31 +279,39 @@ func (api *API) registerGET(w http.ResponseWriter, req *http.Request, _ httprout
 
 // registerPOST registers a new user based on a challenge-response.
 func (api *API) registerPOST(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
+	// Get the body, we might need to use it several times.
+	body, err := ioutil.ReadAll(io.LimitReader(req.Body, 4*skynet.KiB))
```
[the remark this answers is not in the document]
ro-tex: https://github.com/SkynetLabs/skynet-accounts/issues/111

api/handlers.go
ro-tex: No, this is no longer an issue because if they try to use any email which is not a valid and simple (e.g. joe@example.com) the registration will fail. So there is no need to validate here.

database/challenge.go
```diff
@@ -37,6 +40,10 @@ const (
 )
 
 var (
+	// ErrInvalidChallengeResponse is returned when the received challenge
+	// response object is not valid, i.e. it's either missing one of its
+	// required fields or those do not follow the expected format.
+	ErrInvalidChallengeResponse = errors.New("invalid response")
```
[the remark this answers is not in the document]
ro-tex: #111

api/handlers.go
peterjan: oh ok so we updated registration! ok great thanks

database/challenge.go
ro-tex: It works now.

api/handlers.go
peterjan: Well it guarantees this as long as you remember to consider `payload.Email != parsed.Address` as an error. I'll approve but I don't understand why this condition can't be inside of `ParseAddress`.

peterjan (approved):

api/handlers.go
ro-tex: ParseAddress is a stdlib function. I'll add this to the F/U list.

ChrisSchinnerl (dismissed):

ChrisSchinnerl (approved):

peterjan (approved):

PR #101, opened 2021-11-15
Better naming of test request parameters

ro-tex:
# PULL REQUEST

## Overview

There are no functional changes in this PR, only a transition between POST params being passed as URL parameters (similar to what we do in `skyd`) and passing them as JSON in the body of POST and PUT requests (which is the usual way of doing it).

**This PR depends on #100 being merged, so do not review before that.**

## Checklist
 - [ ] Changelog file created


test/api/api_test.go
```diff
@@ -190,11 +191,10 @@ func TestUserTierCache(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	body := map[string]string{
-		"email":    email,
-		"password": password,
-	}
-	r, _, err := at.Post("/login", nil, body)
+	bodyParams := url.Values{}
+	bodyParams.Add("email", email)
```
ChrisSchinnerl, line 195: Doesn't matter functionality-wise but using `Add` makes me think there is a special reason for using `Add`. Because it's usually used when you need to set multiple values for the same param.

Otherwise `Set` would suffice.

test/api/api_test.go
```diff
@@ -191,10 +191,10 @@ func TestUserTierCache(t *testing.T) {
...
 		t.Fatal(err)
 	}
-	params := url.Values{}
-	params.Add("email", email)
-	params.Add("password", password)
-	r, _, err := at.Post("/login", nil, params)
+	bodyParams := url.Values{}
+	bodyParams.Add("email", email)
+	bodyParams.Add("password", password)
+	r, _, err := at.Post("/login", nil, bodyParams)
```
ChrisSchinnerl, line 197: This doesn't actually change the body as suggested by the PR title right? It's still a json object. Only the argumnt to `Post` changed?

ro-tex: Not really, no. The PR was created when we changed the handlers to use JSONs and then it was side-tracked by other priorities.

test/api/api_test.go
ro-tex: Oh, good point. I just read the implementation and `Set` is definitely what I want.

test/api/api_test.go
ChrisSchinnerl: Ah I see. In that case wouldn't it make more sense to make the argument for the body a `bodyParams interface{}` ? Then instead of `url.Values` we could use actual types.

e.g. here
```go
type LoginRequest struct {
  Email        string `json:"email"`
  Password     string `json:"password"`
}
```
and then
```go
bodyParams := LoginRequest{
  Email: email,
  Password: password,
}
r, _, err := at.Post("/login", nil, bodyParams)
```

peterjan (approved):
approved - but I like Chris 's suggestion with the post body

test/api/api_test.go
ChrisSchinnerl: We can f/u with it.

You shouldn't need type validation after that because all you do is pass it on to `json.Marshal` anyway.
The additional type safety will come from something like the following:

```go
func (c *Client) Login(email, password string) error {
  return c.post("/login", LoginRequest{
    Email: email,
    Password: password,
  }
}
```

Since `post` is no longer exposed, a user can only call the endpoint through the exposed `Login` method which handles the type nicely.

ChrisSchinnerl (approved):

test/api/api_test.go
ro-tex: Added to https://github.com/SkynetLabs/skynet-accounts/issues/138

PR #102, opened 2021-11-24
Option to disable new account registrations

ro-tex:
# PULL REQUEST

### Problem

While migrating from the current accounts system (ORY stack) to the new one (mysky), we will temporarily be running a heterogeneous stack. While we have tools in place that will allow the new stack to work with registrations happening on the old one, we don't have the tools for making the reverse possible. The reason for this is that creating a user registration in the ORY stack is not as straightforward.

### Solution
We will temporarily disable new account registrations on the new stack while running a heterogeneous environment. Once the entire cluster is migrated we will flip the switch and enable registrations again. While this is happening, registrations will still be allowed on the old stack and continually migrated to the new one.

#### Password changes and account recoveries
Since the problem addressed by this PR is the desynchronisation between CockroachDB and MongoDB, we should also address that problem when users change their passwords or recover their accounts. Both those actions are now disabled while new registrations are disabled.

### Implementation
This PR adds a temporary option to disable new registrations that can be (de)activated without a new deployment or a restart. Since the measure is temporary, I've avoided going overboard with optimisations - we're going to hit the DB on each requst with out any caching. Since this is only valid for the registration calls (which are rare) the impact will be minimal to none.

The underlying configuration mechanism is meant to stay, though. It is useful to have a globally-shared set of configuration options that don't rely on restarts or deployments.

**NOTE**: This PR is targeting the `ivo/json_bodies` branch in order to be easily reviewed. It is meant to be merged into `main` and I expect that to happen after `ivo/json_bodies` is merged.

## Checklist
 - [ ] Changelog file created


database/configuration.go
```diff
@@ -0,0 +1,68 @@
...
+	err := sr.Decode(option)
+	if err != nil {
+		return "", err
+	}
+	return option.Value, nil
+}
+
+// WriteConfigValue writes the value for the given key to the dbConfiguration
+// table.
+func (db *DB) WriteConfigValue(ctx context.Context, key, value string) error {
```
peterjan, line 56: Do we ever need this? This is not safe by the way as it's not protected against concurrent writes.

api/handlers.go
```diff
@@ -534,6 +565,18 @@ func (api *API) userPUT(w http.ResponseWriter, req *http.Request, _ httprouter.P
 	}
 
 	if payload.Password != "" {
+		// Check if the registrations are open. If they are not then changing
+		// passwords is also not allowed.
+		val, err := api.staticDB.ReadConfigValue(req.Context(), database.ConfValRegistrationsDisabled)
+		if err != nil && !errors.Contains(err, mongo.ErrNoDocuments) {
```
peterjan, line 571: Maybe it makes sense to return `nil` or default to `false`  in case the error equals `mongo.ErrNoDocuments`, and do this inside `ReadConfigValue`. Not sure if I like it myself yet but it could clean it up a bunch. I think we definitely want to DRY this up a bit it's a bunch of code to read through for such a simple thing.

peterjan (dismissed):

api/handlers.go
```diff
@@ -257,8 +258,18 @@ func (api *API) logoutPOST(w http.ResponseWriter, req *http.Request, _ httproute
 
 // registerGET generates a registration challenge for the caller.
 func (api *API) registerGET(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
+	// Check if the registrations are open.
```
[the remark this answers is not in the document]
ro-tex: Absolutely. This is something we are about to remove right after the migration, though, so it might not be worth it.

database/configuration.go
ro-tex: Hmm... We don't need it right now but we might need it in the future. As I was writing this, I intended it to be a flexible system for DB-level configuration options. So, this is something I want to follow up on. I'll add an issue for it.

ro-tex: F/U: https://github.com/SkynetLabs/skynet-accounts/issues/110

.github/workflows/github-actions-demo.yml
```diff
@@ -8,7 +8,7 @@ jobs:
         uses: actions/checkout@v2
       - uses: actions/setup-go@v2
         with:
-          go-version: '1.16.7'
+          go-version: '1.17.2'
```
MSevey, line 11: why are we deviating from the version used elsewhere?

MSevey (approved):

.github/workflows/github-actions-demo.yml
ro-tex: The auto-lint kept failing with an odd error: https://github.com/SkynetLabs/skynet-accounts/runs/4458893545?check_suite_focus=true

I still don't know why that's happening, only that switching to a newer go version fixes it.

ChrisSchinnerl (approved):

peterjan (approved):
LGTM

PR #105, opened 2021-12-01
Lower Tier Limits

peterjan:
# PULL REQUEST

## Overview

This PR temporarily lowers the `anon` tier download limits to reduce over all download pressure on our production nodes. I applied a `/4` on all download related limits.

## Checklist

N/A

## Issues Closed

N/A

### DIFF
--- a/database/user.go
+++ b/database/user.go
@@ -54,18 +54,22 @@ var (
 	// RegistryDelay delay is in ms.
 	UserLimits = map[int]TierLimits{
 		TierAnonymous: {
-			TierName:          "anonymous",
-			UploadBandwidth:   5 * mbpsToBytesPerSecond,
-			DownloadBandwidth: 20 * mbpsToBytesPerSecond,
+			TierName:        "anonymous",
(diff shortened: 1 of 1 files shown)


database/user.go
```diff
@@ -54,9 +54,11 @@ var (
 	// RegistryDelay delay is in ms.
 	UserLimits = map[int]TierLimits{
 		TierAnonymous: {
-			TierName:          "anonymous",
-			UploadBandwidth:   5 * mbpsToBytesPerSecond,
-			DownloadBandwidth: 20 * mbpsToBytesPerSecond,
+			TierName:        "anonymous",
+			UploadBandwidth: 5 * mbpsToBytesPerSecond,
+			// TODO: temporarily lowered the download bandwidth on the anon tier
+			// from 20mbps to 5mpbs
```
ro-tex, line 60: We don't need TODOs here. Just change the values and we can make out the reason based on your commit message, if we really need to.

ro-tex (approved):

MSevey (approved):

PR #115, opened 2021-12-21
Expose JWT token as a header on login and registration

ro-tex:
# PULL REQUEST

## Overview

This PR exposes the JWT token as a header on login and register requests. This is done for mysky's benefit.

## Checklist
 - [ ] Changelog file created


m-cat (approved):

api/handlers.go
```diff
@@ -211,6 +211,7 @@ func (api *API) loginPOSTToken(w http.ResponseWriter, req *http.Request) {
 		api.WriteError(w, err, http.StatusInternalServerError)
 		return
 	}
+	w.Header().Set("Skynet-Token", tokenStr)
```
peterjan, line 214: is this consistent naming? should it be `Skynet-JWT-Token`?

peterjan (approved):
F/U add tests that verify whether it's present on the routes where it's required

peterjan:
to what service are we exposing this? Doesn't this also require a change to nginx?

api/handlers.go
ro-tex: I think it's fine. It's descriptive enough. I'm not sure we have an established convention for naming these things.

PR #119, opened 2022-01-14
Add docker helper for generating cookie keys and jwks token.

MSevey:
# PULL REQUEST

## Overview
Added a docker helper for generating COOKIE variables and JWKS json file.

## Example for Visual Changes

```
$ cat docker/env 
COOKIE_HASH_KEY=d9ce94ed01adb3b4652af929b5774b04abbe22e20e2527fde6fc3ffeff7b796e
COOKIE_ENC_KEY=4aaf700436d34f63080ee5af639c40c416f3eceff90218b899877d74e27c442e
$ cat docker/jwks.json 
{
  "keys": [
    {
      "use": "sig",
      "kty": "RSA",
      "kid": "private:2f293382-c35a-425f-aec5-b1fb29a64816",
      "alg": "RS256",
      "n": "ywfOTr86l9_WDjSYyui_a92flo9r6f8P0h-MNSIO45VtMo0hSSM5KZIxgib5e69pu7M4I80I2dSpL6xIXRoBqzJOLpcRdNvPTG37CBL9w_L82WR1yoB8yZMW8O8LleNR0-PuwqYgdxoNAhJMpb5HrqJArY7JpQAJAoGp10vTQT4eZfA3L1VdyTCjC111iTEk9Z3Dww-uPZha8R9Fb_D4M0Xe1UkcfqFImiCgjzBcmNIe_V5mj_I4lZKkxA5fuTRIUjJ1N1reaOF4lg5lbBw9Kfx8SUCg9e36yZrV5UMbHKXidn47G8d3jaNGfSapzQ0FRQpqTfCFGlFzoEHVBOxTyu9mfSCijJU6RUvCJIB4wCJ4G_sQTgP7Hs3Va432cKlt6BBfaQhYRYi5HJkbXMPpsjsTmWjSW0b_qWFozI9gA_bLTS_nYd4Z1JxiN5-L4VrDuEkMWEnU2XDBc9WxxUWlDxT-B3IPYWJ4uNXu2MPnOJ-GpkgQif3Kpbqtlb_PiNx7U22iOMcF67I1kej4lSPsNOTC3tUi4_eZNs_gwBnt-knaQ2wRbhGh3hXAsTEdC7RQY-29nXKVez4e2fxbY_IsjwzRTW_js2Spp6ohb1c-jdzyPfzHxDB9lLBs34DOj7AlQpa2az5F5_mNa2Fm99u4ccYU045Su1FkIpsQh3C-D50",
      "e": "AQAB",
      "d": "o4jiEGPPmk1daYvMFoRqy0N0LB3R2mTsJV_r21BXldDiZ_ltei-491ceMaubAJlGHT46VfvsySiWLiQjFXrBJVgbyUM9eFclHnbn4JULei3apHBawIXF_cdQjzi8qDq1ymqqAnc1ZznAiI6c1lPyASQH3MRKf3EL0M0INLaKrt6ONaXLjYDBTjCy45Cf_6VqE7zD1fZnbo2tktKRZ509Dm-VEU8WeJkIlbz6Ftvped6lv4euLPv5v-CifpjlWZRXPH4QFEpJI0mZi6SE5RPJ4GvS4F4CSJUrj__7kfL_f4_OXYtHqRR9o8SEHoJp-xQlznqZLbYyQASjkc30-0YwJrJb4X-vqmjRrmJBO0-xnQKN9MF0mcZlvtZiIG_F2HmdRkSPKTjpg0Fr3-7bKQgbFcAPlAz-4c1NS8JLrEN7tOIvlBo6-848wqBdmuBTbY2jlhR-NoB8ogcZQU4cWSJW4b9WrvCZ03Lb9xApy712bN9zByatVGOaYbPF789GqTZjEsOL2QB-NT2NkAYtOh5cD13jm-GosMcVqZV7pJFujE-yEdlMmbBmIRvPlUyoJhNqrAGeO27SsVeBcmTd3TAsc9QdlKnHXuqcR11Pee2h4EHg8hvsP3dLIsRiZ-lsWx_TAr7ozuZECRQH34EQNt7saPeadLPgPmofnW9CE0ii3bE",
      "p": "3y4ArIrqRNKJi9W37sj0qnlvvNFGP8Bs2VFv2CaHrjcL7eXhRqOqNkTM-y1hg5m9OkJxHZlQ8uAGEMjzK1gpFlSz6BBdqURUpkuztHSy7jsjVZL_JdR5xesKyTo10I67n56h0H7Opm9ANNq9T9koQtnffEdc4HFV2v_tJ-FdJc80zrrE9y_zmzgXv2ucW6ZPRMnPItTuC0TyhBdDC-xFqyZFXEB4MjbFTQoaRqB2frlnqgSGe9v_z1f014HrriDmIS2PFd_l8tKteV4bXXebhyc7vbsl_ZLi5yRcmPxBq9MTBkxi3y_euivfopm5VvSXqnXwpqy46UAqDlm2QB3GKw",
      "q": "6ONAIHDlI5g2t1T5Pc3DB5Yee52ysLKNMdQ9bC8xjZrqynoVjuhIehbhyiSRpnGuiHuNTFpkVgdsdVkmMWG5yx9zRT61ytayx4zZUeP_1-XZt8O0-o0nm382evO2pf2I1Pwp1FiM7lvxk8lYspC4SM-9OXE_f-_vEMb-rW5fP_jzbWMJXxiw5iaiAf3KieWTZ733Ic2v23-8OITADShMwBN7plXm-E6rwYNaZFcNnfMDk4lOdRM31pVXrhJ4uLMCZvykdTsGt-kV9hG4bzI_xkc3abbXaCcSUJS-0-_YhSyFaXHwPo9qLklWyIWMWVXsDSIEh8ICbyIrQy1oBLKlVw"
    },
    {
      "use": "sig",
      "kty": "RSA",
      "kid": "public:2f293382-c35a-425f-aec5-b1fb29a64816",
      "alg": "RS256",
      "n": "ywfOTr86l9_WDjSYyui_a92flo9r6f8P0h-MNSIO45VtMo0hSSM5KZIxgib5e69pu7M4I80I2dSpL6xIXRoBqzJOLpcRdNvPTG37CBL9w_L82WR1yoB8yZMW8O8LleNR0-PuwqYgdxoNAhJMpb5HrqJArY7JpQAJAoGp10vTQT4eZfA3L1VdyTCjC111iTEk9Z3Dww-uPZha8R9Fb_D4M0Xe1UkcfqFImiCgjzBcmNIe_V5mj_I4lZKkxA5fuTRIUjJ1N1reaOF4lg5lbBw9Kfx8SUCg9e36yZrV5UMbHKXidn47G8d3jaNGfSapzQ0FRQpqTfCFGlFzoEHVBOxTyu9mfSCijJU6RUvCJIB4wCJ4G_sQTgP7Hs3Va432cKlt6BBfaQhYRYi5HJkbXMPpsjsTmWjSW0b_qWFozI9gA_bLTS_nYd4Z1JxiN5-L4VrDuEkMWEnU2XDBc9WxxUWlDxT-B3IPYWJ4uNXu2MPnOJ-GpkgQif3Kpbqtlb_PiNx7U22iOMcF67I1kej4lSPsNOTC3tUi4_eZNs_gwBnt-knaQ2wRbhGh3hXAsTEdC7RQY-29nXKVez4e2fxbY_IsjwzRTW_js2Spp6ohb1c-jdzyPfzHxDB9lLBs34DOj7AlQpa2az5F5_mNa2Fm99u4ccYU045Su1FkIpsQh3C-D50",
      "e": "AQAB"
    }
  ]
}                  
```
## Checklist
 - [ ] Changelog file created


.gitignore
```diff
@@ -29,3 +29,6 @@ debug.test
 # OSX auto-generated file
 **/.DS_Store
 *.env
+
+# docker builds
+docker
```
kwypchlo, line 34:
```suggestion
docker

```

kwypchlo (dismissed):
accounts-genenv is a bit of confusing name, probably would go with accounts-generate or sth

also docker as output directory name seems to be weird choice name, I would probably go with output

Makefile
```diff
@@ -145,4 +159,17 @@ test-single: export COOKIE_ENC_KEY="65d31d12b80fc57df16d84c02a9bb62e2bc3b633388b
...
+# The sleep is to handle a slight delay in the jwks.json file being generated in
+# the docker container. Without it the file is blank.
+docker-genenv: clean
+	@mkdir docker
+	@docker build . -t accounts-genenv
+	@docker run --name genenv -d accounts-genenv
+	sleep 3
+	@docker cp genenv:/app/env ./docker/env
+	@docker cp genenv:/app/jwks.json ./docker/jwks.json
+	@docker stop genenv || true && docker rm genenv || true
```
ro-tex, line 173: I get what you are doing here but I wonder whether we can achieve the same in a cleaner way if we touch the files before the run and then mount them as volumes. This will allow us to get rid of the `for {}` and we won't need a `sleep 3` either - the run will finish as soon as it's done without any risk of being stopped too early.

ro-tex (commented):
Looks good overall.

The only things I would like to change is the placement of the files (own dir) and the way we get the files we produce out of the docker container.

Makefile
MSevey: So the sleep is really a factor of there needing to be time for the docker container to start up. 
The docker run command returns instantly and then we are immediately issuing docker stop. 
I removed the `for {}` in the main.go file as that actually didn't do anything.

I think the sleep is the simplest way to consistently run this command without error.

ro-tex: Usually when the entry point of a docker container exits, the entire docker container exits as well. If the container is marked with `-rm` then its image is also removed (I might be mixing up `docker` and `docker-compose`). So, if the entry point binary does all it needs to do and exits with 0 then we shouldn't need to call `stop` and `rm`.

In any case, this is a secondary deliberation. The current state should be good enough for what we need. I was worried that three seconds might not be enough on certain systems and that we'd get weird and hard to debug error reports about missing JWKS and such but I might be overthinking there.

ro-tex (changes requested):
It's great that the dockerfile and the code executed in it are no longer in the root dir but it's not that great that they are now mixed with the actual code of the service. Having support code that generates stuff on deployment live in the same dir as active code is not very confusing. Let's move the support code to its own dir which won't be mistaken as part of the service code.

.gitignore
```diff
@@ -29,3 +29,7 @@ debug.test
...
```
kwypchlo, line 34: I think this can be dropped from gitignore now that you moved the generated content to output dir

kwypchlo (changes requested):

ro-tex (approved):

kwypchlo (approved):

PR #124, opened 2022-02-02
General API Keys

ro-tex:
# PULL REQUEST

## Overview

This PR implements the generic version of API keys:
* created by the user on their dashboard
* do not expire
* revocable
* valid as an authentication mechanism ([via a header or a get parameter](https://swagger.io/docs/specification/authentication/api-keys/))
* give full access to all user actions
* private

## Checklist
 - [ ] Changelog file created


test/api/handlers_test.go
```diff
@@ -78,6 +78,10 @@ func TestHandlers(t *testing.T) {
 		{name: "Challenge-Response/Registration", test: testRegistration},
 		// GET /register, POST /register, GET /login, POST /login
 		{name: "Challenge-Response/Login", test: testLogin},
+		// GET /user/apikey, POST /user/apikey, DELETE /user/apikey/:apiKey
+		{name: "APIKeysFlow", test: testAPIKeysFlow},
+		// POST /user/apikey, GET /user/stats?api_key
```
peterjan, line 83: took me a while to figure out what this means 😛

test/api/apikeys_test.go
```diff
@@ -0,0 +1,127 @@
...
+		t.Fatal(err)
+	}
+	var ak database.APIKey
+	err = json.Unmarshal(body, &ak)
+	if err != nil {
+		t.Fatal(err)
+	}
+	// Get user stats without a cookie or headers - pass the API key via a query
+	// variable.
+	_, body, err = at.Get("/user/stats?api_key="+ak.Key, nil)
```
peterjan, line 118: are we being consistent here? just want to make sure the `_` is well thought through

test/api/apikeys_test.go
```diff
@@ -0,0 +1,127 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	var us database.UserStats
+	err = json.Unmarshal(body, &us)
+	if err != nil {
+		t.Fatal(err)
+	}
```
peterjan, line 126: `semi-nit` we probably want to assert that the user stats effectively match the stats for the user we created

test/api/handlers_test.go
```diff
@@ -78,6 +78,10 @@ func TestHandlers(t *testing.T) {
 		{name: "Challenge-Response/Registration", test: testRegistration},
 		// GET /register, POST /register, GET /login, POST /login
 		{name: "Challenge-Response/Login", test: testLogin},
+		// GET /user/apikey, POST /user/apikey, DELETE /user/apikey/:apiKey
```
peterjan, line 81: I think it should be `/user/apikeys` seeing a user can have multiple?

test/api/apikeys_test.go
peterjan, line 118: use `url.Values` for this, it's easy to extend if we decide to add parameters later

database/apikeys.go
```diff
@@ -0,0 +1,100 @@
...
+// APIKeyDelete deletes an API key.
+func (db *DB) APIKeyDelete(ctx context.Context, user User, ak string) error {
+	if user.ID.IsZero() {
+		return errors.New("invalid user")
+	}
+	filter := bson.M{
+		"key":     ak,
+		"user_id": user.ID,
+	}
+	_, err := db.staticAPIKeys.DeleteOne(ctx, filter)
```
peterjan, line 76: does this return `mongo no documents` error when it can't find it? maybe we should make that a no-op?

database/apikeys.go
```diff
@@ -0,0 +1,100 @@
...
+	return err
+}
+
+// APIKeyList lists all API keys that belong to the user.
+func (db *DB) APIKeyList(ctx context.Context, user User) ([]*APIKey, error) {
+	if user.ID.IsZero() {
+		return nil, errors.New("invalid user")
+	}
+	aks := make([]*APIKey, 0)
```
peterjan, line 85: `var aks []*APIKey`

peterjan: would also define this closer to reading from the cursor, no reason to do it here right?

database/apikeys.go
```diff
@@ -0,0 +1,100 @@
...
+	c, err := db.staticAPIKeys.Find(ctx, bson.M{"user_id": user.ID})
+	if err != nil {
+		return nil, err
+	}
+	defer func() {
+		if errDef := c.Close(ctx); errDef != nil {
```
peterjan, line 91: I believe this is redundant, you should only close a cursor manually if you don't exhaust it fully, which `.All` does.

database/database.go
```diff
@@ -48,6 +48,8 @@ var (
 	// dbConfiguration defines the name of the db table with configuration
 	// settings.
 	dbConfiguration = "configuration"
+	// dbAPIKeys defines the name of the db table with API keys for users.
+	dbAPIKeys = "api_keys"
```
peterjan, line 52: I've seen this a couple of times, I never really made a remark on it but I guess perhaps we should think to clean it up. In mongo you have a database that has collections, we often misuse the `db` prefix to label collections. This should be a collection I think so it should be `collAPIKeys`.

Another thing is the underscores, there's no real conventions around this but in general for collection names you want to go lowercase and avoid camel casing and underscores, I think a better collection name would be `apikeys`. Which is perfectly fine and is just way simpler.

database/apikeys.go
```diff
@@ -0,0 +1,100 @@
...
+	ior, err := db.staticAPIKeys.InsertOne(ctx, ak)
+	if err != nil {
+		return nil, err
+	}
+	ak.ID = ior.InsertedID.(primitive.ObjectID)
+	return &ak, nil
+}
+
+// APIKeyFetch fetches an API key from the DB.
+func (db *DB) APIKeyFetch(ctx context.Context, user User, id primitive.ObjectID) (*APIKey, error) {
```
peterjan, line 51: is this function used? the fact that it expects a `primitive.ObjectID` is a bit of a flag for me, the object id should not be exposed right? so how can we provide it to fetch an API key and where is it used

database/apikeys.go
```diff
@@ -0,0 +1,100 @@
...
+)
+
+// APIKeyCreate creates a new API key.
+func (db *DB) APIKeyCreate(ctx context.Context, user User) (*APIKey, error) {
+	if user.ID.IsZero() {
+		return nil, errors.New("invalid user")
+	}
+	ak := APIKey{
+		UserID:    user.ID,
+		Key:       base64.URLEncoding.EncodeToString(fastrand.Bytes(PubKeySize)),
```
peterjan, line 39: Let's add an index that's unique over `user_id` and `key`, that makes sense I think because your deletes do `DeleteOne`. I know it'll "never" happen but one day we'll manually insert these, or migrate these, or whatever, and we'll forget about the constraint.

api/routes.go
```diff
@@ -1,16 +1,26 @@
...
+	jwt2 "github.com/lestrrat-go/jwx/jwt"
 
 	"github.com/julienschmidt/httprouter"
 	"gitlab.com/NebulousLabs/errors"
 )
 
+var (
+	// APIKeyHeader holds the name of the header we use for API keys. This
+	// header name matches the established standard used by Swagger and others.
+	APIKeyHeader = "X-API-KEY"
```
peterjan, line 18: when defining headers I think it's important to domain prefix them, perhaps we should call it `X-SKYNET-API-KEY`?

api/routes.go
```diff
@@ -65,17 +80,26 @@ func (api *API) noValidate(h httprouter.Handle) httprouter.Handle {
...
-			api.staticLogger.Debugln("Error fetching token from request:", err)
-			api.WriteError(w, err, http.StatusUnauthorized)
-			return
-		}
-		token, err := jwt.ValidateToken(tokenStr)
-		if err != nil {
-			api.staticLogger.Debugln("Error validating token:", err)
-			api.WriteError(w, err, http.StatusUnauthorized)
-			return
+		var token jwt2.Token
```
peterjan, line 83: maybe it makes sense to rename this function to `WithAuthentication`, to be consistent with `WithDBSession` and also to be more descriptive, I think `validate` is not a good name for this middleware

api/routes.go
```diff
@@ -91,53 +115,57 @@ func (api *API) logRequest(r *http.Request) {
...
-	tokenMap, err := token.AsMap(r.Context())
-	if err != nil {
-		return nil
+	if ak == "" {
+		return "", ErrNoAPIKey
 	}
-	sub, exists := tokenMap["sub"]
-	if !exists {
-		return nil
+	return ak, nil
```
peterjan, line 140: should we validate the API key here? should we make it its own type to have some extra type safety?

api/apikeys.go
```diff
@@ -0,0 +1,69 @@
...
+import (
+	"net/http"
+
+	"github.com/SkynetLabs/skynet-accounts/jwt"
+	"github.com/julienschmidt/httprouter"
+)
+
+// userAPIKeyPOST creates a new API key for the user.
+func (api *API) userAPIKeyPOST(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
+	sub, _, _, err := jwt.TokenFromContext(req.Context())
```
MSevey, line 12: nit: seems like we also have this and don't use the `sub` outside of getting the user. Could refactor to a helper.

api/apikeys.go
```diff
@@ -0,0 +1,69 @@
...
+	if err != nil {
+		api.WriteError(w, err, http.StatusInternalServerError)
+		return
+	}
+	api.WriteJSON(w, aks)
+}
+
+// userAPIKeyDELETE removes an API key.
+func (api *API) userAPIKeyDELETE(w http.ResponseWriter, req *http.Request, ps httprouter.Params) {
+	sub, _, _, err := jwt.TokenFromContext(req.Context())
```
peterjan, line 52: I think we can simplify this a bit. If you adopt the suggestion to rename `validate` to something like `WithAuthentication` we can return `http.StatusUnauthorized` from the middleware. That would avoid checking this manually in all these handlers.

Furthermore, I noticed that we are fetching a user from the database a lot. In a single request's lifetime I think it happens at least twice in all of the happy flows. Perhaps we should make use of `context.WithValue` and attach the user, or decorate the request object with the user object.

api/apikeys.go
```diff
@@ -0,0 +1,69 @@
...
+	if err != nil {
+		api.WriteError(w, err, http.StatusUnauthorized)
+		return
+	}
+	u, err := api.staticDB.UserBySub(req.Context(), sub, true)
+	if err != nil {
+		api.WriteError(w, err, http.StatusInternalServerError)
+		return
+	}
+	ak, err := api.staticDB.APIKeyCreate(req.Context(), *u)
```
MSevey, line 22: Is there a reason we would want to limit this? I'm thinking about how we handle wallet addresses in `skyd`. Instead of blindly creating new keys all the time, try and return an existing key first. 

just a thought, if users are generating keys programmatically, they might add a bunch of bloat to mongo if they accidentally great a new key for every request.

api/handlers.go
```diff
@@ -384,17 +378,21 @@ func (api *API) userGET(w http.ResponseWriter, req *http.Request, _ httprouter.P
...
 func (api *API) userLimitsGET(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
-	t, err := tokenFromRequest(req)
-	if err != nil {
-		api.WriteJSON(w, database.UserLimits[database.TierAnonymous])
-		return
+	// First check for an API key.
+	ak, err := apiKeyFromRequest(req)
+	if err == nil {
+		u, err := api.staticDB.UserByAPIKey(req.Context(), ak)
+		if err == nil {
```
MSevey, line 385: Why wouldn't an error here be a reason to fail?  I would think that if you can't get a user by their API key that something is wrong or at least the request is invalid.

peterjan (changes requested):
Great start I think. Is there an API keys spec? I think the current format is quite loose? It doesn't really consider future extensions, like permissions for example. Should we think about stuff like that already? I don't mind it being loosely defined, I'm just asking.

database/apikeys.go
peterjan: I later found out you have it but on just the `key`, which is fine. Leaving the comment here in case I'm wrong and it was an oversight on your part, but I don't think that's the case, having it on `key` is the way to go.

api/routes.go
```diff
@@ -65,17 +80,26 @@ func (api *API) noValidate(h httprouter.Handle) httprouter.Handle {
...
-			return
+		var token jwt2.Token
+		// Check for an API key. We only return an error if an invalid API key
+		// is provided.
+		ak, err := apiKeyFromRequest(req)
+		if err == nil {
+			// We have an API key. Let's generate a token based on it.
+			token, err = api.tokenFromAPIKey(req.Context(), ak)
+			if err != nil {
+				api.staticLogger.Debugln("Error generating token for API key:", err)
```
MSevey, line 91: nit: in general I would have all errors logs as printlns so that we don't need to rebuild in dev mode to get error logs.

api/routes_test.go
```diff
@@ -0,0 +1,128 @@
...
+	if tk != token2 {
+		t.Fatalf("Expected '%s', got '%s'.", token2, tk)
+	}
+}
+
+// TestTokenFromRequest ensures that tokenFromRequest works as expected.
+func TestTokenFromRequest(t *testing.T) {
+	// FIXME This is a weird and ugly workaround for the fact that I don't know
+	// 	how to refer to a file relative to the project's root dir.
+	jwt.AccountsJWKSFile = "../jwt/fixtures/jwks.json"
```
MSevey, line 58: this is how we do it in skyd fwiw. we have a whole `compatibility` directory of files for tests and the tests just reference them with relative paths.
you could make a const for the `fixtures` directory path if it makes you feel better.

MSevey (commented):
Generally lgtm

test/api/handlers_test.go
ro-tex: The name or the comments?

The comments are a bit wacky. I started them as little notes for myself while writing the tests and then I somehow kept them. I'm not convinced it's a good idea to have them there, as they might not be terribly useful for others. Should I remove them?

ro-tex:
> Great start I think. Is there an API keys spec? I think the current format is quite loose? It doesn't really consider future extensions, like permissions for example. Should we think about stuff like that already? I don't mind it being loosely defined, I'm just asking.

This is the spec: https://hackmd.io/tB6WDA73Q7iQGSFTqinVZg

My intention was to get the simple case going first and then build on top of it. We agreed that there won't be any permissions for the moment, as that would expand the scope a lot.

We want to have the so-called "sponsor" api keys which allow anyone to download a skylink with the owner of the api key paying for the traffic. I see those as a separate PR because there will be quite some code around them as well and I don't want this to be a 2000 lines PR. I'd rather iterate over the idea in 2, 3, even 5 PRs.

api/routes.go
ro-tex: We don't need to rebuild - there's an env var called `SKYNET_ACCOUNTS_LOG_LEVEL` which is read on startup and configures the logger. This allows me to put way more logging than generally needed and spread it across levels, so I can turn more and more of it until I see enough to be able to diagnose the problem without drowning in a sea of text.

If anything, I would like to make it more modular, so I can increase the log level in a specific module in order to diagnose what's happening there. I've used this approach in the past and it was amazing for quickly pinning down where the issue is.

api/handlers.go
ro-tex: Makes sense. If there is a supplied API key and we cannot match it to a user we'll return `anonymous` limits.

test/api/apikeys_test.go
ro-tex: I agree that this goes against our conventions. The reason I used this is that wherever I read about API keys, I always stumbled on the Swagger standard for those (https://swagger.io/docs/specification/authentication/api-keys/) and that works with the `X-API-KEY` header or the `api_key` query param. It's a choice between compatibility with a well-established web standard and our conventions. I'm very much open to suggestions here - I'd be happy to change that to `apiKey`, if that's the consensus.

test/api/apikeys_test.go
ro-tex: Sure. I'm not convinced that it would add anything to the test but I don't see a problem with adding that validation.

The reason I don't see it as very useful is that the aim of the test it make sure that supplying the API key is enough for the call to go through the `api.validate()` middleware check. As long as I'm not getting an `Unauthorized` here, I'm happy. :) But I'll add the validation - there is no harm in that.

api/apikeys.go
```diff
@@ -0,0 +1,69 @@
...
+	if err != nil {
+		api.WriteError(w, err, http.StatusUnauthorized)
+		return
+	}
+	u, err := api.staticDB.UserBySub(req.Context(), sub, true)
+	if err != nil {
+		api.WriteError(w, err, http.StatusInternalServerError)
+		return
+	}
+	aks, err := api.staticDB.APIKeyList(req.Context(), *u)
```
ChrisSchinnerl, line 42: Referring to my previous comment. We might not actually want to return the actual keys here but only some information like a friendly name or id, when it was created, when it was used the last time etc. Basically information that tells a user they have keys and a way to identify them but actually the secret.

Neither Github nor Gitlab seem to allow you to fetch a token after its creation so it might be good practice now to not hand them out after creation.

api/apikeys.go
ChrisSchinnerl: I think this aligns with my comment above around middleware that basically passes you the User after implicitly fetching it.

api/handlers.go
```diff
@@ -190,18 +190,12 @@ func (api *API) loginPOSTCredentials(w http.ResponseWriter, req *http.Request, e
 func (api *API) loginPOSTToken(w http.ResponseWriter, req *http.Request) {
 	// Fetch a JWT token from the request. This token will tell us who the user
 	// is and until when their current session is going to stay valid.
-	tokenStr, err := tokenFromRequest(req)
+	token, tokenStr, err := tokenFromRequest(req)
```
ChrisSchinnerl, line 193: Seems a bit redundant and prone to inconsistencies that it would return both `token` and `tokenStr` instead of just the token with a `String()` method that would be called if you ever need to represent the token as a string again.

api/routes.go
ChrisSchinnerl: Yeah. If the user supplies something we should parse it, validate it and then put it in a type. Having a type will make sure we will never use unvalidated user input by accident.

test/api/handlers_test.go
ro-tex: Sure, we're using a plural in uploads and downloads, so we can use one here as well.

api/handlers.go
```diff
@@ -384,17 +378,21 @@ func (api *API) userGET(w http.ResponseWriter, req *http.Request, _ httprouter.P
 
 // userLimitsGET returns the speed limits which apply to this user.
 func (api *API) userLimitsGET(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
-	t, err := tokenFromRequest(req)
-	if err != nil {
-		api.WriteJSON(w, database.UserLimits[database.TierAnonymous])
-		return
+	// First check for an API key.
+	ak, err := apiKeyFromRequest(req)
```
ChrisSchinnerl, line 382: Once we have middleware this can probably also be moved to that.

e.g. we could have a `mustAuthenticate` and a `canAuthenticate` middleware. This would use `canAuthenticate` and would potentially pass a nil user that you can check for to see if the user was either loged in via api key or token.

api/routes.go
ChrisSchinnerl: I'd just use `Skynet-Api-Key` to match all our other headers like `Skynet-Disable-Force`, `Skynet-File-Layout`, `Skynet-File-Metadata`, `Skynet-Proof` and `Skynet-Skylink`.

The `X-` convention for custom headers has been deprecated for almost 10 years as part of some RFC.

database/apikeys.go
ro-tex: Hm, good catch, I didn't think this one through.

What do you think about a Bad Request if the record is not found? It's a bit more in line with REST than the no-op approach and it gives the caller reassurance when they call DELETE on the same API key twice and get one OK and one Bad Request after that - they are certain the API key is deleted.

api/routes_test.go
ChrisSchinnerl: Yep this is fine. If you want to be safe that it works on all OSs you could use `filepath.Join` although we don't do that everywhere in Sia either. Go might just magically make it work on Windows or maybe Windows itself does even with the `/`.

test/api/handlers_test.go
ChrisSchinnerl: If you want to keep them I'd drop a quick comment in line 51 to explain what they mean.

database/apikeys.go
ro-tex: Done.

database/apikeys.go
ro-tex: You are right. I hadn't noticed that All closes the cursor.

database/database.go
ro-tex: That's a good point about the prefix. I'll change that.

I'm not so sure about the collection name, though. We're already using a bunch of underscores around and changing that will need migration code. I'd rather keep the underscores.

test/api/apikeys_test.go
ChrisSchinnerl: I wouldn't go out of our way to try and be compliant with some standard that's mostly useful for autogenerating APIs if we don't plan on doing that ourselves anyway. We probably diverge in way too many places already anyway.

database/user.go
```diff
@@ -159,6 +159,20 @@ type (
 	}
 )
 
+// UserByAPIKey returns the user who owns the given API key.
+func (db *DB) UserByAPIKey(ctx context.Context, ak string) (*User, error) {
+	sr := db.staticAPIKeys.FindOne(ctx, bson.M{"key": ak})
```
ChrisSchinnerl, line 164: f/u: I wonder if there is a way to try and look up a user by both api key as well as token in a single query to reduce the number of trips to the db.

database/database.go
ChrisSchinnerl: Agreed

database/apikeys.go
ro-tex: Ha! No, it's not. I had an endpoint that fetched a single API key before (and the WIP version used `ObjectID`) but I removed it. I've forgotten to clean up. Thanks! :)

database/apikeys.go
ChrisSchinnerl: I think you are right that the cursor gets closed when there are no more elements. But I would still like to keep it here. For two reasons.

1. you never know how the code below changes and always closing a `io.Closer` once you are done with it is just great practice. 
2. I don't know what happens if `c.All` fails halfway through.

database/apikeys.go
ChrisSchinnerl: No, it will not return an error. But the delete result (which we ignore) will contain 0 elements.

database/apikeys.go
ChrisSchinnerl: Agreed. Even if it is used we should avoid weird internal types.

api/routes.go
ro-tex: I already mentioned this elsewhere but using `X-API-KEY` makes us compatible with the Swagger implementation of API keys. If we don't want that then we can rename the query param to `apiKey` as well.

ChrisSchinnerl (changes requested):

api/apikeys.go
ro-tex: This is a great suggestion. Do you mind if we do it as a F/U? I'm afraid it's going to touch far too many lines in the service and while it's going to make it smaller and faster, it's going to make this PR far too large.

F/U: #125

test/api/handlers_test.go
ro-tex: After thinking about it a bit, I don't think they bring enough value.

api/apikeys.go
ro-tex: PJ proposed below to combine with with the `validate` (now `withAuth`) middleware and embed the user var in the context instead of the JWT token. I think that's a great solution for this.

api/apikeys.go
ro-tex: I'm not sure we should worry about this at this point. Creating an API key is one more request and you also need to store it somewhere, so that's extra I/O on your end. So, I'm more inclined to think that people would err on the side of creating too few API keys rather than too many. We might get the occasional skapp that would just spam these like crazy but I don't think that's a problem. MongoDB can handle *a lot* of documents in one table.

We can definitely jump on a call/channel and discuss this.

test/api/apikeys_test.go
ro-tex: Swagger has been on my mind for quite a while now. I've used if before and it's some of the best API docs I've ever used. Combined with a developer portal with allows you to make API calls to a test server from a web page by pasting your request bodies/params, it makes for an amazing dev tool.

That said, I don't think there's anything preventing us from using Swagger in the future with non-standard headers. If you prefer the `X-SKYNET-API_KEY`/`apiKey` approach, I don't mind.

database/apikeys.go
ro-tex: This calls for a unit test.

api/routes.go
ro-tex: > I'd just use `Skynet-Api-Key` to match all our other headers like `Skynet-Disable-Force`, `Skynet-File-Layout`, `Skynet-File-Metadata`, `Skynet-Proof` and `Skynet-Skylink`.
> 
> The `X-` convention for custom headers has been deprecated for almost 10 years as part of some RFC.

I didn't see this when writing my reply. Sounds good to me, I'm attached to the header name. Changing.

api/routes_test.go
ro-tex: I'm not sure I'm 100% happy with this staying the final solution but I *am* happy that we're thinking alike.

api/apikeys.go
ro-tex: Good point. We shouldn't display them more than once. I'll change the list endpoint to return only the key's ID and time of creation and the delete endpoint to take a key ID.

api/handlers.go
ro-tex: I was thinking about this, too. I kept the string return value because it simplifies the code elsewhere by removing the need to serialize the token and handle the error.

I removed it now and it looks OK. It still seems like an overhead to not return the string, as we already have it right there, but it's not so bad.

api/handlers.go
ro-tex: `withAuth` is already a middleware, so we have that. We additionally call `apiKeyFromRequest` from `userLimitsGET` because we want to specifically not use the middleware for that endpoint because we want it to handle both authenticated and unauthenticated calls.

database/user.go
ro-tex: If we find an API key in the request (no DB call so far) we try to fetch the user based on it and if we fail we return an error. So, there is no scenario in which we make two trips to the DB in order to find the authenticated user.

We *do* make two trips in case the user supplies an API key and we get them from the DB, create a token for them, and then embed the token in the context. Later on we fetch the user from the DB based on that. But this inefficiency has already been pointed out and we have a solution for it - fetch the user from the DB in `withAuth` (either by API key or by token) and embed the user object in the context instead of the token.

Based on that, I don't think we need an extra follow up here.

database/apikeys.go
peterjan: I disagree but I don't mind. If `c.All` fails the cursor will get closed anyway. This `defer` and extra close is just completely redundant and adds complexity, it's good practice to close your resources obviously but in this particular case it is not necessary, it's very much like closing the request body in the HTTP handlers. But I don't mind if you keep it in. Just for this one time, I disagree with Chris 😛 let it be known.

test/api/apikeys_test.go
```diff
@@ -0,0 +1,159 @@
...
+	r, _, err := at.CreateUserPost(name+"@siasky.net", name+"_pass")
+	if err != nil {
+		t.Fatal(err)
+	}
+	at.Cookie = test.ExtractCookie(r)
+
+	aks := make([]database.APIKeyRecord, 0)
+
+	// List all API keys this user has. Expect the list to be empty.
+	r, body, err := at.Get("/user/apikeys", nil)
```
ChrisSchinnerl, line 28: f/u at some point we should clean these hardcoded routes up. Ideally we have a single client implementation that we extend like in skyd and just call the helper on that instead of keeping the same `GET` followed by a `json.Unmarshal` around 8 or so times in more than 1 test function.

peterjan (approved):
LGTM, but we should def. make sure to F/U with all of the F/U's

api/apikeys.go
ro-tex: F/U: https://github.com/SkynetLabs/skynet-accounts/issues/127

api/handlers.go
ro-tex: F/U: https://github.com/SkynetLabs/skynet-accounts/issues/127

cmd/crdb_transition/main.go
```diff
@@ -259,7 +235,7 @@ func mongoEmailsUnique(ctx context.Context, db *mongo.Database) error {
 }
 
 func latestMigrationTS(ctx context.Context, mgr *mongo.Database) (time.Time, error) {
```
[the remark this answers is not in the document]
ro-tex: F/U: https://github.com/SkynetLabs/skynet-accounts/pull/127

test/api/apikeys_test.go
ro-tex: F/U: https://github.com/SkynetLabs/skynet-accounts/issues/128

PR #131, opened 2022-02-14
Better information provided by /user/limits

### LINKED ISSUE #121: /limits should include active subscription status and boolean whether it's anonymous or authenticated
Instead of current "tierName" which can be arbitrary, we need to make sure nginx knows whether user is authenticated and whether has active subscription to better support https://github.com/SkynetLabs/skynet-webportal/pull/1575 - booleans would do.

ro-tex:
# PULL REQUEST

## Overview

Endpoint `/user/limits` should provide numerical representation of the user's tier in order to make tier detection more reliable by not requiring string comparisons. This will also make the detection of when a user is logged in better.

Notes:
* `tierID == 0` means that the user is not logged in
* `tierID == 1` means that the user is logged in as a free user
* `tierID > 1` means that the user is logged in and has active paid subscription

The full current tier information is available here: https://account.skynetpro.net/api/limits

### Before:
```json
{
    "tierName": "anonymous",
    "upload": 655360,
    "download": 655360,
    "maxUploadSize": 1073741824,
    "registry": 250
}
```
### After:
```json
{
    "tierID": 0,
    "tierName": "anonymous",
    "upload": 655360,
    "download": 655360,
    "maxUploadSize": 1073741824,
    "registry": 250
}
```
## Checklist
 - [ ] Changelog file created

## Issues Closed
Closes #121


kwypchlo (approved):
approving based on api docs, I'll leave code review to @peterjan and @ChrisSchinnerl :)

api/handlers.go
```diff
@@ -371,25 +371,33 @@ func (api *API) userGET(u *database.User, w http.ResponseWriter, _ *http.Request
 func (api *API) userLimitsGET(_ *database.User, w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
 	// First check for an API key.
 	ak, err := apiKeyFromRequest(req)
+	respAnon := database.UserLimitsResponse{
```
ChrisSchinnerl, line 374: The types we respond with should be dedicated types in this file. It's not obvious for someone touching a type in `database` that the API response changes as well.

It's also not obvious to me how a number is going to be better than a string identifier for the tier from this MR.

main.go
```diff
@@ -160,12 +160,14 @@ func main() {
 		}
 	}
 	// Fetch the configuration for maximum number of API keys allowed per user.
-	maxAPIKeys, err := strconv.Atoi(os.Getenv(envMaxNumAPIKeysPerUser))
-	if err != nil {
-		log.Printf("Warning: Failed to parse %s env var. Error: %s", envMaxNumAPIKeysPerUser, err.Error())
-	}
-	if maxAPIKeys > 0 {
-		database.MaxNumAPIKeysPerUser = maxAPIKeys
+	if maxAPIKeysStr := os.Getenv(envMaxNumAPIKeysPerUser); maxAPIKeysStr != "" {
```
ChrisSchinnerl, line 163: I'd use `LookupEnv` here. That way you can check if the user had the intention to set the environment variable and you can return an error if the user set it to an empty string by accident.

test/api/api_test.go
```diff
@@ -205,13 +205,13 @@ func TestUserTierCache(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	var ul database.TierLimits
+	var ul database.UserLimitsResponse
 	err = json.Unmarshal(b, &ul)
 	if err != nil {
 		t.Fatal(err)
 	}
-	if ul.TierName != database.UserLimits[database.TierFree].TierName {
```
ChrisSchinnerl, line 213: We should still check for the id and the name for backwards compatibility.

ChrisSchinnerl (changes requested):

main.go
ro-tex: Hmm. Am I alone in thinking that when I set and env var to "" it's the same as if it's gone? I mean, I know that it's detectable but I usually presume that the software will ignore the empty value.

That said, I don't mind. This will print a warning but it won't block. I'll change it.

test/api/api_test.go
ro-tex: OK.

api/handlers.go
ro-tex: Watch this PR explode now!

I do agree, though. We should use dedicated types for data representation and dedicated types for user communication. I'll add those. It will blow up the PR but it's useful and it should also be easy to review.

api/handlers.go
```diff
@@ -30,9 +30,27 @@ const (
...
-	// LimitsPublic provides public information of the various limits this
+	// ChallengePublic is the response of GET /login, GET /register,
+	// GET /user/pubkey/register
+	ChallengePublic struct {
+		// Challenge is a hex-encoded representation of the []byte challenge.
+		Challenge string `bson:"challenge" json:"challenge"`
+	}
+	// DownloadsGET is the response of GET /user/downloads
+	DownloadsGET struct {
+		Items    []database.DownloadResponse `json:"items"`
```
ChrisSchinnerl, line 41: Why is `database.DownloadResponse` in `database` instead of in here?
We can f/u with this but it seems a bit strange.

Overall I'd avoid `database` types as fields in the request/response types.
Same reason as before. It's just not obvious enough that a change to the persistence type updates the http response.

ChrisSchinnerl (approved):

api/handlers.go
ro-tex: I tried to move them all from `database` to `api` but I've missed this one.

I'm not sure I'm 100% onboard with your general line of thinking. Anyone who sees a `json` annotation on a data type should be immediately aware that this is an API thing. Our storage uses `bson` and our API uses `json` - it shouldn't be a surprise, IMO.

We should, by all means, place any API-only fields in specific types used for API-level communication but I don't see any harm in using the DB-level structs as modules of the API-level structs. Otherwise we'll end up manually deep copying data structures whenever we want to send anything over the API. If this is what's expected then I'll need a much bigger refactoring but I would also like to discuss it live before we go that far.

peterjan: In this particular case, it's probably just because the struct name has the word "Response" in it, making it feel a bit quirky/weird. If it would've been `[]database.Item` it wouldn't have triggered me personally. I think Chris's argument about updating persistence indirectly changing the API response is a really good point. Perhaps we can simply get away with it though by dropping a `NOTE` in the docstring that warns developers any change to the struct has a direct effect on what's returned by the API.

main.go
```diff
@@ -160,12 +160,14 @@ func main() {
 		}
 	}
 	// Fetch the configuration for maximum number of API keys allowed per user.
-	maxAPIKeys, err := strconv.Atoi(os.Getenv(envMaxNumAPIKeysPerUser))
-	if err != nil {
-		log.Printf("Warning: Failed to parse %s env var. Error: %s", envMaxNumAPIKeysPerUser, err.Error())
-	}
-	if maxAPIKeys > 0 {
-		database.MaxNumAPIKeysPerUser = maxAPIKeys
+	if maxAPIKeysStr, ok := os.LookupEnv(envMaxNumAPIKeysPerUser); !ok {
```
peterjan, line 163:
```suggestion
	if maxAPIKeysStr, ok := os.LookupEnv(envMaxNumAPIKeysPerUser); ok {
```
right?

api/handlers.go
peterjan: By the way, if these structs are your database entities and you return them as-is in the API responses, I actually feel they should live in `database` rather than `api`. So I would actually move all of the other ones back to `database` even. Your database entities belong in the database package right.

main.go
ro-tex: *facepalm*
yes...

Since I'm changing this, I'll use the opportunity to add a warning when the value is being ignored - Chris requested it for another parameter, it only makes sense to have that here as well.

api/handlers.go
ro-tex: I only moved types which are only used as responses. Where there were no such types, I created them. So they are not database types. I can't access types in module `api` from module `database` because of circular dependencies.

ChrisSchinnerl: Right the `DownloadResponse` name makes it seem like it's an API response because I'd usually not consider something coming from a DB a response.

ChrisSchinnerl (approved):

peterjan (approved):

PR #132, opened 2022-02-15
Improved environment variables processing

### LINKED ISSUE #123: F/U JWT TTL env var with a unit test and an err on 0
## Overview

F/U https://github.com/SkynetLabs/skynet-accounts/pull/122 with a unit test and an error when the value is set to zero.

Ref: https://discord.com/channels/542938080349519882/820776189357981736/937983688166240306

ro-tex:
# PULL REQUEST

## Overview

We currently have no way of unit testing our handling of environment variables. We want to change that.
What we want in this PR:
* Move all environment variables parsing into a separate function.
* Add unit tests for that function.

## Checklist
 - [ ] Changelog file created

## Issues Closed
Closes #123


main.go
```diff
@@ -160,12 +155,38 @@ func main() {
...
+	_ = godotenv.Load()
+	dbCreds, err := loadDBCredentials()
 	if err != nil {
-		log.Printf("Warning: Failed to parse %s env var. Error: %s", envMaxNumAPIKeysPerUser, err.Error())
+		log.Fatal(errors.AddContext(err, "failed to fetch DB credentials"))
 	}
-	if maxAPIKeys > 0 {
-		database.MaxNumAPIKeysPerUser = maxAPIKeys
+
+	emailURI, err := parseEnvironmentVariables(logger)
```
peterjan, line 187: I think it's a great idea to improve this, but in this case I think it's a bit quirky in the sense that sometimes it returns the env variable and other times it will go and overwrite exposed variables on other packages. It's quite magical right, you call it and you don't know what's going to happen at all.

I found that some of the variables don't mention in the docstring they are (possibly) overwritten by a value on the environment, I think `jwt.TTL` is an example of where it's missing. I dislike that pattern personally but that's just personal preference I guess, it's quite annoying to debug IMO because if you would look for the TTL, you'd find a certain value and not know whether it got overwritten, in this case you wouldn't even know it gets overwritten, but that's just because of the oversight of mentioning it in the docstring, which is easily overlooked.

peterjan (dismissed):
I would personally either add an environment validator or have the parse function return a helper struct that defines and contains the env variables. Both those solutions would improve readability a bit and lower the overall magic that happens.

But nothing wrong with this PR, it's a bit of a nit I know so definitely approve! 
Great that you took the time to improve this stuff, lord knows I have some repos that could use something like this.

main.go
ro-tex: That's a valid concern, on both counts.

On the side of variables that do not state which env vars control them the solution is pretty obvious - I'll just update their docstrings.

On the side of variables being directly set on packages instead of being returned - I'm not sure how I want to attack that.
* Having the parsing method return a bunch of variables, so we can set them in `main` doesn't sound appealing. 
* Adding a struct that defines all env vars and using that as return value doesn't improve the situation much. 
* Moving all these values to a separate package might be a way to go but it will move the variables themselves away from the code that uses them. The initialisation flow will be cleaner but the actual usage will be less so.

I am open to suggestions here and I don't have a preference.

main.go
```diff
@@ -15,11 +15,11 @@ import (
...
 	"github.com/SkynetLabs/skynet-accounts/metafetcher"
+	"github.com/joho/godotenv"
+	"github.com/stripe/stripe-go/v71"
 	"gitlab.com/SkynetLabs/skyd/skymodules"
 
-	"github.com/joho/godotenv"
 	"github.com/sirupsen/logrus"
-	"github.com/stripe/stripe-go/v71"
 	"gitlab.com/NebulousLabs/errors"
 )
```
m-cat, line 25: Should the env var names be `const` instead of `var`?

main_test.go
```diff
@@ -0,0 +1,150 @@
...
+	err := os.Unsetenv(envDBUser)
+	if err != nil {
+		t.Fatal(err)
+	}
+	_, err = loadDBCredentials()
+	if err == nil || !strings.Contains(err.Error(), "missing env var "+envDBUser) {
+		t.Log(err)
+		t.Fatal("Failed to error out on missing", envDBUser)
+	}
+	err = os.Setenv(envDBUser, "user")
```
m-cat, line 107: Where are these values e.g. `"user"` coming from? The test should be setting the env vars back to what they were before unsetting them, in a `defer` so that it happens if the test errors out, as well. Also, we should make sure that this test doesn't run in parallel with other tests (I think it doesn't, but have forgotten a lot about Go already).

m-cat (changes requested):
Couple comments.

go.mod
```diff
@@ -8,29 +8,30 @@ require (
 	github.com/go-stack/stack v1.8.1 // indirect
```
ChrisSchinnerl, line 8: Not sure what is doing that but we should stop casually updating a bunch of dependencies on every PR. When multiple people work on the same codebase it makes merging a nightmare and there are other unforeseeable things that might go unnoticed.

main.go
ChrisSchinnerl: I agree with PJ. Modifying global vars outside of the scope of a function is a nightmare to maintain. 
I wouldn't expect a parsing method to actually set anything that's not passed in by reference or returned. Instead I'd want it to return a struct with some configuration values which I then pass on to whatever needs them.

main_test.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+	_, err = parseEnvironmentVariables(logger)
+	if err == nil || !strings.Contains(err.Error(), "invalid email URI") {
+		t.Log(err)
+		t.Fatal("Failed to error out on invalid", envEmailURI)
+	}
+	err = os.Setenv(envEmailURI, "smtps://disabled@example.net:not-a-password@smtp.gmail.com:465/?skip_ssl_verify=false")
+	if err != nil {
+		t.Fatal(err)
+	}
```
ChrisSchinnerl, line 91: If you change this to return a struct we should also check the individual struct values for correctness.

main_test.go
ChrisSchinnerl: This is a good point. We should make sure no other test will be influenced by this. The test shouldn't run in parallel with other tests but I'm not sure if every test is an isolated process with its own set of env vars.

ChrisSchinnerl (changes requested):

main.go
ro-tex: Sure.

main_test.go
ro-tex: Restoring the values is a good point. It doesn't have any direct effect but it's the right thing to do.

I am not worried about this test affecting any other test because no other test should be relying on the same env vars - they are supposed to be read only in `main`. That said, I don't think I have any way of preventing tests from other packages from being affected (go runs tests within the same package sequentially but runs the tests for different packages in parallel). If that happens I'll just remove this test - it's not terribly useful anyway.

The hardcoded test values are just that - test values. But I can just as well use the original values. That might reduce the confusion.

main.go
ro-tex: OK, in that case I'll define a configuration type and I'll return it from the parsing func. Then I'll set the global vars from `main` as before. This should achieve both the visibility we had before and the testability we wanted to get from this PR. It will also simplify `main` quite a bit, something I wanted to get.

main_test.go
ro-tex: Done.

go.mod
ro-tex: I actually updated them in one place (I think?) and since I had to create a bunch of small PRs this one change started bouncing back and forth and causing trouble.

m-cat (approved):

main.go
```diff
@@ -61,6 +61,24 @@ var (
 	envMaxNumAPIKeysPerUser = "ACCOUNTS_MAX_NUM_API_KEYS_PER_USER"
 )
 
+type (
+	// ServiceConfig represents all configuration values we expect to receive
+	// via environment variables or config files.
+	ServiceConfig struct {
+		DBCreds               database.DBCredentials
```
ChrisSchinnerl, line 68: f/u docstrings for fields

main.go
```diff
@@ -96,77 +114,123 @@ func logLevel() logrus.Level {
...
+	database.PortalName = config.PortalName
+	jwt.PortalName = config.PortalName
+	email.PortalAddressAccounts = config.PortalAddressAccounts
+	email.ServerLockID = config.ServerLockID
+	stripe.Key = config.StripeKey
+	api.StripeTestMode = config.StripeTestMode
+	jwt.AccountsJWKSFile = config.JWKSFile
+	jwt.TTL = config.JWTTTL
+	email.From = config.EmailFrom
+	database.MaxNumAPIKeysPerUser = config.MaxAPIKeys
```
ChrisSchinnerl, line 233: I can tell that this is going to blow up this MR so we can refactor it in a f/u but this is a similar issue as before within the parsing function. Global vars are not supposed to be used for configuration. This way you can never create two instances of an object in a thread safe way. e.g. two databases with different portal names in parallel.

You want to set the `MaxNumAPIKeysPerUser`? That should be a `staticMaxNumAPIKeysPerUser` field on the `database` instance and be passed in via something like `database.New(ctx, creds, logger, portalName, maxAPIKeysPerUser)` instead.

ChrisSchinnerl (approved):
approved with f/u

main.go
```diff
@@ -96,77 +114,123 @@ func logLevel() logrus.Level {
...
+	logger := logrus.New()
+	logger.SetLevel(logLevel())
+
+	// Load the environment variables from the .env file.
+	_ = godotenv.Load()
+	config, err := parseConfiguration(logger)
+	if err != nil {
+		log.Fatal(err)
 	}
+	database.PortalName = config.PortalName
```
peterjan, line 224: `nit` could've been grouped by package for clarity

main_test.go
```diff
@@ -0,0 +1,336 @@
...
+	"github.com/SkynetLabs/skynet-accounts/database"
+	"github.com/sirupsen/logrus"
+	"gitlab.com/NebulousLabs/errors"
+)
+
+// TestParseConfiguration ensures that we properly parse and validate all
+// required environment variables.
+func TestParseConfiguration(t *testing.T) {
+	// Fetch current state of env and make sure we restore it on exit.
+	{
```
peterjan, line 19: could've been a separate helper function called`restoreEnv` - I have this in a blocker PR if you want to copy paste

main_test.go
```diff
@@ -0,0 +1,336 @@
...
+	}
+	if config.MaxAPIKeys != maxKeys {
+		t.Fatalf("Expected %d, got %d", maxKeys, config.MaxAPIKeys)
+	}
+}
+
+// TestLoadDBCredentials ensures that we validate that all required environment
+// variables are present.
+func TestLoadDBCredentials(t *testing.T) {
+	{
```
peterjan, line 241: could've reused the `restoreEnv` helper here

peterjan (approved):
approved but left some comments, perhaps we can tackle those in a F/U

overall though this PR doesn't remove the pattern where we overwrite public variables, I know you mentioned the env var names in the docstring now but I still think it's a bit of an anti-pattern, I agree it's convenient but it comes at a cost

ro-tex:
> approved but left some comments, perhaps we can tackle those in a F/U
> 
> overall though this PR doesn't remove the pattern where we overwrite public variables, I know you mentioned the env var names in the docstring now but I still think it's a bit of an anti-pattern, I agree it's convenient but it comes at a cost

This was never the intention of this PR. The goal was to write a unit test which ensures we are parsing the env vars correctly. For that we needed to move said parsing to a separate function. Everything else that happened in this PR is scope creep.

It seems that we want to stop using global (public) variables for configuration. Sure, that can be done but it needs its own PR.

main_test.go
ro-tex: https://github.com/SkynetLabs/skynet-accounts/issues/138

main.go
ro-tex: https://github.com/SkynetLabs/skynet-accounts/issues/138

main.go
ro-tex: https://github.com/SkynetLabs/skynet-accounts/issues/138

main.go
ro-tex: https://github.com/SkynetLabs/skynet-accounts/issues/138

PR #135, opened 2022-02-21
Don't start DB sessions if we don't need them

ro-tex:
# PULL REQUEST

## Overview

Recently we've had some issues with DB writes being cancelled because of overlapping changes. One way we can resolve that is to make sure that we use DB sessions only when we actually need them (read then write, write to multiple documents, etc.). This PR does that. 

## Checklist
 - [ ] Changelog file created

### DIFF
--- a/api/routes.go
+++ b/api/routes.go
@@ -42,7 +42,7 @@ func (api *API) buildHTTPRoutes() {
 	api.staticRouter.GET("/login", api.WithDBSession(api.noAuth(api.loginGET)))
 	api.staticRouter.POST("/login", api.WithDBSession(api.noAuth(api.loginPOST)))
 	api.staticRouter.POST("/logout", api.withAuth(api.logoutPOST))
-	api.staticRouter.GET("/register", api.WithDBSession(api.noAuth(api.registerGET)))
+	api.staticRouter.GET("/register", api.noAuth(api.registerGET))
 	api.staticRouter.POST("/register", api.WithDBSession(api.noAuth(api.registerPOST)))
 
 	// Endpoints at which Nginx reports portal usage.
--- a/api/routes.go
+++ b/api/routes.go
@@ -41,32 +41,32 @@ func (api *API) buildHTTPRoutes() {
 
 	api.staticRouter.GET("/login", api.WithDBSession(api.noAuth(api.loginGET)))
 	api.staticRouter.POST("/login", api.WithDBSession(api.noAuth(api.loginPOST)))
-	api.staticRouter.POST("/logout", api.WithDBSession(api.withAuth(api.logoutPOST)))
+	api.staticRouter.POST("/logout", api.withAuth(api.logoutPOST))
 	api.staticRouter.GET("/register", api.WithDBSession(api.noAuth(api.registerGET)))
 	api.staticRouter.POST("/register", api.WithDBSession(api.noAuth(api.registerPOST)))
 
 	// Endpoints at which Nginx reports portal usage.
-	api.staticRouter.POST("/track/upload/:skylink", api.WithDBSession(api.withAuth(api.trackUploadPOST)))
-	api.staticRouter.POST("/track/download/:skylink", api.WithDBSession(api.withAuth(api.trackDownloadPOST)))
-	api.staticRouter.POST("/track/registry/read", api.WithDBSession(api.withAuth(api.trackRegistryReadPOST)))
-	api.staticRouter.POST("/track/registry/write", api.WithDBSession(api.withAuth(api.trackRegistryWritePOST)))
+	api.staticRouter.POST("/track/upload/:skylink", api.withAuth(api.trackUploadPOST))
+	api.staticRouter.POST("/track/download/:skylink", api.withAuth(api.trackDownloadPOST))
+	api.staticRouter.POST("/track/registry/read", api.withAuth(api.trackRegistryReadPOST))
+	api.staticRouter.POST("/track/registry/write", api.withAuth(api.trackRegistryWritePOST))
 
-	api.staticRouter.POST("/user", api.WithDBSession(api.noAuth(api.userPOST))) // This will be removed in the future.
-	api.staticRouter.GET("/user", api.WithDBSession(api.withAuth(api.userGET)))
+	api.staticRouter.POST("/user", api.noAuth(api.userPOST)) // This will be removed in the future.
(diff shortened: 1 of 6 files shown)


api/routes.go
```diff
@@ -41,32 +41,32 @@ func (api *API) buildHTTPRoutes() {
...
 
 	// Endpoints for user API keys.
 	api.staticRouter.POST("/user/apikeys", api.WithDBSession(api.withAuth(api.userAPIKeyPOST)))
-	api.staticRouter.GET("/user/apikeys", api.WithDBSession(api.withAuth(api.userAPIKeyGET)))
-	api.staticRouter.DELETE("/user/apikeys/:id", api.WithDBSession(api.withAuth(api.userAPIKeyDELETE)))
+	api.staticRouter.GET("/user/apikeys", api.withAuth(api.userAPIKeyGET))
+	api.staticRouter.DELETE("/user/apikeys/:id", api.withAuth(api.userAPIKeyDELETE))
 
 	// Endpoints for email communication with the user.
 	api.staticRouter.GET("/user/confirm", api.WithDBSession(api.noAuth(api.userConfirmGET))) // TODO POST
```
peterjan, line 72: still needs a session?

api/routes.go
```diff
@@ -41,32 +41,32 @@ func (api *API) buildHTTPRoutes() {
 
 	api.staticRouter.GET("/login", api.WithDBSession(api.noAuth(api.loginGET)))
 	api.staticRouter.POST("/login", api.WithDBSession(api.noAuth(api.loginPOST)))
-	api.staticRouter.POST("/logout", api.WithDBSession(api.withAuth(api.logoutPOST)))
+	api.staticRouter.POST("/logout", api.withAuth(api.logoutPOST))
 	api.staticRouter.GET("/register", api.WithDBSession(api.noAuth(api.registerGET)))
```
peterjan, line 45: still needs a session?

api/stripe.go
```diff
@@ -226,19 +225,19 @@ func (api *API) processStripeSub(ctx context.Context, s *stripe.Subscription) er
 		Prorate:    &True,
 	}
 	for _, subsc := range subs {
-		if subsc == nil || (mostRecentSub != nil && subsc.ID == mostRecentSub.ID) {
+		if subsc == nil || subsc.ID == "" || (mostRecentSub != nil && subsc.ID == mostRecentSub.ID) {
```
peterjan, line 228: reminder to remove

database/user.go
```diff
@@ -349,6 +349,7 @@ func (db *DB) UserCreate(ctx context.Context, emailAddr, pass, sub string, tier
 		Sub:                              sub,
 		Tier:                             tier,
 	}
+	// TODO This part can race and create multiple accounts with the same email, unless we add DB-level uniqueness restriction.
```
peterjan, line 352: we should definitely have a unique index on the user's email field, do they all have an email? can only do that if they do obviously, that's probably why it's not added yet?

database/user.go
```diff
@@ -455,6 +456,14 @@ func (db *DB) UserDelete(ctx context.Context, u *User) error {
 	if err != nil {
 		return errors.AddContext(err, "failed to delete user registry writes")
 	}
+	_, err = db.staticAPIKeys.DeleteMany(ctx, filter)
```
peterjan, line 459: why is this not a composed error? we want to try and delete as much as possible right? no reason to return on error but instead compose alle errors?

api/routes.go
```diff
@@ -41,32 +41,32 @@ func (api *API) buildHTTPRoutes() {
 
 	api.staticRouter.GET("/login", api.WithDBSession(api.noAuth(api.loginGET)))
```
peterjan, line 42: still needs a session?

peterjan: so I guess for these routes the reason you still need a session is because of the challenge right?
just wondering whether it's an issue you have more than one challenge?

ro-tex: There is a theoretical situation in which we delete the user and then try to create a challenge for them. That's what makes the session necessary. Not that it's going to have much of an effect, aside from leaving junk in the DB.

api/stripe.go
```diff
@@ -182,17 +182,16 @@ func readStripeEvent(w http.ResponseWriter, req *http.Request) (*stripe.Event, i
 // adjusts the user's record accordingly.
 func (api *API) processStripeSub(ctx context.Context, s *stripe.Subscription) error {
 	api.staticLogger.Traceln("Processing subscription:", s.ID)
+	u, err := api.staticDB.UserByStripeID(ctx, s.Customer.ID)
+	if err != nil {
+		return errors.AddContext(err, "failed to fetch user from DB based on subscription info")
```
peterjan, line 187: `nit` could improve the error here and say "no user found from stripe customer with ID '%v'", would be annoying to find that error without a mention of the stripe customer :)

database/user.go
peterjan, line 352: kind of a scary TODO to leave in here like that, can't we quickly add a unique index instead?

api/routes.go
ro-tex: Hm, we can remove this one.

api/routes.go
peterjan: ok, this also means a login might block a delete though right, feel free to resolve it was just a question

database/user.go
ro-tex: That's one way to do it. We delete the user last here, so the client can retry the deletion until they succeed.

I would be fine with "delete the user first and then as much as possible".

database/user.go
ro-tex: We can. We removed it when we added the challenge/response because users wanted the email to be optional and the challenge/response doesn't provide an email. So, accounts registered with that will have empty emails, which won't be possible if they are unique

ro-tex: It was there, we removed it. See below.

database/user.go
ChrisSchinnerl: I think the f/u would be to mark the user for deletion and then retry indefinitely until all of their data is gone. Finally we remove the "mark".

We wouldn't want to have partial data stored.

ChrisSchinnerl (approved):

database/user.go
ro-tex: F/U: https://github.com/SkynetLabs/skynet-accounts/issues/136

peterjan (approved):

PR #140, opened 2022-02-25
Public API Keys

ro-tex:
# PULL REQUEST

## Overview

Public API keys are a special type of API keys which are meant to be shared with the public. Their validity is very limited - they only allow users to download skylinks and even that is limited to a pre-determined list of skylinks.

The main questions in this PR are:
* Should we store the public API keys together with the regular ones or not? From my perspective, they are sufficiently different to require separate storage and handling.
* Are we happy with the authentication flow? It seems a bit too complicated to me, maybe there are ways to simplify it.

Currently open TODOs:
 - handlers need to updated to support public api keys
 - we need integration tests 
 - we need tests for everything new in `auth.go`

## Checklist
 - [ ] Changelog file created


api/cache.go
```diff
@@ -46,8 +46,10 @@ func (utc *userTierCache) Get(sub string) (int, bool) {
 	return ce.Tier, true
 }
 
-// Set stores the user's tier in the cache.
-func (utc *userTierCache) Set(u *database.User) {
+// Set stores the user's tier in the cache. If the customCacheKey is not empty,
+// it will be used to store the user in the cache, otherwise the user's sub will
+// be used.
+func (utc *userTierCache) Set(u *database.User, customCacheKey string) {
```
ChrisSchinnerl, line 52: This signature is a bit strange. I don't like the idea of being able to pass an optional value type and the caller being the one to figure out whether to pass `""` or not. So I'd probably just be explicit about it and make the second argument a `key` that's always either set to `u.Sub` or `akStr`.

I would also change the order of the arguments. Usually the key comes first followed by the value.

database/publicapikeys.go
```diff
@@ -0,0 +1,197 @@
...
+	"gitlab.com/NebulousLabs/errors"
+	"gitlab.com/NebulousLabs/fastrand"
+	"go.mongodb.org/mongo-driver/bson"
+	"go.mongodb.org/mongo-driver/bson/primitive"
+	"go.mongodb.org/mongo-driver/mongo"
+	"go.mongodb.org/mongo-driver/mongo/options"
+)
+
+type (
+	// TODO: I am still not sure whether we should use separate collections or
```
ChrisSchinnerl, line 17: Tbh after reviewing the PR it seems like using a separate collection results in a lot of blowup. e.g. double the amount of deletion, fetching etc. methods. More endpoints and so on.

When actually a public api key is just a normal api key with a skylink whitelist. So I think we could safe a lot of code if we just treated them the same as regular api keys but if the skylinks field is not empty we restrict access to the specified links.

ro-tex: Right, this is one of the big questions here. I'm not completely sure about this because public and private API keys need to be treated in different ways in many situations.

Examples:
1. When listing/getting API keys, private API keys must not be displayed, while public ones should be displayed in their entirety.
2. Editing private API keys is not possible.

Otherwise, keeping both types of API keys in the same collection will make it a little more straightforward to validate them.

Also, we will need a special flag, differentiating between public and private API keys - the list of skylinks is not enough. A user can `PUT`/`PATCH` a public API key and remove all of the skylinks it enables and that shouldn't turn the public API key into a private one. This is a minor details but it's relevant to your comment, so I'm mentioning it right away.

api/apikeys.go
```diff
@@ -3,16 +3,91 @@ package api
...
 
+type (
+	// APIKeyPOST describes the body of a POST request that creates an API key
+	APIKeyPOST struct {
+		Public   bool     `json:"public,string"`
+		Skylinks []string `json:"skylinks"`
+	}
+	// APIKeyPUT describes the request body for updating an API key
+	APIKeyPUT struct {
+		Skylinks []string
```
peterjan, line 23: no `json` annotation here?

api/apikeys.go
```diff
@@ -3,16 +3,91 @@ package api
...
+	}
+	// APIKeyPATCH describes the request body for updating an API key by
+	// providing only the requested changes
+	APIKeyPATCH struct {
+		Add    []string
```
peterjan, line 28: no `json` annotation here?

api/apikeys.go
```diff
@@ -3,16 +3,91 @@ package api
...
+	// also reveals the value of the Key field. This should only be used on key
+	// creation.
+	APIKeyResponseWithKey struct {
+		APIKeyResponse
+		Key database.APIKey `json:"key"`
+	}
+)
+
+// Valid checks if the request and its parts are valid.
+func (akp APIKeyPOST) Valid() bool {
```
peterjan, line 50: so this is nice but I think we might benefit from an error providing some context here,
because of the fact that `akp.Skylinks` can be huge and it will be annoying for the
caller to find out which one is giving him issues

api/handlers.go
```diff
@@ -27,6 +27,9 @@ const (
 	// LimitBodySizeSmall defines a size limit for requests that we don't expect
 	// to contain a lot of data.
 	LimitBodySizeSmall = 4 * skynet.KiB
+	// LimitBodySizeLarge defines a size limit for requests that we expect to
+	// contain a lot of data.
+	LimitBodySizeLarge = 4 * skynet.MiB
```
peterjan, line 32: this will put a limit on the amount of skylinks covered by an API key right?
we should probably put an actual predefined limit on that and then measure it and update this value?

api/apikeys.go
ro-tex: No need - we are not doing anything custom, so the standard `skylinks` representation is exactly what we want.

api/handlers.go
```diff
@@ -447,25 +463,73 @@ func (api *API) userLimitsGET(_ *database.User, w http.ResponseWriter, req *http
...
+// the given skylink. This method exists to accommodate public API keys.
+//
+// NOTE: This handler needs to use the noAuth middleware in order to be able to
+// optimise its calls to the DB and the use of caching.
+func (api *API) userLimitsSkylinkGET(u *database.User, w http.ResponseWriter, req *http.Request, ps httprouter.Params) {
+	respAnon := userLimitsGetFromTier(database.TierAnonymous, false)
+	// Validate the skylink.
+	skylink := ps.ByName("skylink")
+	if !database.ValidSkylinkHash(skylink) {
+		api.staticLogger.Tracef("Invalid skylink: %s", skylink)
```
peterjan, line 487: `nit` 
```suggestion
		api.staticLogger.Tracef("Invalid skylink: '%s'", skylink)
```

api/handlers.go
```diff
@@ -447,25 +463,73 @@ func (api *API) userLimitsGET(_ *database.User, w http.ResponseWriter, req *http
...
+		api.userLimitsGET(u, w, req, ps)
+		return
+	}
+	if err != nil {
+		api.staticLogger.Debugf("Error while processing API key: %s", err)
+		api.WriteJSON(w, respAnon)
+		return
+	}
+	// Check the cache before hitting the database.
+	tier, qe, ok := api.staticUserTierCache.Get(ak.String() + skylink)
```
peterjan, line 505: `nit` ok -> cached

api/handlers.go
```diff
@@ -447,25 +463,73 @@ func (api *API) userLimitsGET(_ *database.User, w http.ResponseWriter, req *http
...
+	}
+	// Get the owner of this API key from the database.
+	user, err := api.staticDB.UserByID(req.Context(), akr.UserID)
+	if err != nil {
+		api.staticLogger.Tracef("Failed to get user for user ID: %v", err)
+		api.WriteJSON(w, respAnon)
+		return
+	}
+	// Store the user in the cache with a custom key.
+	api.staticUserTierCache.Set(ak.String()+skylink, user)
```
peterjan, line 531: so this is a guess but usually with caches the cache invalidation is where the bugs are, do we invalidate this cache when the user gets updated?

database/apikeys.go
```diff
@@ -84,8 +105,37 @@ func (ak *APIKey) LoadBytes(b []byte) error {
...
+	return string(ak)
+}
+
+// CoversSkylink tells us whether a given API key covers a given skylink.
+// Private API keys cover all skylinks while public ones - only a limited set.
+func (akr APIKeyRecord) CoversSkylink(sl string) bool {
+	if !akr.Public {
+		return true
+	}
+	for _, s := range akr.Skylinks {
```
peterjan, line 129: what's the limit on `akr.Skylinks` here, this might seem like something we shouldn't consider but if these API keys are going to be evaluated before every single download, shouldn't we consider optimizing this search here?

peterjan: not sure myself to be honest, just thinking out loud, would be nice to find a fast way to tell whether a skylink is covered or not

database/apikeys.go
```diff
@@ -150,3 +231,86 @@ func (db *DB) APIKeyList(ctx context.Context, user User) ([]APIKeyRecord, error)
...
+// APIKeyUpdate updates an existing API key. This works by replacing the
+// list of Skylinks within the API key record. Only valid for public API keys.
+func (db *DB) APIKeyUpdate(ctx context.Context, user User, akID primitive.ObjectID, skylinks []string) error {
+	if user.ID.IsZero() {
+		return errors.New("invalid user")
+	}
+	// Validate all given skylinks.
+	for _, s := range skylinks {
+		if !ValidSkylinkHash(s) {
+			return ErrInvalidSkylink
```
peterjan, line 244: maybe add the skylink hash as context?

database/apikeys.go
```diff
@@ -150,3 +231,86 @@ func (db *DB) APIKeyList(ctx context.Context, user User) ([]APIKeyRecord, error)
...
+		}
+	}
+	filter := bson.M{
+		"_id":     akID,
+		"public":  &True, // you can only update public API keys
```
peterjan, line 249: I never really understood the `&True` stuff, why don't you do `{$eq: true}`

database/apikeys.go
```diff
@@ -150,3 +231,86 @@ func (db *DB) APIKeyList(ctx context.Context, user User) ([]APIKeyRecord, error)
...
+			return ErrInvalidSkylink
+		}
+	}
+	filter := bson.M{
+		"_id":    akID,
+		"public": &True, // you can only update public API keys
+	}
+	var update bson.M
+	// First, all new skylinks to the record.
+	if len(addSkylinks) > 0 {
```
peterjan, line 284: this is not great, I think you want to dedupe this first and then also you want to use `$addToSet` instead of `$push`

database/apikeys.go
```diff
@@ -150,3 +231,86 @@ func (db *DB) APIKeyList(ctx context.Context, user User) ([]APIKeyRecord, error)
...
+		ur, err := db.staticAPIKeys.UpdateOne(ctx, filter, update, &opts)
+		if err != nil {
+			return err
+		}
+		if ur.ModifiedCount == 0 {
+			return errors.New("public API key not found, no keys updated")
+		}
+	}
+	// Then, remove all skylinks that need to be removed.
+	if len(removeSkylinks) > 0 {
```
peterjan, line 300: `probably want to dedupe this first`

database/user.go
```diff
@@ -48,6 +48,8 @@ const (
 var (
 	// True is a helper for when we need to pass a *bool to MongoDB.
 	True = true
+	// False is a helper for when we need to pass a *bool to MongoDB.
+	False = false
```
peterjan, line 52: why is this needed again? can't you just type `false` whenever you need it in mongo queries?

test/api/apikeys_test.go
```diff
@@ -162,3 +134,155 @@ func testAPIKeysUsage(t *testing.T, at *test.AccountsTester) {
 			uploadSize, us.TotalDownloadsSize, us.NumUploads, skynet.BandwidthUploadCost(uploadSize), us.BandwidthUploads)
 	}
 }
+
+// TestPublicAPIKeyFlow validates the creation, listing, and deletion of public
```
peterjan, line 138: update docstring

test/database/apikeys_test.go
```diff
@@ -0,0 +1,151 @@
...
+	"github.com/SkynetLabs/skynet-accounts/test"
+)
+
+// TestAPIKeys ensures the DB operations with API keys work as expected.
+func TestAPIKeys(t *testing.T) {
+	t.Parallel()
+
+	ctx := context.Background()
+	dbName := test.DBNameForTest(t.Name())
+	db, err := database.NewCustomDB(ctx, dbName, test.DBTestCredentials(), nil)
```
peterjan, line 17: I like your `test` package. I think it could be better though if you'd make it `test.NewDatabase()` and then also just handle the `dbName` thing internally, so it'd be `test.NewDatabase(ctx, t.Name(), nil)` that would be pretty epic and then I'll copy it 😛

test/utils.go
```diff
@@ -34,8 +35,28 @@ type (
 		*database.User
 		staticDB *database.DB
 	}
+	// ResponseWriter is a testing ResponseWriter implementation.
+	ResponseWriter struct {
```
peterjan, line 39: you should check out `httptest.NewRecorder()` maybe it's something you like and can adopt in one of your repos

peterjan (changes requested):
looks good overall, just requesting changes because I have couple of suggestions

api/apikeys.go
ro-tex: The defaults is what we want here.

api/apikeys.go
```diff
@@ -3,16 +3,91 @@ package api
...
+	for _, s := range akp.Skylinks {
+		if !database.ValidSkylinkHash(s) {
+			return false
+		}
+	}
+	return true
+}
+
+// FromAPIKey populates the struct's fields from the given API key.
+func (rwk *APIKeyResponse) FromAPIKey(ak database.APIKeyRecord) {
```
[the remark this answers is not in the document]
ro-tex: sure

api/apikeys.go
```diff
@@ -22,32 +97,57 @@ func (api *API) userAPIKeyPOST(u *database.User, w http.ResponseWriter, req *htt
...
-		*ak,
-		ak.Key,
+	var resp APIKeyResponseWithKey
+	resp.FromAPIKey(*ak)
+	api.WriteJSON(w, resp)
+}
+
+// userAPIKeyGET returns a single API key.
+func (api *API) userAPIKeyGET(u *database.User, w http.ResponseWriter, req *http.Request, ps httprouter.Params) {
+	akID, err := primitive.ObjectIDFromHex(ps.ByName("id"))
```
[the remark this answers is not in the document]
ro-tex: If `id` is not set the request won't be routed to `/user/apikeys/:id` at all, so we won't hit this handler. `id` might be invalid but we can rely on it being there. If it's not there the user will hit `/user/apikeys` and receive a list of all API keys but at that point it's their own fault.

api/apikeys.go
```diff
@@ -22,32 +97,57 @@ func (api *API) userAPIKeyPOST(u *database.User, w http.ResponseWriter, req *htt
...
+	if err != nil {
+		api.WriteError(w, err, http.StatusBadRequest)
+		return
+	}
+	ak, err := api.staticDB.APIKeyGet(req.Context(), akID)
```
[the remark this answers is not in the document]
ro-tex: I appreciate the suggestion but I prefer to have the visibility of an explicit error when we don't find any documents.
I agree that using a DB-level error in the handler is not ideal but we're far away from that kind of nice polish. At some point we'll gradually get there.

api/apikeys.go
```diff
@@ -58,3 +158,47 @@ func (api *API) userAPIKeyDELETE(u *database.User, w http.ResponseWriter, req *h
...
+		api.WriteError(w, err, http.StatusBadRequest)
+		return
+	}
+	var body APIKeyPUT
+	err = parseRequestBodyJSON(req.Body, LimitBodySizeLarge, &body)
+	if err != nil {
+		api.WriteError(w, err, http.StatusBadRequest)
+		return
+	}
+	err = api.staticDB.APIKeyUpdate(req.Context(), *u, akID, body.Skylinks)
```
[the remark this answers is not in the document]
ro-tex: Because the method doesn't take a pointer.

The general rule is that you pass a reference when:
a) you want to modify the value
b) the value is nested and working with a copy might create concurrency issues (e.g. copying mutexes, etc.)
c) the value is large and copying it will incur a performance hit

Neither of those is present here.

api/apikeys.go
```diff
@@ -58,3 +158,47 @@ func (api *API) userAPIKeyDELETE(u *database.User, w http.ResponseWriter, req *h
...
+		return
+	}
+	api.WriteSuccess(w)
+}
+
+// userAPIKeyPATCH patches an API key. The difference between PUT and PATCH is
+// that PATCH only specifies the changes while PUT provides the expected list of
+// covered skylinks. Only possible for public API keys.
+func (api *API) userAPIKeyPATCH(u *database.User, w http.ResponseWriter, req *http.Request, ps httprouter.Params) {
+	akID, err := primitive.ObjectIDFromHex(ps.ByName("id"))
```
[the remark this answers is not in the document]
ro-tex: same answer...

it might be better to start referring to all occurrences of an issue in the original comment, rather than commenting "same" - it sometimes get confusing.

api/auth.go
```diff
@@ -0,0 +1,105 @@
...
+	if err != nil {
+		return nil, nil, err
+	}
+	akr, err := api.staticDB.APIKeyByKey(req.Context(), ak.String())
+	if err != nil {
+		return nil, nil, err
+	}
+	// If we're dealing with a public API key, we need to validate that this
+	// request is a GET for a covered skylink.
+	if akr.Public {
```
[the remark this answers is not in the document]
ro-tex: Why would there be? This is called from `withAuth` which is a middleware we use with all types of requests.

api/handlers.go
ro-tex: sure, we can do that. for the moments I don't see a particular need for it - 4MB is a lot of data. if we see a use case in which a client wants to give public access to that many skylinks we can specify/change the limit. the main reason this is here is to prevent a malicious actor from running us OOM.

api/handlers.go
ro-tex: I prefer to keep the `ok`, if that's fine with you.

database/apikeys.go
ro-tex: Do I see a real-life use case for bloom filters?! I would love to implement that.

database/apikeys.go
ro-tex: sure

database/apikeys.go
ro-tex: Believe me, I would love to. But the types used in MongoDB's go driver define the field as `*bool`, so I need to define a variable and give it a pointer to that. If you know of a nice way to avoid that, I would be happy to hear it.

test/utils.go
ro-tex: Sounds good, I'll take a look!

F/U: https://github.com/SkynetLabs/skynet-accounts/pull/152

api/apikeys.go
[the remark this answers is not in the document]
peterjan: fair enough

api/apikeys.go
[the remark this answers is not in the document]
peterjan: > Because the method doesn't take a pointer.

this made me laugh 😁

database/apikeys.go
peterjan: this is true but only in the case of option structs I think, in filters you can do `bson.M{"$eq": true}`

database/apikeys.go
peterjan: yeah  feel free to resolve, the `$addToSet` was the important one, the reason I would dedupe is that you just make it easier on your database engine 🤷‍♂️

api/apikeys.go
```diff
@@ -3,16 +3,107 @@ package api
...
+		Key database.APIKey `json:"key"`
+	}
+)
+
+// Validate checks if the request and its parts are valid.
+func (akp APIKeyPOST) Validate() error {
+	if !akp.Public && len(akp.Skylinks) > 0 {
+		return errors.New("public API keys cannot refer to skylinlks")
+	}
+	errs := make([]error, 0)
```
peterjan, line 54:
```suggestion
	var errs []error
```

api/apikeys.go
```diff
@@ -3,16 +3,107 @@ package api
...
+	for _, s := range akp.Skylinks {
+		if !database.ValidSkylinkHash(s) {
+			errs = append(errs, errors.New("invalid skylink:"+s))
```
peterjan, line 57:
```suggestion
			errs = append(errs, errors.New("invalid skylink: "+s))
```
not sure whether the concat operator adds a space by default there?

api/apikeys.go
```diff
@@ -22,32 +113,51 @@ func (api *API) userAPIKeyPOST(u *database.User, w http.ResponseWriter, req *htt
...
-	akID := ps.ByName("id")
-	err := api.staticDB.APIKeyDelete(req.Context(), *u, akID)
+	akID, err := primitive.ObjectIDFromHex(ps.ByName("id"))
+	if err != nil {
+		api.WriteError(w, err, http.StatusBadRequest)
+		return
+	}
+	err = api.staticDB.APIKeyDelete(req.Context(), *u, akID)
 	if err == mongo.ErrNoDocuments {
 		api.WriteError(w, err, http.StatusBadRequest)
```
peterjan, line 162: not found?

api/apikeys.go
```diff
@@ -58,3 +168,47 @@ func (api *API) userAPIKeyDELETE(u *database.User, w http.ResponseWriter, req *h
...
+	}
+	var body APIKeyPUT
+	err = parseRequestBodyJSON(req.Body, LimitBodySizeLarge, &body)
+	if err != nil {
+		api.WriteError(w, err, http.StatusBadRequest)
+		return
+	}
+	err = api.staticDB.APIKeyUpdate(req.Context(), *u, akID, body.Skylinks)
+	if err != nil {
+		api.WriteError(w, err, http.StatusInternalServerError)
```
peterjan, line 187: there's no `NotFound` on a `PUT`?

api/apikeys.go
```diff
@@ -58,3 +168,47 @@ func (api *API) userAPIKeyDELETE(u *database.User, w http.ResponseWriter, req *h
...
+		api.WriteError(w, err, http.StatusBadRequest)
+		return
+	}
+	var body APIKeyPATCH
+	err = parseRequestBodyJSON(req.Body, LimitBodySizeLarge, &body)
+	if err != nil {
+		api.WriteError(w, err, http.StatusBadRequest)
+		return
+	}
+	err = api.staticDB.APIKeyPatch(req.Context(), *u, akID, body.Add, body.Remove)
```
peterjan, line 208: there's no NotFound on a `PATCH`?

api/auth.go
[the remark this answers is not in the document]
peterjan: yeah just wanted to make sure there's no `PUT`, `POST`, `PATCH` or `DELETE` routes that have a skylink hash in the request URI, but you can resolve if it's all fine, just double-checking because you explicitly mention a `GET` in the comment and then not check `req.Method`

peterjan (dismissed):
looks good to me!

api/auth.go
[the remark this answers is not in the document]
ro-tex: Oh, I see. Yes, we can check that here and exit early when the request is not a GET, as that is guaranteed to fail later.
I'll do it in a F/U. 👍

api/apikeys.go
ro-tex: No, it doesn't. `fmt.Println(s1, s2)` adds a space.

api/apikeys.go
ro-tex: Good point. Added.

api/apikeys.go
ro-tex: added

api/auth.go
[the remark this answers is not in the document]
ro-tex: Actually, I just did it here because I had to make some changes anyway.

database/apikeys.go
ro-tex: Oh, you are right. Inertia, I guess. :)

ChrisSchinnerl (approved):

peterjan (approved):

PR #141, opened 2022-02-28
Fix /user/limits reporting wrong tier

ro-tex:
# PULL REQUEST

## Overview

When reporting the user's tier, we check whether their subscription has expired and, if it has, we report `free`, regardless of their tier. This causes issues with users who are promoted manually, without having a running subscription.

The solution is to not do this manual check and rely on the Stripe webhook to take care of subscription levels.

## Checklist
 - [ ] Changelog file created

### DIFF
--- a/api/cache.go
+++ b/api/cache.go
@@ -50,12 +50,7 @@ func (utc *userTierCache) Get(sub string) (int, bool) {
 func (utc *userTierCache) Set(u *database.User) {
 	var ce userTierCacheEntry
 	now := time.Now().UTC()
-	if u.SubscribedUntil.Before(now) {
-		ce = userTierCacheEntry{
-			Tier:      database.TierFree,
-			ExpiresAt: now.Add(userTierCacheTTL),
-		}
-	} else if u.QuotaExceeded {
+	if u.QuotaExceeded {
 		ce = userTierCacheEntry{
 			Tier:      database.TierAnonymous,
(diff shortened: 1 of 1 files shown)


ChrisSchinnerl (approved):

api/cache.go
```diff
@@ -50,12 +50,7 @@ func (utc *userTierCache) Get(sub string) (int, bool) {
 func (utc *userTierCache) Set(u *database.User) {
 	var ce userTierCacheEntry
 	now := time.Now().UTC()
-	if u.SubscribedUntil.Before(now) {
-		ce = userTierCacheEntry{
-			Tier:      database.TierFree,
-			ExpiresAt: now.Add(userTierCacheTTL),
-		}
-	} else if u.QuotaExceeded {
+	if u.QuotaExceeded {
```
peterjan, line 53: are you sure we are correctly handling the webhooks? that happens in `processStripeSub`, I might be reading that wrong but I'm not 100% convinced that we appropriately update the user's tier based on the subscription start and end period. Do you mind verifying that? If it's fine then I'll approve but this might introduce more issues than it's causing and it's a dangerous PR because we're relying on something that's not in the diff

ro-tex: I believe it works as expected. It hasn't been touched for months and I haven't heard of any indications that it's not working as expected. Also, if that handling is buggy, then the subscription expiry date in the user's record will also be off because it's set by the same function.

PR #150, opened 2022-03-10
/user/limits in bits or bytes

ro-tex:
# PULL REQUEST

## Overview

The `/limits` endpoint returns the bandwidth limits in bits per second, which is the format most people expect to see.
The `/user/limits` endpoint, though, returns the bandwidth in bytes per second because that is what Nginx expects to see.

This discrepancy is very confusing for users and it needs to be resolved. To that end this PR adds a new query parameter which would let Nginx request the limits in bytes (see https://github.com/SkynetLabs/skynet-webportal/pull/1855) while everyone else will receive the limits in bits.

## Checklist
 - [ ] Changelog file created

### DIFF
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -9,6 +9,7 @@ import (
 	"net/mail"
 	"net/url"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/SkynetLabs/skynet-accounts/build"
@@ -407,7 +408,7 @@ func (api *API) userLimitsGET(_ *database.User, w http.ResponseWriter, req *http
 	// inBytes is a flag indicating that the caller wants all bandwidth limits
 	// to be presented in bytes per second. The default behaviour is to present
 	// them in bits per second.
-	inBytes := req.FormValue("unit") == "byte"
+	inBytes := strings.EqualFold(req.FormValue("unit"), "byte")
 	// First check for an API key.
 	ak, err := apiKeyFromRequest(req)
 	respAnon := userLimitsGetFromTier(database.TierAnonymous, inBytes)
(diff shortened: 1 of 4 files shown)


api/handlers.go
```diff
@@ -402,22 +404,26 @@ func (api *API) userGET(u *database.User, w http.ResponseWriter, _ *http.Request
 // NOTE: This handler needs to use the noAuth middleware in order to be able to
 // optimise its calls to the DB and the use of caching.
 func (api *API) userLimitsGET(_ *database.User, w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
+	// inBytes is a flag indicating that the caller wants all bandwidth limits
+	// to be presented in bytes per second. The default behaviour is to present
+	// them in bits per second.
+	inBytes := req.FormValue("unit") == "byte"
```
ChrisSchinnerl, line 410: we should have a test for this flag

ChrisSchinnerl (approved):

test/tester.go
```diff
@@ -312,8 +313,12 @@ func (at *AccountsTester) TrackRegistryWrite() (int, error) {
 }
 
 // UserLimits performs a `GET /user/limits` request.
-func (at *AccountsTester) UserLimits() (api.UserLimitsGET, int, error) {
-	r, b, err := at.request(http.MethodGet, "/user/limits", nil, nil)
+func (at *AccountsTester) UserLimits(unit string) (api.UserLimitsGET, int, error) {
+	queryParams := url.Values{}
+	if unit != "" {
```
peterjan, line 318: `nit` redundant

peterjan (approved):

ro-tex:
DO NOT MERGE!

Nginx is not passing the right query parameters when asking for user limits, so if we merge this right now it will get the answer in bits and interpret it in bytes.

Blocker: https://github.com/SkynetLabs/skynet-webportal/pull/1855

PR #152, opened 2022-03-15
Refactor the Tester package

ro-tex:
# PULL REQUEST

## Overview

The Tester package has room for improvement and PJ had some good input:
- [x] cleaner test database creation: https://github.com/SkynetLabs/skynet-accounts/pull/140#discussion_r824755853
- [ ] cleaner http request testing: https://github.com/SkynetLabs/skynet-accounts/pull/140#discussion_r824757825 (https://pkg.go.dev/net/http/httptest)

## Checklist
 - [ ] Changelog file created


test/tester_test.go
```diff
@@ -0,0 +1,30 @@
...
+			name:     "long name",
+			input:    "this_is_a_very_long_name_and_it_will_be_trimmed_to_sixty_four_characters",
+			expected: "this_is_a_very_long_name_and_it_will_be_trimmed_to_sixty_four_characters"[:64],
+		},
+	}
+
+	for _, tt := range tests {
+		out := CleanName(tt.input)
+		if out != tt.expected {
+			t.Errorf("Test '%s' failed. Expected '%s', got '%s'", tt.name, tt.expected, out)
```
ChrisSchinnerl, line 27: If you call `t.Run(tt.name, func(t *testing.T){ // ... })` here it's a bit nicer. Then you can also execute the individual subtests separately.

test/api/challenge_test.go
```diff
@@ -24,73 +22,51 @@ func testRegistration(t *testing.T, at *test.AccountsTester) {
...
-		err = json.Unmarshal(b, &ch)
-		if err != nil {
-			t.Fatal("Failed to get a challenge:", err)
-		}
-		chBytes, err = hex.DecodeString(ch.Challenge)
-		if err != nil {
-			t.Fatal("Invalid challenge:", err)
-		}
+	ch, status, err := at.RegisterGET(pk[:])
+	chBytes, err := hex.DecodeString(ch.Challenge)
```
ChrisSchinnerl, line 40: Might be able to avoid this by changing the field's type to a byte slice and implementing the json interfaces to marshal it as a hex.

test/tester.go
```diff
@@ -319,17 +288,244 @@ func (at *AccountsTester) HealthGet() (api.HealthGET, int, error) {
...
+	return at.post("/login", nil, params)
+}
+
+// LoginPubKeyGET performs `GET /login`
+func (at *AccountsTester) LoginPubKeyGET(pk database.PubKey) (api.ChallengePublic, int, error) {
+	params := url.Values{}
+	if pk != nil {
+		params.Set("pubKey", hex.EncodeToString(pk[:]))
+	}
+	r, b, err := at.Request(http.MethodGet, "/login", params, nil, nil)
```
ChrisSchinnerl, line 309: If you feel fancy you can add wrappers for post, get and delete to avoid the `http.MethodGet`, `http.MethodPost` etc.

test/tester.go
```diff
@@ -319,17 +288,244 @@ func (at *AccountsTester) HealthGet() (api.HealthGET, int, error) {
...
+	err = json.Unmarshal(b, &resp)
+	if err != nil {
+		return api.ChallengePublic{}, 0, errors.AddContext(err, "failed to marshal the body JSON")
+	}
+	return resp, r.StatusCode, nil
+}
+
+// RegisterPOST performs `POST /register`
+func (at *AccountsTester) RegisterPOST(response, signature []byte, email string) (api.UserGET, int, error) {
+	// bb, err := json.Marshal(database.ChallengeResponse{
```
ChrisSchinnerl, line 363: commented out code

test/api/challenge_test.go
ro-tex: This is out of scope for the current PR. All we're doing here is refactoring the `tester` package.

test/tester.go
ro-tex: I don't think those will bring enough value at the cost of adding another level of abstraction.

I was considering hiding the entire response processing within `Request` but it seemed like too much complexity with all moving parts such as expected status codes and so on. I might put in there the body processing, though.

ChrisSchinnerl (approved):

database/challenge.go
```diff
@@ -238,28 +245,16 @@ func (cr *ChallengeResponse) LoadFromBytes(b []byte) error {
...
-	if err != nil {
-		return errors.AddContext(err, ErrInvalidChallengeResponse.Error())
-	}
-	return cr.LoadFromBytes(b)
-}
-
 // LoadString loads a PubKey from its hex-encoded string form.
 func (pk *PubKey) LoadString(s string) error {
-	bytes, err := hex.DecodeString(s)
+	bb, err := hex.DecodeString(s)
```
peterjan, line 250: not important but sometimes I have trouble trying to figure out your variable names, what does `bb` stand for?

test/api/apikeys_test.go
```diff
@@ -344,23 +344,22 @@ func testAPIKeysAcceptance(t *testing.T, at *test.AccountsTester) {
 		{verb: http.MethodPost, endpoint: "/user/reconfirm"},
 	}
 
-	var b []byte
 	for _, tt := range tests {
 		switch tt.verb {
 		case http.MethodGet:
-			r, b, err = at.Get(tt.endpoint, nil)
+			r, err = at.Request(http.MethodGet, tt.endpoint, nil, nil, nil, nil)
```
peterjan, line 350: why not just check whether `tt.verb` is allowed and then do `r, err = at.Request(tt.verb, tt.endpoint, nil, nil, nil, nil)`

database/challenge.go
ro-tex: Hm, this is actually pretty important. Both `b` and `bb` usually stand for a `[]byte`. I don't like using `bytes` because it's a package name and once you shadow it, you can't use its functions. I usually use `b`, which is our standard approach in `skyd`.

test/tester.go
```diff
@@ -42,6 +44,20 @@ type (
 	}
 )
 
+// CleanName sanitizes the input for all kinds of unwanted characters and
+// replaces those with underscores.
+// See https://docs.mongodb.com/manual/reference/limits/#naming-restrictions
+func CleanName(s string) string {
```
peterjan, line 50: `nit` prefer calling it `SanitizeX` over `CleanX`

test/database/user_test.go
```diff
@@ -596,7 +596,7 @@ func TestUserSetTier(t *testing.T) {
 func TestUserStats(t *testing.T) {
 	ctx := context.Background()
 	dbName := test.DBNameForTest(t.Name())
-	db, err := database.NewCustomDB(ctx, dbName, test.DBTestCredentials(), nil)
+	db, err := test.NewDatabase(ctx, dbName, nil)
```
peterjan, line 599: maybe we worth checking if the 3rd param here is ever used and if not remove it

test/api/apikeys_test.go
ro-tex: Oh! That's a great idea. If I use a map with the allowed verbs as keys it will be a single `if`. Thanks!

peterjan (approved):

test/api/apikeys_test.go
peterjan: well yeah or you do 
```
switch tt.verb {
    case 1:
    case 2:
    case 3:
    case 4:
         // continue
    default:
         // throw error
}
// use tt.verb
```

PR #153, opened 2022-03-16
Skylink race

ro-tex:
# PULL REQUEST

## Overview

This PR fixes a problem caused by a race condition.

The race condition happens when we try to fetch a skylink from the database, we don't find it, so we decide to insert it. When inserting it, we use `upsert` in order to avoid overwriting the ID if it just so happens that another server creates that exact skylink in that very moment. The issue is that in that case the upsert operation does not do anything and it does not return an `UpsertID`. This leaves the skylink record returned by the function with a valid hash but an empty ID. This empty ID later on causes this skylink to be evaluated as invalid.

This PR makes two changes:
 - instead of "invalid skylink" we now return "skylink doesn't exist" to better capture the situation of an empty ID
 - ~~we specifically check whether the `UpsertID` we get is empty and if that's the case we fetch the skylink record from the database~~
 - we replace the `find and if it's not there upsert` flow with a single call to `findOneAndUpdate` that will only perform an update if if doesn't find the record in the DB (thanks to `$setOnInsert`), thus giving us an atomic way to either find or insert the record.

## Checklist
 - [ ] Changelog file created

### DIFF
--- a/database/skylink.go
+++ b/database/skylink.go
@@ -8,7 +8,6 @@ import (
 	"gitlab.com/SkynetLabs/skyd/skymodules"
 	"go.mongodb.org/mongo-driver/bson"
 	"go.mongodb.org/mongo-driver/bson/primitive"
-	"go.mongodb.org/mongo-driver/mongo"
 	"go.mongodb.org/mongo-driver/mongo/options"
 )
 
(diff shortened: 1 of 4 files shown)


database/skylink.go
```diff
@@ -65,10 +65,17 @@ func (db *DB) Skylink(ctx context.Context, skylink string) (*Skylink, error) {
...
 		ur, err = db.staticSkylinks.UpdateOne(ctx, filter, upsert, opts)
+		if err != nil {
+			return nil, err
+		}
 		// The UpsertedID might be nil in case the skylink got added to the DB
-		// by another server in between the calls.
-		if err == nil && ur.UpsertedID != nil {
+		// by another server in between the calls. In that case we'll fetch the
+		// skylink record from the DB.
+		if ur.UpsertedID != nil {
```
ChrisSchinnerl, line 74: Seems like this could have been avoided by using `db.staticSkylinks.FindOneAndUpdate` with upsert?
I think that would also clean the code up. 

Otherwise we'd need to add a regression test which tests the race. This way all we need to do is make sure that it's upserting correctly which I assume is tested already.

ChrisSchinnerl (changes requested):

ChrisSchinnerl (approved):

peterjan (approved):
👍

PR #159, opened 2022-03-16
Implement `DELETE /user/pubkey/:pubKey`

### LINKED ISSUE #147: DELETE /user/pubkey/:pubkey
## Overview

We need an endpoint for removing a pubkey from an account.

ro-tex:
# PULL REQUEST

## Overview

Implements `DELETE /user/pubkey/:pubKey`

## Checklist
 - [ ] Changelog file created

## Issues Closed
Closes #147


api/handlers.go
```diff
@@ -656,6 +657,43 @@ func (api *API) userPUT(u *database.User, w http.ResponseWriter, req *http.Reque
 	api.loginUser(w, u, true)
 }
 
+// userPubKeyDELETE removes a given pubkey from the list of pubkeys associated
+// with this user. It does not require a challenge-response because the used
```
m-cat, line 661: Think it should be this, unless it was meant as an homage to Stallman.

```suggestion
// with this user. It does not require a challenge-response because the user
```

test/api/challenge_test.go
```diff
@@ -309,3 +309,93 @@ func testUserAddPubKey(t *testing.T, at *test.AccountsTester) {
...
+	}
+	// Solve the challenge.
+	response := append(chBytes, append([]byte(database.ChallengeTypeUpdate), []byte(database.PortalName)...)...)
+	bodyParams := url.Values{}
+	bodyParams.Set("response", hex.EncodeToString(response))
+	bodyParams.Set("signature", hex.EncodeToString(ed25519.Sign(sk[:], response)))
+	r, b, err = at.Post("/user/pubkey/register", nil, bodyParams)
+	if err != nil {
+		t.Fatalf("Failed to confirm the update. Status %d, body '%s', error '%s'", r.StatusCode, string(b), err)
+	}
```
m-cat, line 352: nit: I wonder if you can extract this into a helper function?

m-cat (dismissed):
Couple nits.

test/api/challenge_test.go
ro-tex: I intend to refactor all `at.Post()` and `at.Get()` calls into dedicated helpers which will shrink these 20 lines into 5-6, at which stage it probably won't be worth it to move them to a helper.

api/handlers.go
```diff
@@ -656,6 +657,43 @@ func (api *API) userPUT(u *database.User, w http.ResponseWriter, req *http.Reque
...
+			keyIdx = i
+			break
+		}
+	}
+	if keyIdx == -1 {
+		build.Critical("Reaching this should be impossible. It would indicate a concurrent change of the user struct.")
+	}
+	// Remove the pubkey.
+	u.PubKeys = append(u.PubKeys[:keyIdx], u.PubKeys[keyIdx+1:]...)
+	err = api.staticDB.UserSave(ctx, u)
```
ChrisSchinnerl, line 689: It might be safer to create a dedicated pubkey deletion method which uses [$pull](https://docs.mongodb.com/manual/reference/operator/update/pull/) here. Because then you can call this method on 2 different servers with different pubkeys without running into a race condition. This way, one call might add back the pubkey of the previous one if the timing was unfortunate. 

In a perfect world (probably not worth it right now) this would be the only db operation happening when this endpoint is called. Including the authentication.

ro-tex: Oh, you're right, this is potentially very dangerous.

ro-tex: So, this is now tackled but the solution is a little incomplete.

What is done:
When we're adding pubkeys we check if the list of pubkeys is `null`. If it is, we save the entire user record in order to create the array of pubkeys. Otherwise, we use `$addToSet` or `$pull`.

The issue is that we cannot use `$addToSet` when the array is not initialised in mongo - it fails with `Cannot apply $addToSet to non-array field. Field named 'pub_keys' has non-array type null` and there doesn't seem to be a nice way around it.

I'll mitigate this in a follow up by locking the user record any time I change it. I need to do this for other reasons as well. I don't want to do it in this PR because it will blow it up.

database/user.go
```diff
@@ -477,6 +475,49 @@ func (db *DB) UserSave(ctx context.Context, u *User) error {
...
+func (db *DB) UserPubKeyRemove(ctx context.Context, u User, pk PubKey) error {
+	filter := bson.M{"_id": u.ID}
+	update := bson.M{
+		"$pull": bson.M{"pub_keys": pk},
+	}
+	opts := options.UpdateOptions{
+		Upsert: &False,
+	}
+	_, err := db.staticUsers.UpdateOne(ctx, filter, update, &opts)
+	if err != nil && strings.Contains(err.Error(), "Cannot apply $pull to a non-array value") {
```
ChrisSchinnerl, line 515: Might be more robust to update the `update` with the same condition as the code that adds a pubkey to avoid checking for this error. Because strings can change and ruin your day if they do so.

ChrisSchinnerl (approved):

database/user.go
ro-tex: Yes, this is not great. Replaced by smarted filtering.

ChrisSchinnerl (approved):

api/handlers.go
```diff
@@ -731,6 +731,34 @@ func (api *API) userPUT(u *database.User, w http.ResponseWriter, req *http.Reque
 	api.loginUser(w, u, true)
 }
 
+// userPubKeyDELETE removes a given pubkey from the list of pubkeys associated
+// with this user. It does not require a challenge-response because the used
```
peterjan, line 735:
```suggestion
// with this user. It does not require a challenge-response because the user
```

database/user.go
```diff
@@ -477,6 +474,52 @@ func (db *DB) UserSave(ctx context.Context, u *User) error {
...
+					"$cond": bson.A{
+						bson.M{"$eq": bson.A{bson.M{"$type": "$pub_keys"}, "array"}},
+						bson.M{"$setUnion": bson.A{"$pub_keys", bson.A{pk}}},
+						bson.A{pk},
+					}},
+			},
+		},
+	}
+	opts := options.UpdateOptions{
+		Upsert: &False,
```
peterjan, line 498: redundant

peterjan (approved):
lgtm 👍

database/user.go
ro-tex: F/U https://github.com/SkynetLabs/skynet-accounts/pull/170

PR #164, opened 2022-03-21
Implement names for API keys.

ro-tex:
# PULL REQUEST

## Checklist
 - [ ] Changelog file created

### DIFF
--- a/API.md
+++ b/API.md
@@ -236,6 +236,26 @@ Changes the user's password without them being logged in.
 
 ## API Keys endpoints
 
+### PATCH `/user/apikeys/:id`
+
+Updates the list of skylinks covered by a public API key.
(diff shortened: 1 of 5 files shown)


ChrisSchinnerl (approved):

peterjan (commented):
shouldn't API key names be unique per user?

ro-tex:
> shouldn't API key names be unique per user?

I see no reason to introduce such a limitation. That will mean that there can be only one API key without a name (the name is ""). Even if we add an exception for that, people might want to name their API keys according to the machine that's using them or something like this.

Overall, I don't see an unequivocal benefit of adding this limitation.

peterjan (approved):

PR #167, opened 2022-03-23
Anonymous uploads + track upload IPs

ro-tex:
# PULL REQUEST

## Overview

Track anonymous uploads and downloads.
Track the IP of any uploader.

Note:
Don't worry about the anonymous download tracking - we are going to stop tracking downloads, anyway. I just wanted to keep them in sync for the sake of my own mental model. That also helped root out a bug.

## Checklist
 - [ ] Changelog file created


test/tester.go
```diff
@@ -436,8 +436,10 @@ func (at *AccountsTester) TrackDownload(skylink string, bytes int64) (int, error
 }
 
 // TrackUpload performs a `POST /track/upload/:skylink` request.
-func (at *AccountsTester) TrackUpload(skylink string) (int, error) {
-	r, _, err := at.request(http.MethodPost, "/track/upload/"+skylink, nil, nil, nil)
+func (at *AccountsTester) TrackUpload(skylink string, ip string) (int, error) {
+	form := url.Values{}
+	form.Set("ip", ip)
+	r, _, err := at.request(http.MethodPost, "/track/upload/"+skylink, form, nil, nil)
```
ChrisSchinnerl, line 442: I might be confusing things but didn't we introduce `at.post` in another PR at some point?

database/upload.go
```diff
@@ -45,15 +46,18 @@ func (db *DB) UploadByID(ctx context.Context, id primitive.ObjectID) (*Upload, e
...
 // UploadCreate registers a new upload and counts it towards the user's used
 // storage.
-func (db *DB) UploadCreate(ctx context.Context, user User, skylink Skylink) (*Upload, error) {
-	if user.ID.IsZero() {
-		return nil, errors.New("invalid user")
+func (db *DB) UploadCreate(ctx context.Context, user *User, ip string, skylink Skylink) (*Upload, error) {
+	if user == nil {
+		// If there is no user passed, we initialise it with the zero ID in
+		// order to denote an anonymous upload.
+		user = &User{ID: primitive.ObjectID{}}
```
ChrisSchinnerl, line 53: Similar thing as below. This seems like it could hide errors.
I would feel more comfortable with specifying a dedicated and global `AnonUser = &User{ID: primitive.ObjectID{}}` which we pass explicitly.

ChrisSchinnerl (commented):
Looking pretty good. Only a few minor comments.

api/handlers.go
```diff
@@ -1302,3 +1338,27 @@ func userLimitsGetFromTier(tierID int, quotaExceeded, inBytes bool) *UserLimitsG
 		RegistryDelay:     limitsTier.RegistryDelay,
 	}
 }
+
+// validateIP is a simple pass-through helper that returns valid IPs as they are
+// and returns an empty string for invalid IPs.
+func validateIP(ip string) string {
```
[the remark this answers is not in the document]
ro-tex: I didn't know it was there and none of the SO articles I read about parsing IPs with go mentioned it.

I'll switch to it immediately, it will make this PR so much nicer. Thanks!

test/tester.go
ro-tex: I think I used to have those before but I decided that i don't need that many layers of abstraction. If they're not to be used outside of tester's implementation then it's more overhead than help.

Also, I believe Pj is using that approach in one of the abuse services.

test/utils.go
```diff
@@ -157,7 +157,10 @@ func RandomSkylink() string {
 // RegisterTestUpload registers an upload of the given skylink by the given user.
 // Returns the skylink, the upload's id and error.
 func RegisterTestUpload(ctx context.Context, db *database.DB, user *database.User, skylink *database.Skylink) (*database.Skylink, primitive.ObjectID, error) {
-	up, err := db.UploadCreate(ctx, *user, *skylink)
+	if user == nil {
```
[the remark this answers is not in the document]
ro-tex: That's fair. I guess I'm being way too lazy here and wanting to pass `null`s in my calls. I'll refactor it. The anon user constant that you proposed in another comment makes sense. I'll probably use that, unless it turns out it breaks something.

api/handlers.go
```diff
@@ -1096,11 +1105,13 @@ func (api *API) trackUploadPOST(u *database.User, w http.ResponseWriter, req *ht
...
 	// a separate one.
-	go api.checkUserQuotas(context.Background(), u)
+	if u != nil && !u.ID.IsZero() {
+		go api.checkUserQuotas(context.Background(), u)
+	}
 }
 
 // trackDownloadPOST registers a new download in the system.
-func (api *API) trackDownloadPOST(u *database.User, w http.ResponseWriter, req *http.Request, ps httprouter.Params) {
+func (api *API) trackDownloadPOST(_ *database.User, w http.ResponseWriter, req *http.Request, ps httprouter.Params) {
```
peterjan, line 1114: I wouldn't track anon downloads

api/handlers.go
```diff
@@ -1136,7 +1147,13 @@ func (api *API) trackDownloadPOST(u *database.User, w http.ResponseWriter, req *
 		api.WriteError(w, err, http.StatusInternalServerError)
 		return
 	}
-	err = api.staticDB.DownloadCreate(req.Context(), *u, *skylink, downloadedBytes)
+	u, _, err := api.userFromRequest(req)
+	if err != nil {
+		// This will be tracked as an anonymous request.
```
peterjan, line 1152: `nit` 
```
u, _, _ := api.userFromRequest(req) // ignore error, user can be nil
```

api/handlers.go
```diff
@@ -1225,6 +1242,25 @@ func (api *API) checkUserQuotas(ctx context.Context, u *database.User) {
 	}
 }
 
+// userFromRequest checks the requests for various forms of authentication (API
+// key, cookie, authorization header) and returns user information based on
+// those.
+func (api *API) userFromRequest(req *http.Request) (*database.User, jwt2.Token, error) {
+	// Check for an API key.
+	u, tk, err := api.userAndTokenByAPIKey(req)
+	if err != nil && !errors.Contains(err, ErrNoAPIKey) {
```
peterjan, line 1251: we're changing the error handling here quite a bit, right? before we were checking all kinds of errors and sometimes even return an internal server error, it seems like that's all gone now, should we consider this is a breaking change wrt backwards compat? 

I don't think so in this case, but I just wanted to comment on it because I'm not sure why all of those error checks are not needed anymore all of a sudden

database/upload.go
```diff
@@ -15,6 +15,7 @@ import (
 type Upload struct {
 	ID        primitive.ObjectID `bson:"_id,omitempty" json:"id"`
 	UserID    primitive.ObjectID `bson:"user_id,omitempty" json:"userId"`
+	IP        string             `bson:"ip" json:"IP"`
```
peterjan, line 18: I think `uploader_ip` is a more descriptive name for this field.

test/database/download_test.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"context"
+	"testing"
+
+	"github.com/SkynetLabs/skynet-accounts/database"
+	"github.com/SkynetLabs/skynet-accounts/test"
+)
+
+// TestDownloadCreateAnon ensures that UploadCreate can create anonymous downloads.
```
peterjan, line 11:
```suggestion
// TestDownloadCreateAnon ensures that DownloadCreate can create anonymous downloads.
```

test/database/download_test.go
```diff
@@ -0,0 +1,29 @@
...
+func TestDownloadCreateAnon(t *testing.T) {
+	ctx := context.Background()
```
peterjan, line 13: `nit` this is not great in tests because it'll just hang if something is wrong on db level

test/database/download_test.go
```diff
@@ -0,0 +1,29 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	sl := test.RandomSkylink()
+	skylink, err := db.Skylink(ctx, sl)
+	if err != nil {
+		t.Fatal(err)
+	}
+	// Register an anonymous download.
+	err = db.DownloadCreate(ctx, nil, *skylink, 123)
```
peterjan, line 25: this is not really testing anything right, if the impl. of `DownloadCreate` was just `return nil` this test would succeed, should query the database and assert it got inserted, and the user id was the anon user

test/database/upload_test.go
```diff
@@ -240,3 +239,41 @@ func TestUnpinUploads(t *testing.T) {
...
+		t.Fatal(err)
+	}
+	sl := test.RandomSkylink()
+	skylink, err := db.Skylink(ctx, sl)
+	if err != nil {
+		t.Fatal(err)
+	}
+	// Register an anonymous upload.
+	ip := "1.0.2.233"
+	up, err := db.UploadCreate(ctx, nil, ip, *skylink)
```
peterjan, line 258: kind of odd the return values for `UploadCreate` and `DownloadCreate` are not identical

api/handlers.go
peterjan: This will cause issues for sure, it's a huge amount of db records.  
I don't think it's a good idea to track anon downloads for that reason.

peterjan: I thought about this a bit more and I think we'll be adding about 70-80mb in storage space for every 1M downloads. So let's say we have 10M downloads a month, but probably it's closer to 20M and soon it'll be more, so every month we'll be easily adding 1GiB of storage. 

Then the second thing is that there's 3 indices, which is about 25mib per 1M documents. So again here after 1M of conservative downloads that's 250mib in extra index size, every month. Mongo usually keeps it entire index in RAM, not always and not all the time, but in general it does. When you have a mongo cluster you should generally make it so all indexes fit in memory. I'm not saying it won't, but it's not terribly scalable.

Then third is that I'm concerned about the writes to the PRIMARY, we write to a single primary and I have once experienced writes to easily take up to 200ms. Tracking anon downloads will essentially cause our total req/s to be writes/s to our PRIMARY. This could very well lead to worse write times.

I would love download stats, like everyone, I'm just not sure mongo is the place for it, or the tool to get that job done.
And yes, we can tackle this problem when it arises, I'm just saying that it will, and it will probably be in 3-6 maybe even 9 months from now, but it will cause issues so I don't see why we should add something we know we'll have to eventually fix.

ChrisSchinnerl: It's probably a ton of data but my impression from the description was that this will be removed in a f/u?

peterjan:
what was the bug by the way? 🙂

ro-tex:
> what was the bug by the way? 🙂

`DownloadRecent` didn't have a user parameter. It was getting the most recent download of the given skylink (within the permitted time range), regardless of which user owned it. This results in some users' downloads being counted towards other users'. While the total downloaded amount stays the same, this brings down the total number of downloads.

api/handlers.go
```diff
@@ -1077,7 +1079,14 @@ func (api *API) trackUploadPOST(u *database.User, w http.ResponseWriter, req *ht
...
 		return
 	}
-	_, err = api.staticDB.UploadCreate(req.Context(), *u, *skylink)
+	u, _, err := api.userFromRequest(req)
+	if err != nil {
+		// This will be tracked as an anonymous request.
+		// The assignment below is redundant but adds clarity.
+		u = nil
+	}
+	ip := validateIP(req.Form.Get("ip"))
```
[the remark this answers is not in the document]
ro-tex: yes, this is easy in nginx.

api/handlers.go
ro-tex: Removed.

database/upload.go
ro-tex: Added.

test/utils.go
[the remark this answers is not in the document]
ro-tex: I added the `AnonUser` but I also changed a lot of arguments to just not be pointers. Go's auto zero values will do the rest for us and we won't need to make extra checks.

database/upload.go
ro-tex: I find that a bit redundant but fine, no harm in that.

test/database/download_test.go
ro-tex: file deleted

test/database/download_test.go
ro-tex: file deleted

test/database/download_test.go
ro-tex: I'm sure I follow. What is going to hang, the test or the context?
The context should be garbage collected after the function exits, right?

api/handlers.go
ro-tex: The only change of behaviour is that `internal server errors` coming from processing API keys are now replaced by `unauthorized` errors. This has always been the standard behaviour when processing cookies and authorization tokens. Having in mind that API keys haven't really been announced, I wouldn't worry about being that strict with backwards compatibility.

If we want to be stricter with this and give the user really accurate information why their request is being denied, I can go over all the possible errors that can be returned by `userAndTokenByAPIKey` and `userAndTokenByRequestToken` and add proper responses.

test/database/download_test.go
peterjan: the test might hang, because the context won't time out 🤷‍♂️    
was just a suggestion to add a timeout

api/handlers.go
peterjan: ok, I was talking about `database.ErrUserNotFound` in particular, we used to handle that explicitly and now we don't anymore, but if you say it's fine then I'm fine with that

database/upload.go
```diff
@@ -13,11 +13,12 @@ import (
...
 // Upload ...
 type Upload struct {
-	ID        primitive.ObjectID `bson:"_id,omitempty" json:"id"`
-	UserID    primitive.ObjectID `bson:"user_id,omitempty" json:"userId"`
-	SkylinkID primitive.ObjectID `bson:"skylink_id,omitempty" json:"skylinkId"`
-	Timestamp time.Time          `bson:"timestamp" json:"timestamp"`
-	Unpinned  bool               `bson:"unpinned" json:"-"`
+	ID         primitive.ObjectID `bson:"_id,omitempty" json:"id"`
+	UserID     primitive.ObjectID `bson:"user_id,omitempty" json:"userId"`
+	UploaderIP string             `bson:"uploader_ip" json:"UploaderIP"`
```
peterjan, line 18:
```suggestion
	UploaderIP string             `bson:"uploader_ip" json:"uploaderIP"`
```

peterjan (approved):

api/handlers.go
```diff
@@ -1225,6 +1235,25 @@ func (api *API) checkUserQuotas(ctx context.Context, u *database.User) {
...
+	u, tk, err := api.userAndTokenByAPIKey(req)
+	if err != nil && !errors.Contains(err, ErrNoAPIKey) {
+		return nil, nil, err
+	}
+	// If there is no API key check for a token.
+	if errors.Contains(err, ErrNoAPIKey) {
+		u, tk, err = api.userAndTokenByRequestToken(req)
+		if err != nil {
+			return nil, nil, err
+		}
```
ChrisSchinnerl, line 1252: Suggestion:

```go
if errors.Contains(err, ErrNoAPIKey) {
  return api.userAndTokenByRequestToken(req)
}
return u, tk, err
```

ChrisSchinnerl (approved):

api/handlers.go
ro-tex: I agree with this approach. I've gotten some flak in reviews for using it, though.
Will add it in a F/U.

PR #173, opened 2022-03-24
API keys are only accepted by the `/track/*` and `/user/limits` endpoints

ro-tex:
# PULL REQUEST

## Overview

API keys are only accepted by the `/track/*` and `/user/limits` endpoints

API keys cannot be used via a GET parameter.

## Checklist
 - [ ] Changelog file created

### DIFF
--- a/api/routes.go
+++ b/api/routes.go
@@ -14,6 +14,9 @@ var (
 	// APIKeyHeader holds the name of the header we use for API keys. This
 	// header name matches the established standard used by Swagger and others.
 	APIKeyHeader = "Skynet-API-Key" // #nosec
+	// ErrAPIKeyNotAllowed is an error returned when an API key was passed to an
+	// endpoint that doesn't allow API key use.
+	ErrAPIKeyNotAllowed = errors.New("this endpoint does not allow the use of API keys")
 	// ErrNoAPIKey is an error returned when we expect an API key but we don't
 	// find one.
 	ErrNoAPIKey = errors.New("no api key found")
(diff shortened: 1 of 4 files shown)


test/api/apikeys_test.go
```diff
@@ -124,20 +121,12 @@ func testPrivateAPIKeysUsage(t *testing.T, at *test.AccountsTester) {
...
-	err = json.Unmarshal(body, &us)
-	if err != nil {
-		t.Fatal(err)
+	h := map[string]string{
+		api.APIKeyHeader: akWithKey.Key.String(),
 	}
-	if us.TotalUploadsSize != uploadSize || us.NumUploads != 1 || us.BandwidthUploads != skynet.BandwidthUploadCost(uploadSize) {
-		t.Fatalf("Unexpected user stats. Expected TotalUploadSize %d (got %d), NumUploads 1 (got %d), BandwidthUploads %d (got %d).",
-			uploadSize, us.TotalDownloadsSize, us.NumUploads, skynet.BandwidthUploadCost(uploadSize), us.BandwidthUploads)
+	ulg, _, err := at.UserLimits("", h)
```
ChrisSchinnerl, line 127: Seems like you are not validating the error here?

api/routes.go
```diff
@@ -89,11 +92,30 @@ func (api *API) noAuth(h HandlerWithUser) httprouter.Handle {
...
+		}
+
 		// Check for an API key.
-		u, token, err := api.userAndTokenByAPIKey(req)
+		ak, err := apiKeyFromRequest(req)
+		if err != nil {
+			api.WriteError(w, err, http.StatusUnauthorized)
+			return
+		}
+		if !allowsAPIKey {
```
peterjan, line 114: move up this check? if it's not allowed we shouldn't try and parse it either

peterjan (commented):
should we add a unit test that asserts API keys are no longer accepted by most routes?

api/routes.go
```diff
@@ -89,11 +92,30 @@ func (api *API) noAuth(h HandlerWithUser) httprouter.Handle {
...
+		if err == nil {
+			// Embed the verified token in the context of the request.
+			ctx := jwt.ContextWithToken(req.Context(), token)
+			h(u, w, req.WithContext(ctx), ps)
+			return
+		}
+
 		// Check for an API key.
-		u, token, err := api.userAndTokenByAPIKey(req)
+		ak, err := apiKeyFromRequest(req)
```
ChrisSchinnerl, line 109: f/u suggestion

`apiKeyFromRequest` should only be called in this method. Then it attaches the api key to the context.
All handlers read the token from the context.

That's for extra safety in case of potential inconsistencies because we are only checking whether API Key authentication is allowed in this method and not in the handlers.

api/routes.go
ro-tex: Chris wanted a meaningful error message when a client tries to authenticate with an API key on an endpoint that doesn't support it. That's why I check for a valid API key first and then issue this special error, instead of 
```go
if !allowAPIKey {
   // respond with a 401
   return
}
```

test/api/apikeys_test.go
ro-tex: Added.

api/routes.go
ro-tex: This is almost exactly what's happening. The only exceptions are the two handlers which don't use the `withAuth` middleware because they need to allow unauthenticated requests as well - `userLimitsGET` and `userLimitsSkylinkGET`.

Outside of those two, we only check the API key here, then fetch the issuing user and pass it to the handlers directly.

ro-tex:
> should we add a unit test that asserts API keys are no longer accepted by most routes?

I'm not sure how useful that would be but added.

test/api/handlers_test.go
```diff
@@ -941,12 +942,10 @@ func testTrackingAndStats(t *testing.T, at *test.AccountsTester) {
...
 	expectedStats.RawStorageUsed += skynet.RawStorageUsed(0)
 
-	// Call trackDownload without a cookie. Expect this to succeed.
-	// While we expect this to succeed, it won't be counted towards the user's
-	// quota, so we don't increment the expected stats.
+	// Call trackDownload without a cookie. Expect this to fail.
 	at.ClearCredentials()
 	_, err = at.TrackDownload(skylink.String(), 100)
-	if err != nil {
+	if err == nil {
```
peterjan, line 948: you still fail with `t.Fatal(err)` in the line below

peterjan (approved):

kwypchlo (approved):

test/api/handlers_test.go
ro-tex: Oh, right, will F/U.

PR #182, opened 2022-04-01
Implement `GET /uploadinfo/:skylink`

ro-tex:
# PULL REQUEST

## Overview

Implement `GET /uploadinfo/:skylink`

## Checklist
 - [ ] Changelog file created

### DIFF
--- a/api/routes.go
+++ b/api/routes.go
@@ -81,6 +81,9 @@ func (api *API) buildHTTPRoutes() {
 	api.staticRouter.GET("/stripe/prices", api.noAuth(api.stripePricesGET))
 
 	api.staticRouter.GET("/.well-known/jwks.json", api.noAuth(api.wellKnownJWKSGET))
+
+	// Internal endpoint. NEVER EXPOSE THESE!
+	api.staticRouter.GET("/uploadinfo/:skylink", api.noAuth(api.uploadInfoGET))
 }
 
 // noAuth is a pass-through method used for decorating the request and
--- a/api/routes.go
+++ b/api/routes.go
@@ -82,7 +82,7 @@ func (api *API) buildHTTPRoutes() {
 
 	api.staticRouter.GET("/.well-known/jwks.json", api.noAuth(api.wellKnownJWKSGET))
 
-	// Internal endpoint. NEVER EXPOSE THESE!
+	// Internal endpoint. Never expose these!
 	api.staticRouter.GET("/uploadinfo/:skylink", api.noAuth(api.uploadInfoGET))
 }
 
--- a/api/upload.go
+++ b/api/upload.go
@@ -49,15 +49,13 @@ func (api *API) uploadInfoGET(_ *database.User, w http.ResponseWriter, req *http
 		api.WriteError(w, errors.AddContext(err, "failed to get uploads"), http.StatusInternalServerError)
 		return
 	}
-	// Get a set of all users that created these uploads.
-	uploaderSet := make(map[primitive.ObjectID]interface{})
-	for _, up := range ups {
-		uploaderSet[up.UserID] = struct{}{}
-	}
-	// Fet those users' data.
+	// Get the user data of all uploaders.
 	uploaders := make(map[primitive.ObjectID]database.User)
-	for upID := range uploaderSet {
-		u, err := api.staticDB.UserByID(ctx, upID)
+	for _, up := range ups {
+		if _, exist := uploaders[up.UserID]; exist {
+			continue
+		}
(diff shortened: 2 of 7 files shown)


api/routes.go
```diff
@@ -81,6 +81,9 @@ func (api *API) buildHTTPRoutes() {
 	api.staticRouter.GET("/stripe/prices", api.noAuth(api.stripePricesGET))
 
 	api.staticRouter.GET("/.well-known/jwks.json", api.noAuth(api.wellKnownJWKSGET))
+
+	// Internal endpoint. NEVER EXPOSE THESE!
```
peterjan, line 85: wouldn't yell, always kind of weird to see that in a codebase imo

api/upload.go
```diff
@@ -0,0 +1,87 @@
...
+		api.WriteError(w, errors.AddContext(err, "failed to get skylink"), http.StatusInternalServerError)
+		return
+	}
+	// Get all uploads of this skylink.
+	ups, err := api.staticDB.UploadsBySkylinkID(ctx, sl.ID)
+	if err != nil {
+		api.WriteError(w, errors.AddContext(err, "failed to get uploads"), http.StatusInternalServerError)
+		return
+	}
+	// Get a set of all users that created these uploads.
```
peterjan, line 52: `nit` you can just do this in one loop, no need to build two separate maps for this

```go
uploaders := make(map[primitive.ObjectID]database.User)
for _, up := range ups {
     _, fetched := uploaders[up.UserID]
     if fetched {
         continue
     }
     ...
}
```

api/upload.go
```diff
@@ -0,0 +1,87 @@
...
+		}
+		if err != nil {
+			api.WriteError(w, errors.AddContext(err, "failed to get uploader"), http.StatusInternalServerError)
+			return
+		}
+		uploaders[u.ID] = *u
+	}
+	// Create the final list of hydrated uploads.
+	upsHydrated := make([]UploadInfo, len(ups))
+	for i, up := range ups {
```
peterjan, line 72: hm so this means that if I pass you 2 skylinks, and you can't find any info for the first and you find 1 upload for the second, I only get one return element, I was kind of hoping the accounts API would return a shell object that only has the skylink

thinking about it more though, I'm not sure what's the cleanest design, I can obviously do this on my side as well

test/tester.go
```diff
@@ -498,3 +498,36 @@ func (at *AccountsTester) UserLimitsSkylink(sl string, unit, apikey string, head
...
+	err = json.Unmarshal(b, &resp)
+	if err != nil {
+		return nil, http.StatusInternalServerError, errors.AddContext(err, "failed to marshal the body JSON")
+	}
+	return resp, r.StatusCode, nil
+}
+
+// UploadsDELETE performs `DELETE /user/uploads/:skylink`
+// TODO Remove this one when we merge the tester refactoring.
+func (at *AccountsTester) UploadsDELETE(skylink string) (int, error) {
```
peterjan, line 524: was this committed by accident perhaps?

peterjan (dismissed):

api/routes.go
ro-tex: Hm, ok. I'm used to seeing all-caps notices in codebases. I don't really consider it yelling - this is not a conversation.

test/tester.go
ro-tex: No, I already have this method on another branch and they are bound to either conflict on merge or, even worse, merge seamlessly and then fail to compile. This is a tag that will tell me which one to delete.

api/upload.go
ro-tex: Oh, that's true. Thanks!

peterjan (approved):

api/upload.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	ctx := req.Context()
+	sl, err := api.staticDB.Skylink(ctx, skylink)
+	if err != nil {
+		api.WriteError(w, errors.AddContext(err, "failed to get skylink"), http.StatusInternalServerError)
+		return
+	}
+	// Get all uploads of this skylink.
+	ups, err := api.staticDB.UploadsBySkylinkID(ctx, sl.ID)
```
ChrisSchinnerl, line 47: f/u 

Definitely not a requirement for this PR but I wonder if PJ could turn this into some fancy query which fetches all uploads for a skylink id and then joins them with their users where possible. So that all of that happens with a single query.

ChrisSchinnerl (approved):

api/upload.go
ro-tex: That's fine, I can do that as well. I'll open a F/U.

PR #190, opened 2022-04-06
Auto Release

MSevey:
# PULL REQUEST

## Overview
This PR updates the github actions to enable daily releases as well as more automated manual releases. 

The dockerfile lint action was refactored out to use the common action for the organization.

The old test action was renamed to `tests.yml` and converted to a reusable workflow.

A new `ci.yml` action was added to be used as the stand CI action performed on pull requests and pushes to main.

The `release.yml` file is run daily at 1:15am and will trigger a release with release notes if the lining and tests pass and there has been changes since the last tag.  This action also allows for manually triggering it and being able to create a path, minor, or major release. 

Pending https://github.com/SkynetLabs/.github/pull/5


.github/workflows/ci_release.yml
```diff
@@ -0,0 +1,69 @@
...
+  changes:
+    needs: [hadolint, test]
+    runs-on: ubuntu-latest
+    outputs:
+      updates: ${{steps.changes.outputs.any == 'true'}}
+    if: ${{ github.event_name != 'push' }}
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0 # Required due to the weg Git works, without it this action won't be able to find any or the correct tags
```
m-cat, line 54:
```suggestion
          fetch-depth: 0 # Required due to the way Git works, without it this action won't be able to find any or the correct tags
```

Should be this I think.

.github/workflows/ci_release.yml
```diff
@@ -0,0 +1,69 @@
...
+      version:
+        # Friendly description to be shown in the UI instead of 'name'
+        description: "Semver type of new version (major / minor / patch)"
+        # Input has to be provided for the workflow to run
+        required: true
+        type: choice
+        options:
+          - patch
+          - minor
+          - major
```
m-cat, line 19: Is this input used anywhere?

m-cat (changes requested):
Looks like there is an unused `version` input, unless I'm missing something. Other than that, looks good.

.github/workflows/ci_release.yml
MSevey: These are options that you as a user will be able to select when triggering the workflow manually to create a release.

m-cat: That much I got. 😅 What is the `version` used for in this workflow?

MSevey: Oh i see the error. fixed.

m-cat (approved):

PR #196, opened 2022-04-12
Fix user creation time being zero.

### LINKED ISSUE #195: Wrong creation date for new accounts
## Describe the bug
Newly created accounts report `"0001-01-01T00:00:00Z"` in `createdAt` field.

## To Reproduce
1. Go to skynetfree.net
2. Create an account
3. Open devtools and inspect `/user` response.

## Expected behavior
Default to current datetime, I think?

## Screenshots
image

ro-tex:
# PULL REQUEST

## Checklist
 - [ ] Changelog file created

## Issues Closed
Closes #195

### DIFF
--- a/database/user.go
+++ b/database/user.go
@@ -49,10 +49,6 @@ var (
 	// AnonUser is a helper struct that we can use when we don't have a relevant
 	// user, e.g. when an upload is made by an anonymous user.
 	AnonUser = User{}
-	// True is a helper for when we need to pass a *bool to MongoDB.
-	True = true
-	// False is a helper for when we need to pass a *bool to MongoDB.
-	False = false
 	// UserLimits defines the speed limits for each tier.
 	// RegistryDelay delay is in ms.
 	UserLimits = map[int]TierLimits{
@@ -60,7 +56,7 @@ var (
 			TierName:        "anonymous",
 			UploadBandwidth: 5 * mbpsToBytesPerSecond,
 			// TODO: temporarily lowered the download bandwidth on the anon tier
-			// from 20mbps to 5mpbs
+			// from 20mbps to 5mbps
 			DownloadBandwidth: 5 * mbpsToBytesPerSecond,
 			MaxUploadSize:     1 * skynet.GiB,
 			MaxNumberUploads:  0,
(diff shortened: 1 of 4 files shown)


test/api/handlers_test.go
```diff
@@ -133,6 +133,11 @@ func testHandlerUserPOST(t *testing.T, at *test.AccountsTester) {
 	if err != nil {
 		t.Fatal("Error while fetching the user from the DB. Error ", err.Error())
 	}
+	// Make sure the creation timestamp is correct.
+	now := time.Now().UTC()
+	if u.CreatedAt.Before(now.Add(-1*time.Minute)) || u.CreatedAt.After(now.Add(time.Minute)) {
```
MSevey, line 138: nit: would it not be sufficient to just verify that the timestamp is not zero?

MSevey (approved):

test/api/handlers_test.go
ro-tex: For this particular issue - yes, I can just compare to the zero value. But since we're adding a check anyway, we might as well make sure the time is accurate. This would catch, for example, an omitted `.UTC()` call and other things like that.

database/user.go
```diff
@@ -334,8 +330,18 @@ func (db *DB) UserCreate(ctx context.Context, emailAddr, pass, sub string, tier
...
 		Tier:                             tier,
+		CreatedAt:                        time.Now().UTC(),
+		MigratedAt:                       time.Time{},
+		SubscribedUntil:                  time.Time{},
+		SubscriptionStatus:               "",
+		SubscriptionCancelAt:             time.Time{},
+		SubscriptionCancelAtPeriodEnd:    false,
+		StripeID:                         "",
+		QuotaExceeded:                    false,
+		PubKeys:                          make([]PubKey, 0),
```
peterjan, line 344: this removes the need for that type query right?

database/user.go
```diff
@@ -334,8 +330,18 @@ func (db *DB) UserCreate(ctx context.Context, emailAddr, pass, sub string, tier
 		EmailConfirmationToken:           emailConfToken,
 		EmailConfirmationTokenExpiration: time.Now().UTC().Add(EmailConfirmationTokenTTL).Truncate(time.Millisecond),
 		PasswordHash:                     string(passHash),
+		RecoveryToken:                    "",
 		Sub:                              sub,
 		Tier:                             tier,
+		CreatedAt:                        time.Now().UTC(),
+		MigratedAt:                       time.Time{},
```
peterjan, line 337: what's the reason we explicitly define these defaults?

test/api/handlers_test.go
peterjan: `nit` could use `Truncate` maybe 

```go
if u.CreatedAt.Truncate(time.Minute) != now.Truncate(time.Minute)
```

peterjan (approved):

test/api/handlers_test.go
ro-tex: > `nit` could use `Truncate` maybe
> 
> ```go
> if u.CreatedAt.Truncate(time.Minute) != now.Truncate(time.Minute)
> ```

That opens us up to NDFs. `Truncate` doesn't round, it just drops the rest of the time, so 12:10:59.99 will truncate to 12:10 and 12:11:00.01 will truncate to 12:11.

database/user.go
ro-tex: We still have all pre-existing users who don't have any pubkeys.

PR #197, opened 2022-04-12
Stop tracking registry accesses

ro-tex:
# PULL REQUEST

## Overview

As we have high priority work that trumps DB performance tracking and optimisations, we wanted to have some quick wins before we park this effort. This PR provides those quick wins by removing registry tracking.

What's done here: Stop tracking registry accesses and drop all DB collections related to those.

## Checklist
 - [ ] Changelog file created

### DIFF
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -1200,22 +1200,14 @@ func (api *API) trackDownloadPOST(u *database.User, w http.ResponseWriter, req *
 }
 
 // trackRegistryReadPOST registers a new registry read in the system.
-func (api *API) trackRegistryReadPOST(u *database.User, w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
-	_, err := api.staticDB.RegistryReadCreate(req.Context(), *u)
-	if err != nil {
-		api.WriteError(w, err, http.StatusInternalServerError)
-		return
-	}
+func (api *API) trackRegistryReadPOST(_ *database.User, w http.ResponseWriter, _ *http.Request, _ httprouter.Params) {
+	// Tracking registry reads is disabled.
 	api.WriteSuccess(w)
 }
 
 // trackRegistryWritePOST registers a new registry write in the system.
-func (api *API) trackRegistryWritePOST(u *database.User, w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
-	_, err := api.staticDB.RegistryWriteCreate(req.Context(), *u)
-	if err != nil {
-		api.WriteError(w, err, http.StatusInternalServerError)
-		return
-	}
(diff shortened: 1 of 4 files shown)


api/handlers.go
```diff
@@ -1200,22 +1200,14 @@ func (api *API) trackDownloadPOST(u *database.User, w http.ResponseWriter, req *
...
 // trackRegistryReadPOST registers a new registry read in the system.
-func (api *API) trackRegistryReadPOST(u *database.User, w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
-	_, err := api.staticDB.RegistryReadCreate(req.Context(), *u)
-	if err != nil {
-		api.WriteError(w, err, http.StatusInternalServerError)
-		return
-	}
+func (api *API) trackRegistryReadPOST(_ *database.User, w http.ResponseWriter, _ *http.Request, _ httprouter.Params) {
+	// Tracking registry reads is disabled.
 	api.WriteSuccess(w)
```
MSevey, line 1205: nit: shouldn't we return an error if it is disabled?

MSevey:
Is registry tracking useful information to have at some point? Are we just dropping it because it is currently impacting performance?

If so, we could disable the API but still keep the data collections right?

ro-tex:
> Is registry tracking useful information to have at some point? Are we just dropping it because it is currently impacting performance?
> 
> If so, we could disable the API but still keep the data collections right?

1. We don't have any indication that we have (or don't have) any performance issues. I keep repeating this but I want to make it clear that we're basing all of this on gut feeling and no data.
2. Personally, I find it positive to have this metric and I would have preferred to keep it. The fact is, though, that since its inception nobody has ever looked at it, so maybe I'm alone in my interest in it.
3. The data collection is the only action we perform on this data. The DB writes are the only thing that could potentially affect our performance.

In a perfect world, my preferred solution would be:
1. Keep gathering the metrics.
2. Replace our current strategy of one DB record per read/write with an aggregation - one record which holds the number of reads and write performed by a single user for each day. (Anon actions will be grouped together.) This will reduce the fidelity of the statistics (we'll only know the calls per day and not specific times during the day) but it will replace a relatively expensive record creation with a relatively inexpensive atomic increment of an integer field.
3. Run a batch job (either manually or in some form of compat code) that aggregates all existing statistics into such aggregations.

api/handlers.go
ro-tex: We could. I don't expect nginx to do anything about it, though. It will only log an error, which is going to fill our error logs while providing no actionable information.

I don't feel strongly about this, so I'm fine with adding an error if you (or anyone else) prefer.

peterjan (approved):
LGTM

MSevey (approved):

PR #202, opened 2022-04-20
Ignore "IndexNotFound" errors and log the rest without panicking.

ro-tex:
# PULL REQUEST

## Checklist
 - [ ] Changelog file created

### DIFF
--- a/database/database.go
+++ b/database/database.go
@@ -199,9 +199,8 @@ func ensureDBSchema(ctx context.Context, db *mongo.Database, schema map[string][
 	}
 	// Drop indexes we no longer need.
 	_, err = db.Collection(collUsers).Indexes().DropOne(ctx, "email_unique")
-	// Ignore namespace not found errors.
-	if err != nil && !strings.Contains(err.Error(), "NamespaceNotFound") {
-		return err
+	if err != nil && !strings.Contains(err.Error(), "IndexNotFound") {
+		log.Debugf("Error while dropping index '%s': %v", "email_unique", err)
 	}
 	// Ensure current schema.
 	for collName, models := range schema {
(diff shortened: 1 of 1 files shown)


database/database.go
```diff
@@ -199,9 +199,8 @@ func ensureDBSchema(ctx context.Context, db *mongo.Database, schema map[string][
 	}
 	// Drop indexes we no longer need.
 	_, err = db.Collection(collUsers).Indexes().DropOne(ctx, "email_unique")
-	// Ignore namespace not found errors.
-	if err != nil && !strings.Contains(err.Error(), "NamespaceNotFound") {
```
MSevey, line 203: do we still want to ignore namespace errors? 

Also I assume there are still come errors we want to return right? This change now ignores all errors and just logs certain errors. 

Shouldn't we be explicating checking for index not found errors, and simply logging them, and all errors should be returned.

MSevey (changes requested):

database/database.go
ro-tex: 1. `NamespaceNotFound` - I thought we didn't need that anymore but it turns out we can still have it there. It's returned when the collection on which we try to drop the index does not exist. So, I'll add it back. Thanks for asking. :)
2. I don't think we should return any error here, no. This is a cleanup step which is there just for optimisation and failing it shouldn't affect the operation of the service. If there is any error that might affect the way the service runs, it will come up during the following steps where we ensure that all collections and indexes we need exist.

database/database.go
```diff
@@ -199,9 +199,8 @@ func ensureDBSchema(ctx context.Context, db *mongo.Database, schema map[string][
...
-		return err
+	if err != nil && !strings.Contains(err.Error(), "IndexNotFound") {
```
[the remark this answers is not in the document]
ro-tex: I'd rather not do that right now. We should probably move all base mongo stuff to a library and share it across all services, so we can stop duplicating functionality.

kwypchlo (approved):

MSevey (approved):

PR #205, opened 2022-04-28
MySky Allowlist

### LINKED ISSUE #204: Allowlist MySky's skylinks
## Overview

Allowlist the domains from which we want to load MySky:
  - skynet-mysky: AQCsSOIwqwn7lLCT0t110ImQJaI39HxrSrJ-GVNSltfUAQ
  - skynet-mysky-dev: AQBIMqRcHbGWXy4rlIwGW4Aa4v0w0xLb6JvUonnXazfxiw
  - sandbridge: AQASyOUdaov383UggiDN7izfcCH8k-3Z0FlPjtNyem1qMg

ro-tex:
# PULL REQUEST

## Overview

Create a list of always allowed skylinks in order to make MySky always available on all portals.

## Checklist
 - [ ] Changelog file created

## Issues Closed
Closes #204

### DIFF
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -40,6 +40,15 @@ var (
 	// flow fails. This error is sent instead of whatever internal error we had
 	// before in order to prevent an attacker from listing our users.
 	ErrInvalidCredentials = errors.New("invalid credentials")
+
+	// MyskyAllowlist contains skylinks we need to make available in order for
+	// users to be able to use MySky on all portals, including ones that require
+	// user authentication.
+	MyskyAllowlist = map[string]interface{}{
+		"AQCsSOIwqwn7lLCT0t110ImQJaI39HxrSrJ-GVNSltfUAQ": struct{}{},
+		"AQBIMqRcHbGWXy4rlIwGW4Aa4v0w0xLb6JvUonnXazfxiw": struct{}{},
+		"AQASyOUdaov383UggiDN7izfcCH8k-3Z0FlPjtNyem1qMg": struct{}{},
+	}
 )
 
 type (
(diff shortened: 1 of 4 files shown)


ChrisSchinnerl (approved):

database/skylink.go
```diff
@@ -118,5 +117,7 @@ func ExtractSkylinkHash(skylink string) (string, error) {
 
 // ValidSkylinkHash returns true if the given string is a valid skylink hash.
```
ChrisSchinnerl, line 118: Is the comment here outdated? It talks about a hash but we are actually just loading a skylink.

ChrisSchinnerl (approved):

api/handlers.go
```diff
@@ -40,6 +40,15 @@ var (
 	// flow fails. This error is sent instead of whatever internal error we had
 	// before in order to prevent an attacker from listing our users.
 	ErrInvalidCredentials = errors.New("invalid credentials")
+
+	// MyskyAllowlist contains skylinks we need to make available in order for
+	// users to be able to use MySky on all portals, including ones that require
+	// user authentication.
+	MyskyAllowlist = map[string]interface{}{
```
peterjan, line 47: Is this something that might eventually live in mongo? Obviously, I'm not suggesting consulting mongo on every call but the whitelist could get reloaded every 5' in the background or whatever. These skylinks are pretty static so I guess hardcoding them is fine but at the same time, it might be a useful feature to grant full access to certain skylinks for everyone. Could turn into a pretty cool feature 👍

api/handlers.go
```diff
@@ -509,12 +518,19 @@ func (api *API) userLimitsSkylinkGET(u *database.User, w http.ResponseWriter, re
 		api.WriteJSON(w, respAnon)
 		return
 	}
+	// For all links that belong to MySky we return the first paid tier, so
+	// anyone can access them, even on portals which require authentication or
+	// premium accounts.
+	if _, ok := MyskyAllowlist[skylink]; ok {
```
peterjan, line 524: `nit` you can move this before the validation, not sure what `database.ValidSkylinkHash` does but it's probably just a `LoadString`? Feel free to resolve this comment if you prefer to have it here.

peterjan (approved):

database/skylink.go
ro-tex: It's incorrectly worded - it was always meant to say "skylink". I'll change it in the next PR.

api/handlers.go
ro-tex: If we're talking about a general allowlist that's in the control of the operator, that makes sense and it could turn into a cool feature if we come up with a nice user story. But here we have a rather special case where we, as the devs of the portal, force a set of skylinks to be accessible on all portals without the portal operator having any say in that. While we have a pretty solid reason to do this, I really hope that this will remain a strictly isolated incident and this list will never grow.

api/handlers.go
ro-tex: I see how we don't really need validation before checking against a hardcoded list but I like having all validation done before any logic is triggered because it's nice for readability. So, I'd rather keep it here.
Thanks for your input, though!

PR #211, opened 2022-05-25
Change the http verb of /stripe/billing from POST to GET.

ro-tex:
# PULL REQUEST

## Checklist
 - [ ] Changelog file created

## Issues Closed
Closes SKY-667

### DIFF
--- a/api/routes.go
+++ b/api/routes.go
@@ -77,7 +77,9 @@ func (api *API) buildHTTPRoutes() {
 	api.staticRouter.POST("/user/recover/request", api.WithDBSession(api.noAuth(api.userRecoverRequestPOST)))
 	api.staticRouter.POST("/user/recover", api.WithDBSession(api.noAuth(api.userRecoverPOST)))
 
-	api.staticRouter.GET("/stripe/billing", api.WithDBSession(api.withAuth(api.stripeBillingGET, false)))
+	api.staticRouter.GET("/stripe/billing", api.WithDBSession(api.withAuth(api.stripeBillingHANDLER, false)))
+	// `POST /stripe/billing` is deprecated. Please use `GET /stripe/billing`.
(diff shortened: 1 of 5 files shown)


meeh0w (approved):
Thanks for that 👍

api/routes.go
```diff
@@ -77,7 +77,7 @@ func (api *API) buildHTTPRoutes() {
 	api.staticRouter.POST("/user/recover/request", api.WithDBSession(api.noAuth(api.userRecoverRequestPOST)))
 	api.staticRouter.POST("/user/recover", api.WithDBSession(api.noAuth(api.userRecoverPOST)))
 
-	api.staticRouter.POST("/stripe/billing", api.WithDBSession(api.withAuth(api.stripeBillingPOST, false)))
+	api.staticRouter.GET("/stripe/billing", api.WithDBSession(api.withAuth(api.stripeBillingGET, false)))
```
MSevey, line 80: Do we need to actually leave the `POST` and return a deprecated message to ensure we maintain API compatibility?

Because if other portal operators are writing any custom tooling to hit this endpoint, we are going to break their tooling right?

ro-tex: Sure, good call. How do we do API deprecation messages? We don't really have API documentation.

MSevey: What you did is fine. 

Sometimes I'll move the endpoint to the bottom of the list too so the old deprecated endpoints are all together. Then if we had API documentation we would remove it.

MSevey: It is also fine to have the updated naming of the handler, since that is internal. We don't care if the name is mismatched for an old deprecated endpoint. Better to make it match for the in use endpoint.

MSevey (approved):

meeh0w (approved):

api/routes.go
ro-tex: Makes sense. I'll cover this in a F/U.

PR #221, opened 2022-06-10
Fix Stripe duplicate sub canceling.

ro-tex:
# PULL REQUEST

## Overview

This PR fixes a problem with our ability to cancel subscriptions for users. We want to be able to do that, so if a user manages to create two subscriptions in parallel, we can cancel the earlier one in order to not charge them twice.  
This code has been in there forever but apparently didn't work properly. It's an [old](https://github.com/cartalyst/stripe/issues/84) peculiarity of Stripe's service.

The PR is a bit longer than it needs to because touching this part of the code allowed me to clean it up a little bit. I'll mark the relevant section.

## Checklist

- [ ] Changelog file created.

### DIFF
--- a/api/stripe.go
+++ b/api/stripe.go
@@ -34,11 +34,6 @@ var (
 	// `https://account.` prepended to it).
 	DashboardURL = "https://account.siasky.net"
 
-	// StripeTestMode tells us whether to use Stripe's test mode or prod mode
-	// plan and price ids. This depends on what kind of key is stored in the
-	// STRIPE_API_KEY environment variable.
-	StripeTestMode = false
-
 	// True is a helper for when we need to pass a *bool to Stripe.
 	True = true
 
@@ -122,9 +117,6 @@ func (api *API) processStripeSub(ctx context.Context, s *stripe.Subscription) er
 	}
 	// Cancel all subs aside from the latest one.
 	p := stripe.SubscriptionCancelParams{
-		Params: stripe.Params{
-			StripeAccount: &s.Customer.ID,
-		},
 		InvoiceNow: &True,
 		Prorate:    &True,
 	}
@@ -329,7 +321,8 @@ func (api *API) stripeWebhookPOST(_ *database.User, w http.ResponseWriter, req *
 			return
 		}
 		// Check the details about this subscription:
-		s, err := sub.Get(hasSub.Sub, nil)
+		var s *stripe.Subscription
+		s, err = sub.Get(hasSub.Sub, nil)
 		if err != nil {
 			api.staticLogger.Debugln("Webhook: Failed to fetch sub:", err)
 			api.WriteError(w, err, http.StatusInternalServerError)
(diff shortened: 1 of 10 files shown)


MSevey (approved):

meeh0w (commented):
Tested it on dev3 and it does indeed cancel the previous subscription, so _yay :tada:_

That said, right after a success message in the logs:

```
Successfully cancelled sub with id 'sub_1LAqNgIzjULiPWN6yzlQWoGh' for user '62a985ddf4de246a1bf21adf' with Stripe customer id 'cus_LsbaAY5TJTolGc'.
```

I'm also seeing a failure message:
```
Failed to cancel sub with id '' for user '62a985ddf4de246a1bf21adf' with Stripe customer id 'cus_LsbaAY5TJTolGc'.

Error: '{\"code\":\"resource_missing\",\"doc_url\":\"https://stripe.com/docs/error-codes/resource-missing\",\"status\":404,\"message\":\"No such subscription: 'sub_1LAqNgIzjULiPWN6yzlQWoGh'\",\"param\":\"id\",\"request_id\":\"req_s10PgGfWowzGMv\",\"type\":\"invalid_request_error\"}'
```

This is kind of confusing, because the 2nd log says sub id is blank, but if you inspect the Stripe error you'll notice we're trying to cancel the same subscription for the 2nd time (id `sub_1LAqNgIzjULiPWN6yzlQWoGh`).

Functionality itself seems to be working, but there's something about this error I don't quite understand so I'd rather wait for @ro-tex to come and investigate.

ro-tex:
Good catch, @meeh0w! The issue is that the `sub.Cancel()` method on Stripe's API returns a sub object which I store in the same var I use to iterate over the list of active subs. If the operation fails, the response is an empty sub and the log output is useless. This is a good example of the advantages of immutable vars over mutable ones.

I fixed the overwrite and added a bit more debug output, so we can get a better understanding of the underlying issue. That issue, I suspect, is that we somehow get duplicate entries in the list of active subscriptions of the user that we get from Stripe.

This is the original code:
```go
	for _, subsc := range subs {
		if subsc == nil || (mostRecentSub != nil && subsc.ID == mostRecentSub.ID) {
			continue
		}
		if subsc.ID == "" {
			api.staticLogger.Warnf("Empty subscription ID! User ID '%s', Stripe ID '%s', subscription object '%+v'", u.ID.Hex(), u.StripeID, subs)
			continue
		}
		subsc, err = sub.Cancel(subsc.ID, &p)
		if err != nil {
			api.staticLogger.Warnf("Failed to cancel sub with id '%s' for user '%s' with Stripe customer id '%s'. Error: '%s'", subsc.ID, u.ID.Hex(), s.Customer.ID, err.Error())
		} else {
			api.staticLogger.Tracef("Successfully cancelled sub with id '%s' for user '%s' with Stripe customer id '%s'.", subsc.ID, u.ID.Hex(), s.Customer.ID)
		}
	}
```
I don't see how else we can hit both the `Failed` and `Successfully` log lines in this code with the same sub id unless it's duplicated. And we know that it's the same sub id from the example above where it's `sub_1LAqNgIzjULiPWN6yzlQWoGh`.

Let's deploy this on dev and see how it goes. I believe the logic is already fine and we only need to make sure that we're getting useful log output.

MSevey (approved):

ChrisSchinnerl (approved):

PR #222, opened 2022-06-10
Enable test-long in CI

MSevey:
# PULL REQUEST

## Overview
This PR fixes some lint errors, cleans up the Makefile, addresses missed `testing.Short()` calls, and enables the long tests in CI.

## Checklist

- [ ] All git commits are signed. (REQUIRED)
- [ ] All new methods or updated methods have clear docstrings.
- [ ] Testing added or updated for new methods.
- [ ] Verified if any changes impact the WebPortal Health Checks.
- [ ] Appropriate documentation updated.
- [ ] Changelog file created.

PR #223, opened 2022-06-20
Revert some changes

ro-tex:
# PULL REQUEST

## Overview

We inadvertently made some changes which break the logic of some unit tests. I'm reverting those changes here, adding some context and adding some other clean-ups.

The [original PR](https://github.com/SkynetLabs/skynet-accounts/pull/222) mentions lint issues. If the changes reverted here were fixes for linting issues then we should fix the linting setup in `Makefile` because I'm not getting any warnings/errors.

## Checklist

- [ ] Changelog file created.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,7 +29,7 @@ else
 	- DEL /F /Q cover output
 endif
 
-run = . 
+run = .
 
 # count says how many times to run the tests.
 count = 1
@@ -85,7 +85,7 @@ ifneq ("$(OS)","Windows_NT")
 endif
 
 # Define docker container name our test MongoDB instance.
-MONGO_TEST_CONTAINER_NAME=blocker-mongo-test-db
+MONGO_TEST_CONTAINER_NAME=accounts-mongo-test-db
 
 # start-mongo starts a local mongoDB container with no persistence.
 # We first prepare for the start of the container by making sure the test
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -136,11 +136,6 @@ test-long-ci:
 	@mkdir -p cover
 	GORACE='$(racevars)' go test -race --coverprofile='./cover/cover.out' -v -failfast -tags='testing debug netgo' -timeout=600s $(pkgs) -run=$(run) -count=$(count)
 
-# Cookie vars
-# TODO: Are these used?
-COOKIE_HASH_KEY="7eb32cfab5014d14394648dae1cf4e606727eee2267f6a50213cd842e61c5bce"
-COOKIE_ENC_KEY="65d31d12b80fc57df16d84c02a9bb62e2bc3b633388b05e49ef8abfdf0d35cf3"
-
```
ro-tex, line 143: These used to be required but now we generate random values for them during testing, so they can be removed.

api/handlers_test.go
```diff
@@ -132,7 +132,7 @@ func TestUserLimitsGetFromTier(t *testing.T) {
 			}
 		}()
 		// The call that we expect to log a critical.
-		_ = userLimitsGetFromTier("", math.MaxInt64, false, true)
+		_ = userLimitsGetFromTier("", math.MaxInt, false, true)
```
ro-tex, line 135: The target field is `int`, so we should use `math.MaxInt`. Even if we only ever run this on 64 bit machines, we should still stick to the correct types.

MSevey: This changed because it was breaking the lint. Maybe my local go version is behind because I didn't see this as a valid option.

MSevey (approved):

api/handlers_test.go
ro-tex: @MSevey can you show me the error you're gerring? It seems that we have some differences in our setups which shouldn't be possible. I am really interested in chasing this one because having consistently identical working environments saves a ton of time and effort in the long run. I triple-checked that `make lint` doesn't show me any warnings or errors.

ChrisSchinnerl (approved):

PR #224, opened 2022-06-23
Add a retry mechanic to requests that fail with MongoDB WriteConflict.

ro-tex:
# PULL REQUEST

## Overview

There is a general problem with multi-step transactions when several of them overlap.  

The problem goes like this:
1. txn1 reads value x, which currently sits at 5
2. txn1 makes some decisions based on that value
3. txn1 prepares to store its output to DB
4. txn2 changes the value of x from 5 to 6
5. txn2 commits
6. the engine realizes that the decisions made by txn1 are based on a now-changed value, i.e. they are invalid
7. the engine cancels txn1 and informs the client that they need to retry it, based on the new value of x

This PR adds the retry mechanic, so we can recover from such errors instead of directly returning them to the user.

## Checklist

- [ ] Changelog file created.

## Issues Closed

Closes SKY-189

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,7 +29,7 @@ else
 	- DEL /F /Q cover output
 endif
 
-run = . 
+run = .
 
 # count says how many times to run the tests.
 count = 1
@@ -136,6 +136,9 @@ test-long-ci:
 	@mkdir -p cover
 	GORACE='$(racevars)' go test -race --coverprofile='./cover/cover.out' -v -failfast -tags='testing debug netgo' -timeout=600s $(pkgs) -run=$(run) -count=$(count)
 
+test-single:
+	GORACE='$(racevars)' go test -race -v -tags='testing debug netgo' -timeout=300s $(pkgs) -run=$(run) -count=$(count)
+
 # Cookie vars
 # TODO: Are these used?
 COOKIE_HASH_KEY="7eb32cfab5014d14394648dae1cf4e606727eee2267f6a50213cd842e61c5bce"
@@ -153,4 +156,4 @@ docker-generate: clean
 	sleep 3
 	@docker stop genenv || true && docker rm --force genenv
 
-.PHONY: all deps fmt install release clean check test test-long test-long-ci start-mongo stop-mongo docker-generate
+.PHONY: all deps fmt install release clean check test test-long test-long-ci test-single start-mongo stop-mongo docker-generate
--- a/Makefile
+++ b/Makefile
@@ -148,4 +148,4 @@ docker-generate: clean
 	sleep 3
 	@docker stop genenv || true && docker rm --force genenv
 
-.PHONY: all deps fmt install release clean check test test-long test-long-ci test-single start-mongo stop-mongo docker-generate
+.PHONY: all deps fmt install release clean check test test-long test-long-ci start-mongo stop-mongo docker-generate
--- a/Makefile
+++ b/Makefile
@@ -136,8 +136,6 @@ test-long-ci:
 	@mkdir -p cover
 	GORACE='$(racevars)' go test -race --coverprofile='./cover/cover.out' -v -failfast -tags='testing debug netgo' -timeout=600s $(pkgs) -run=$(run) -count=$(count)
 
-test-single:
-	GORACE='$(racevars)' go test -race -v -tags='testing debug netgo' -timeout=300s $(pkgs) -run=$(run) -count=$(count)
 # docker-generate is a docker command for env var generation
 #
 # The sleep is to allow time for the docker container to start up after `docker
--- a/Makefile
+++ b/Makefile
@@ -85,7 +85,7 @@ ifneq ("$(OS)","Windows_NT")
 endif
 
 # Define docker container name our test MongoDB instance.
-MONGO_TEST_CONTAINER_NAME=blocker-mongo-test-db
+MONGO_TEST_CONTAINER_NAME=accounts-mongo-test-db
 
 # start-mongo starts a local mongoDB container with no persistence.
 # We first prepare for the start of the container by making sure the test
@@ -138,12 +138,6 @@ test-long-ci:
 
 test-single:
 	GORACE='$(racevars)' go test -race -v -tags='testing debug netgo' -timeout=300s $(pkgs) -run=$(run) -count=$(count)
-
-# Cookie vars
-# TODO: Are these used?
-COOKIE_HASH_KEY="7eb32cfab5014d14394648dae1cf4e606727eee2267f6a50213cd842e61c5bce"
-COOKIE_ENC_KEY="65d31d12b80fc57df16d84c02a9bb62e2bc3b633388b05e49ef8abfdf0d35cf3"
-
 # docker-generate is a docker command for env var generation
 #
 # The sleep is to allow time for the docker container to start up after `docker
--- a/api/api.go
+++ b/api/api.go
@@ -45,7 +45,13 @@ type (
 )
 
 // New returns a new initialised API.
-func New(db *database.DB, mf *metafetcher.MetaFetcher, logger *logrus.Logger, mailer *email.Mailer, deps lib.Dependencies) (*API, error) {
+func New(db *database.DB, mf *metafetcher.MetaFetcher, logger *logrus.Logger, mailer *email.Mailer) (*API, error) {
+	return NewCustom(db, mf, logger, mailer, &lib.ProductionDependencies{})
+}
+
+// NewCustom returns a new initialised API and allows specifying custom
+// dependencies.
+func NewCustom(db *database.DB, mf *metafetcher.MetaFetcher, logger *logrus.Logger, mailer *email.Mailer, deps lib.Dependencies) (*API, error) {
 	if db == nil {
 		return nil, errors.New("no DB provided")
 	}
@@ -67,10 +73,6 @@ func New(db *database.DB, mf *metafetcher.MetaFetcher, logger *logrus.Logger, ma
 			Storage:           t.Storage,
 		}
 	}
-	if deps == nil {
-		deps = &lib.ProductionDependencies{}
-	}
-
 	api := &API{
 		staticDB:            db,
 		staticDeps:          deps,
@@ -115,7 +117,10 @@ func (api *API) WithDBSession(h httprouter.Handle) httprouter.Handle {
 			_ = req.Body.Close()
 		}
 
-		for {
+		// handleFn wraps a full execution of the handler, combined with a retry
+		// detection and counting. It also takes care of creating and cancelling
+		// Mongo sessions and transactions.
+		handleFn := func() (retry bool) {
 			req.Body = io.NopCloser(bytes.NewReader(body))
 			// Create a new db session
 			sess, err := api.staticDB.NewSession()
@@ -142,7 +147,7 @@ func (api *API) WithDBSession(h httprouter.Handle) httprouter.Handle {
 			// If the call succeeded then we're done because both the status and
 			// the response content are already written to the response writer.
 			if mw.ErrorStatus() == 0 {
-				return
+				return false
 			}
 			// If the call failed with a WriteConflict error and we still have
 			// retries left, we'll retry it. Otherwise, we'll write the error to
@@ -154,7 +159,7 @@ func (api *API) WithDBSession(h httprouter.Handle) httprouter.Handle {
 				default:
 					api.staticLogger.Tracef("Retrying call because of WriteConflict (%d out of %d). Request: %+v", numRetriesLeft, dbTxnRetryCount, req)
 					numRetriesLeft--
-					continue
+					return true
 				}
 			}
 			// If the call failed with a non-WriteConflict error  or we ran out
@@ -165,7 +170,12 @@ func (api *API) WithDBSession(h httprouter.Handle) httprouter.Handle {
 			if err != nil {
 				api.staticLogger.Warnf("Failed to write to response writer: %+v", err)
 			}
-			return
+			return false
+		}
+
+		// Keep retrying the handleFn until it returns a false, indicating that
+		// no more retries are needed or possible.
+		for handleFn() {
 		}
 	}
 }
--- a/api/api.go
+++ b/api/api.go
@@ -9,11 +9,17 @@ import (
 	"github.com/SkynetLabs/skynet-accounts/email"
 	"github.com/SkynetLabs/skynet-accounts/metafetcher"
 	"gitlab.com/SkynetLabs/skyd/build"
+	"go.mongodb.org/mongo-driver/mongo"
 
 	"github.com/julienschmidt/httprouter"
 	"github.com/sirupsen/logrus"
 	"gitlab.com/NebulousLabs/errors"
-	"go.mongodb.org/mongo-driver/mongo"
+)
+
+const (
+	// dbTxnRetryCount specifies the number of times we should retry an API
+	// call in case we run into transaction errors.
+	dbTxnRetryCount = 3
 )
 
 type (
@@ -82,33 +88,57 @@ func (api *API) ListenAndServe(port int) error {
 	return http.ListenAndServe(fmt.Sprintf(":%d", port), api.staticRouter)
 }
 
-// WithDBSession injects a session context into the request context of the handler.
+// WithDBSession injects a session context into the request context of the
+// handler. In case of a MongoDB WriteConflict error, the call is retried up to
+// dbTxnRetryCount times or until the request context expires.
 func (api *API) WithDBSession(h httprouter.Handle) httprouter.Handle {
 	return func(w http.ResponseWriter, req *http.Request, ps httprouter.Params) {
-		// Create a new db session
-		sess, err := api.staticDB.NewSession()
-		if err != nil {
-			api.WriteError(w, errors.AddContext(err, "failed to start a new mongo session"), http.StatusInternalServerError)
+		numRetriesLeft := dbTxnRetryCount
+		for {
+			// Create a new db session
+			sess, err := api.staticDB.NewSession()
+			if err != nil {
+				api.WriteError(w, errors.AddContext(err, "failed to start a new mongo session"), http.StatusInternalServerError)
+				return
+			}
+			// Close session after the handler is done.
+			defer sess.EndSession(req.Context())
+			// Create session context.
+			sctx := mongo.NewSessionContext(req.Context(), sess)
+			// Get the special response writer.
+			mw, err := NewMongoWriter(w, sctx, api.staticLogger)
+			if err != nil {
+				api.WriteError(w, errors.AddContext(err, "failed to start a new transaction"), http.StatusInternalServerError)
+				return
+			}
+			// Create a new request with our session context.
+			req = req.WithContext(sctx)
+			// Forward the new response writer and request to the handler.
+			h(&mw, req, ps)
+
+			// If the call succeeded then we're done because both the status and
+			// the response content are already written to the response writer.
+			if mw.ErrorStatus() == 0 {
+				return
+			}
+			// If the call failed with a WriteConflict error and we still have
+			// retries left, we'll retry it. Otherwise, we'll write the error to
+			// the response writer and finish the call.
+			if mw.FailedWithWriteConflict() && numRetriesLeft > 0 {
+				api.staticLogger.Tracef("Retrying call because of WriteConflict (%d out of %d). Request: %+v", numRetriesLeft, dbTxnRetryCount, req)
+				numRetriesLeft--
+				continue
+			}
+			// If the call failed with a non-WriteConflict error  or we ran out
+			// of retries, we write the error and status to the response writer
+			// and finish the call.
+			w.WriteHeader(mw.ErrorStatus())
+			_, err = w.Write(mw.ErrorBuffer())
+			if err != nil {
+				api.staticLogger.Warnf("Failed to write to response writer: %+v", err)
+			}
 			return
 		}
-		// Close session after the handler is done.
-		defer sess.EndSession(req.Context())
-
-		// Create session context.
-		sctx := mongo.NewSessionContext(req.Context(), sess)
-
-		// Get the special response writer.
-		mw, err := NewMongoWriter(w, sctx, api.staticLogger)
-		if err != nil {
-			api.WriteError(w, errors.AddContext(err, "failed to start a new transaction"), http.StatusInternalServerError)
-			return
-		}
-
-		// Create a new request with our session context.
-		req = req.WithContext(sctx)
-
-		// Forward the new response writer and request to the handler.
-		h(&mw, req, ps)
 	}
 }
 
--- a/api/api.go
+++ b/api/api.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/SkynetLabs/skynet-accounts/database"
 	"github.com/SkynetLabs/skynet-accounts/email"
+	"github.com/SkynetLabs/skynet-accounts/lib"
 	"github.com/SkynetLabs/skynet-accounts/metafetcher"
 	"gitlab.com/SkynetLabs/skyd/build"
 	"go.mongodb.org/mongo-driver/mongo"
@@ -28,6 +29,7 @@ type (
 	// API is the central struct which gives us access to all subsystems.
 	API struct {
 		staticDB            *database.DB
+		staticDeps          lib.Dependencies
 		staticMF            *metafetcher.MetaFetcher
 		staticRouter        *httprouter.Router
 		staticLogger        *logrus.Logger
@@ -43,7 +45,7 @@ type (
 )
 
 // New returns a new initialised API.
-func New(db *database.DB, mf *metafetcher.MetaFetcher, logger *logrus.Logger, mailer *email.Mailer) (*API, error) {
+func New(db *database.DB, mf *metafetcher.MetaFetcher, logger *logrus.Logger, mailer *email.Mailer, deps lib.Dependencies) (*API, error) {
 	if db == nil {
 		return nil, errors.New("no DB provided")
 	}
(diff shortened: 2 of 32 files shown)


Makefile
```diff
@@ -136,6 +136,8 @@ test-long-ci:
 	@mkdir -p cover
 	GORACE='$(racevars)' go test -race --coverprofile='./cover/cover.out' -v -failfast -tags='testing debug netgo' -timeout=600s $(pkgs) -run=$(run) -count=$(count)
 
+test-single:
```
MSevey, line 139: You don't need this. 

You just need `run` to be previously defined.

then you can run `make test-long pkgs=mypkg run=mytest`

MSevey: what is the error you get when you try and run a single test that is making you need this additional command?

api/api.go
```diff
@@ -82,33 +90,72 @@ func (api *API) ListenAndServe(port int) error {
...
+		for {
+			req.Body = io.NopCloser(bytes.NewReader(body))
+			// Create a new db session
+			sess, err := api.staticDB.NewSession()
+			if err != nil {
+				api.WriteError(w, errors.AddContext(err, "failed to start a new mongo session"), http.StatusInternalServerError)
+				return
+			}
+			// Close session after the handler is done.
+			defer sess.EndSession(req.Context())
```
MSevey, line 121: If we us defers like this in a loop, aren't we overwritting the defer call and not actually closing the previous calls? i.e. only the lass `defer` session is closed or the last session is closed multiple times because `sess` is overwritten?

Instead, should we refactor this code into a function that does fully close out the session each time before the next retry?

api/api.go
```diff
@@ -1,19 +1,27 @@
...
 	"github.com/julienschmidt/httprouter"
 	"github.com/sirupsen/logrus"
 	"gitlab.com/NebulousLabs/errors"
-	"go.mongodb.org/mongo-driver/mongo"
+)
+
+const (
+	// dbTxnRetryCount specifies the number of times we should retry an API
+	// call in case we run into transaction errors.
+	dbTxnRetryCount = 5
```
MSevey, line 24: Do the accounts endpoints have a general timeout param?

I'm wondering if this should be a `MaxTxnRetryCount` so that if an endpoint has a timeout param, we quit early with an error instead of continuing to retry. 

I'm not sure if we have any instances where accounts endpoint now potentially taking 5x longer would cause issues.

Might be worth a follow up discussion at least, but not necessarily blocking this PR.

api/mongowriter.go
```diff
@@ -49,3 +86,35 @@ func (mw *MongoWriter) WriteHeader(statusCode int) {
...
+
+// ErrorStatus returns the status code with which the last call errored out,
+// if any.
+func (mw *MongoWriter) ErrorStatus() int {
+	return mw.ew.Status
+}
+
+// FailedWithWriteConflict informs us whether the MongoWriter received a MongoDB
+// WriteConflict error.
+func (mw *MongoWriter) FailedWithWriteConflict() bool {
```
MSevey, line 103: We should write a basic unit test that hits these methods.

test/api/api_test.go
```diff
@@ -138,6 +139,72 @@ func TestWithDBSession(t *testing.T) {
...
+	// so they all start at the same time. We want to keep the number of
+	// conflicting goroutines low because we want the WriteConflict to resolve
+	// within the given dxTxnRetryCount attempts.
+	ch := make(chan int)
+	var wg sync.WaitGroup
+	for i := 0; i < 3; i++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			<-ch
```
MSevey, line 197: I'm not sure if this is sufficient to guarantee there is a write conflict, unless you wrote this test first and proved that current state had write conflicts with this test.

One way around this would be with a test dependency that randomly sleeps for a certain amount of time in the actual mongo write, so some txns will start, and then sleep for a second or two before finishing while others complete with no sleep.

Makefile
ro-tex: `make test-long` always starts with a clean mongo and ends up with a wiped mongo.
This command allows me to play around with the content of the DB both before and after the tests are done. This is incredibly helpful when trying to chase a bug.

My typical use of this command is `make stop-mongo && make start-mongo && make test-single run=TestWithDBSession_RetryOnWriteConflict`.

I am using this command a lot. It's helpful to me. I would like to keep it. What is the rationale for removing it?

api/mongowriter.go
ro-tex: Added.

Makefile
MSevey: If it is simply to avoid start and stopping mongo, to keep the mongo state, why not just use `test-long-ci`. That is already refactored.  The only difference is the coverprofile right?

I see the value in running tests that way, so we should support that for sure.

api/api.go
MSevey: discussed on discord, goal is not to keep it open longer, but to ensure we are ending the call with whichever is faster. The number of retries or a provided/default timeout.

api/api.go
MSevey: sounds like this is fine.

fwiw my thought for containing it was simply something like
```go
for {
  err = func()
 // check err and decide to retry or not

}


func() {
 create session
 defer close session
}
```

fine to resolve.

api/handlers.go
```diff
@@ -742,6 +742,10 @@ func (api *API) userPUT(u *database.User, w http.ResponseWriter, req *http.Reque
 		changedEmail = true
 	}
 
+	if api.staticDeps.Disrupt("DependencyUserPutMongoDelay") {
```
MSevey, line 745: So my original suggestion included something like this, so we only sleep half the time, this would ensure other transactions that are submitted afterwards and don't sleep get executed first. 
```suggestion
	if api.staticDeps.Disrupt("DependencyUserPutMongoDelay") && fastrand.Int(0,2) == 0 {
```

test/api/api_test.go
MSevey: 50 out of 50 is pretty good 😄 

Added comment on the dependency as to clear up what my original suggestion was.

Makefile
ro-tex: Fine, we can remove it.

I don't see any reason not to have it there, as I find it helpful. But I also don't want to battle over it again and again. I'll just keep my local tools in a stash that I'll apply and drop.

Makefile
```diff
@@ -148,4 +148,4 @@ docker-generate: clean
 	sleep 3
 	@docker stop genenv || true && docker rm --force genenv
 
-.PHONY: all deps fmt install release clean check test test-long test-long-ci start-mongo stop-mongo docker-generate
+.PHONY: all deps fmt install release clean check test test-long test-long-ci test-single start-mongo stop-mongo docker-generate
```
MSevey, line 151: I'm not against new commands, I'm just not seeing the difference between this command and `test-long-ci`.

MSevey (dismissed):

Makefile
MSevey: this can be removed as a follow up.

api/api.go
```diff
@@ -57,9 +67,13 @@ func New(db *database.DB, mf *metafetcher.MetaFetcher, logger *logrus.Logger, ma
 			Storage:           t.Storage,
 		}
 	}
+	if deps == nil {
+		deps = &lib.ProductionDependencies{}
```
ChrisSchinnerl, line 71: We usually have 2 constructors to avoid these conditional initialisations. 
`NewCustom` which accepts a dependency and `New` which calls `NewCustom` with the production dependency.
In testing we call `NewCustom` to inject dependencies and in production we just call `New`. That way you don't end up with a ton of calls that have `nil` passed in.

api/api.go
```diff
@@ -82,33 +96,77 @@ func (api *API) ListenAndServe(port int) error {
...
+		for {
+			req.Body = io.NopCloser(bytes.NewReader(body))
+			// Create a new db session
+			sess, err := api.staticDB.NewSession()
+			if err != nil {
+				api.WriteError(w, errors.AddContext(err, "failed to start a new mongo session"), http.StatusInternalServerError)
+				return
+			}
+			// Close session after the handler is done.
+			defer sess.EndSession(req.Context())
```
ChrisSchinnerl, line 127: I would extract the content of this for-loop into a separate function that you call once per iteration. Then this `defer` is actually going to end the session between iteration vs right now it's just accumulating defers and ending sessions at the end of this middleware.

api/mongowriter.go
```diff
@@ -49,3 +95,35 @@ func (mw *MongoWriter) WriteHeader(statusCode int) {
...
+}
+
+// Header implementation.
+func (w *bufferResponseWriter) Header() http.Header {
+	return http.Header{}
+}
+
+// Write implementation.
+func (w *bufferResponseWriter) Write(b []byte) (int, error) {
+	return w.Buffer.Write(b)
```
ChrisSchinnerl, line 123: This is something we definitely should keep in mind. We were using a stdlib TimeoutHandler in skyd once which also buffered the whole response without us knowing and as you can imagine downloading a 10GiB would cause 10GiB of RAM usage which wasn't ideal.
So this is something that only works as long as responses are reasonably small.

lib/dependencies.go
```diff
@@ -15,3 +15,22 @@ type ProductionDependencies struct{}
 func (pd *ProductionDependencies) Disrupt(_ string) bool {
 	return false
 }
+
+/* Test dependencies */
```
ChrisSchinnerl, line 19: Is there a reason these are not just in `test/dependencies.go`? Theoretically we should only need the interface and ProductionDependencies in a neutral package and specific test dependencies can be created in the same files as the tests or a dedicated dependencies file in the test dir.

test/api/api_test.go
```diff
@@ -138,13 +140,77 @@ func TestWithDBSession(t *testing.T) {
 	}
 }
 
+// TestWithDBSession_RetryOnWriteConflict ensures that WithDBSession can
+// properly retry requests on MongoDB WriteConflict error.
+func TestWithDBSession_RetryOnWriteConflict(t *testing.T) {
```
ChrisSchinnerl, line 145: Can you add another, a bit more deterministic test? As a suggestion I think you could create a new dependency which simply forces a handler to return the write conflict error n times.

Then you run the test once with 1 failure, once with `maxRetries` failures and once with `maxRetries+1` failures to make sure it passes twice and fails once.

api/handlers.go
```diff
@@ -742,6 +743,10 @@ func (api *API) userPUT(u *database.User, w http.ResponseWriter, req *http.Reque
 		changedEmail = true
 	}
 
+	if api.staticDeps.Disrupt("DependencyUserPutMongoDelay") && fastrand.Intn(2) == 0 {
```
peterjan, line 746: kind of strange to see a `Disrupt` in combination with a chance factor, how does this not introduce NDFs?

api/mongowriter.go
```diff
@@ -1,26 +1,70 @@
...
-// MongoWriter is a custom http.ResponseWriter to commit or abort transactions.
-type MongoWriter struct {
-	logger *logrus.Logger
-	sctx   mongo.SessionContext
-	w      http.ResponseWriter
-}
+const (
+	// writeConflictErrMsg is the error message MongoDB issues when a
+	// transaction needs to be reverted because of a write conflict.
+	writeConflictErrMsg = "(WriteConflict) WriteConflict error: this operation conflicted with another operation. Please retry your operation or multi-document transaction."
```
peterjan, line 16: not sure whether this is the safe thing to do, if they roll out a driver update that slightly changes copy here our code breaks and it'll go unnoticed, is there an option to use the `WriteConcernError` from the driver package and analyse the `WriteErrors` individually to detect the write conflict?

test/api/api_test.go
```diff
@@ -138,13 +140,77 @@ func TestWithDBSession(t *testing.T) {
...
+
+	// Create a test user.
+	userEmail := types.NewEmail(t.Name() + "@siasky.net")
+	userPassword := t.Name() + "pass"
+	_, b, err := at.UserPOST(userEmail.String(), userPassword)
+	if err != nil {
+		t.Fatal(err, string(b))
+	}
+	defer func() {
+		_, _ = at.UserDELETE()
```
peterjan, line 180: so the account tester keeps track of the test user internally? I would've expected that you need to supply a username or id to the delete method.

test/api/api_test.go
```diff
@@ -138,13 +140,77 @@ func TestWithDBSession(t *testing.T) {
...
+	testAPI, err := api.New(at.DB, nil, at.Logger, nil, nil)
+	if err != nil {
+		t.Fatal("Failed to instantiate API.", err)
+	}
+	handler := func(_ http.ResponseWriter, _ *http.Request, _ httprouter.Params) {}
+	req := (&http.Request{}).WithContext(context.Background())
+	testAPI.WithDBSession(handler)(nil, req, nil)
+
+	// Create a test user.
+	userEmail := types.NewEmail(t.Name() + "@siasky.net")
```
peterjan, line 173: `nit` might as well do `userEmail := types.NewEmail(t.Name() + "@siasky.net").String()` 🤷‍♂️

test/api/api_test.go
```diff
@@ -138,13 +140,77 @@ func TestWithDBSession(t *testing.T) {
...
 }
 
+// TestWithDBSession_RetryOnWriteConflict ensures that WithDBSession can
+// properly retry requests on MongoDB WriteConflict error.
+func TestWithDBSession_RetryOnWriteConflict(t *testing.T) {
+	if testing.Short() {
+		t.SkipNow()
+	}
+	dbName := test.DBNameForTest(t.Name())
+	dep := lib.NewDependencyUserPutMongoDelay()
```
peterjan, line 150: It's always good to drop a comment above the dependency you inject that indicates what it does and why you need it. It's not clear to me why you need to delay the write for 100ms in 50% of the cases for this test to pass.

api/mongowriter_test.go
```diff
@@ -0,0 +1,202 @@
...
+	contentOK := "this is OK content"
+	n, err := mw.Write([]byte(contentOK))
+	if err != nil {
+		t.Fatal(err)
+	}
+	if n != len(contentOK) {
+		t.Fatalf("Expected %d bytes to be written, got %d", len(contentOK), n)
+	}
+	// Expect to find the content in the original response writer, i.e. testRW.
+	if subtle.ConstantTimeCompare(testRW.Buffer.Bytes(), []byte(contentOK)) != 1 {
```
peterjan, line 108: why do you use `subtle.ConstantTimeCompare` instead of `bytes.Equal`?

api/mongowriter.go
```diff
@@ -49,3 +95,35 @@ func (mw *MongoWriter) WriteHeader(statusCode int) {
...
+	return mw.ew.Status
+}
+
+// FailedWithWriteConflict informs us whether the MongoWriter received a MongoDB
+// WriteConflict error.
+func (mw *MongoWriter) FailedWithWriteConflict() bool {
+	return mw.ew.Status != 0 && strings.Contains(mw.ew.Buffer.String(), writeConflictErrMsg)
+}
+
+// Header implementation.
```
peterjan, line 116: maybe these should mention the interface you're implementing? or maybe also indicate why it's fine to return an empty header, these docstrings are quite dry `// WriteHeader implementation.` | `// Write implementation.`

api/api.go
ro-tex: Done.

lib/dependencies.go
ro-tex: If we put the dependencies under `test`, similar to what `skyd` is doing, those would be compiled in the binary.

If we put them in the test file where they are used (`api_test.go`) they won't be compiled in the binary but it would make them harder to reuse and it will generally make the project more disorganised.

I'll follow the `skyd` approach for now. Please let me know if you prefer the second approach.
I don't feel strongly about this, so I will do whatever you prefer.

api/handlers.go
ro-tex: Honestly, I can't say. I don't it will. At worst, I guess it might create a false positive in which we break the retry mechanic but the test still passes.

This is a suggestion Matt did twice and I accepted. I'm not sure what it adds to the test but it shouldn't be an NDF.
See https://github.com/SkynetLabs/skynet-accounts/pull/224#discussion_r908836955

test/api/api_test.go
ro-tex: The tester keeps the user's auth token as internal state. This allows us to test authenticated and unauthenticated API calls. This is done for convenience. It can be replaced with something like `at.WithIdentity(token).UserDELETE()`, if needed.

What is the concern with the internal state?

test/api/api_test.go
ro-tex: Erm, sure.

api/handlers.go
ro-tex: Thinking about this more, I don't see a reason to add this randomness. I'll remove it.

@MSevey we can talk about it again if you think that we need it. I really don't see how it's helpful but maybe I'm just missing the point.

test/api/api_test.go
ro-tex: Added comment, removed randomness.

test/api/api_test.go
ro-tex: Added `testWithDBSessionRetryOnWriteConflictRepeatedFailures`.

I took a slightly different approach with the retries but I think I captured the intention. If you prefer a test which we actually run three times with different parameters, I can write that too but it seemed overly complicated while achieving the same thing.

test/api/api_test.go
```diff
@@ -5,27 +5,167 @@ import (
...
 	"go.sia.tech/siad/build"
 
-	"github.com/julienschmidt/httprouter"
-	"github.com/sirupsen/logrus"
 	"gitlab.com/NebulousLabs/errors"
 )
 
-// TestWithDBSession ensures that database transactions are started, committed,
-// and aborted properly.
+// TestUserTierCache ensures out tier cache works as expected.
```
ro-tex, line 26: This is an unchanged test, feel free to ignore.

test/api/api_test.go
```diff
@@ -154,106 +298,98 @@ func TestUserTierCache(t *testing.T) {
...
-	sl, _, err := test.CreateTestUpload(at.Ctx, at.DB, *u.User, database.UserLimits[u.Tier].Storage+1)
-	if err != nil {
-		t.Fatal(err)
+	wg.Wait()
+}
+
+// testWithDBSessionRetryOnWriteConflictRepeatedFailures ensures that
+// WithDBSession will properly retry up to the given number of times and will
+// fail after.
+func testWithDBSessionRetryOnWriteConflictRepeatedFailures(t *testing.T) {
```
ro-tex, line 350: This is the only new test in here.

api/mongowriter.go
peterjan: yeah that's what I meant, you can cast the error as a mongo type though and then parse through its write errors and look for `WriteConflict`, but I think just doing the `strings.Contains` is probably fine also

test/api/api_test.go
peterjan: no concerns, will resolve

api/api.go
```diff
@@ -82,33 +98,85 @@ func (api *API) ListenAndServe(port int) error {
...
 
-		// Get the special response writer.
-		mw, err := NewMongoWriter(w, sctx, api.staticLogger)
-		if err != nil {
-			api.WriteError(w, errors.AddContext(err, "failed to start a new transaction"), http.StatusInternalServerError)
-			return
+		// handleFn wraps a full execution of the handler, combined with a retry
+		// detection and counting. It also takes care of creating and cancelling
+		// Mongo sessions and transactions.
+		handleFn := func() (retry bool) {
```
peterjan, line 123: the named return variable is not necessary here right?

api/api.go
```diff
@@ -82,33 +98,85 @@ func (api *API) ListenAndServe(port int) error {
...
+			req.Body = io.NopCloser(bytes.NewReader(body))
+			// Create a new db session
+			sess, err := api.staticDB.NewSession()
+			if err != nil {
+				api.WriteError(w, errors.AddContext(err, "failed to start a new mongo session"), http.StatusInternalServerError)
+				return
```
peterjan, line 129:
```suggestion
				return false
```

api/api.go
```diff
@@ -82,33 +98,85 @@ func (api *API) ListenAndServe(port int) error {
...
+			// Close session after the handler is done.
+			defer sess.EndSession(req.Context())
+			// Create session context.
+			sctx := mongo.NewSessionContext(req.Context(), sess)
+			// Get a special response writer which provide the necessary tools
+			// to retry requests on error.
+			mw, err := NewMongoWriter(w, sctx, api.staticLogger)
+			if err != nil {
+				api.WriteError(w, errors.AddContext(err, "failed to start a new transaction"), http.StatusInternalServerError)
+				return
```
peterjan, line 140:
```suggestion
				return false
```

test/api/api_test.go
```diff
@@ -154,106 +298,98 @@ func TestUserTierCache(t *testing.T) {
...
-		return nil
-	})
+	at.SetCookie(test.ExtractCookie(r))
+
+	// Try to update the user. Expect this to fail due to the dependency.
+	_, _, err = at.UserPUT(userEmailStr, "new password", "")
+	if err == nil || !strings.Contains(err.Error(), dependencies.DependencyMongoWriteConflictNMessage) {
+		t.Fatalf("Expected a '%s' error, got '%s'", dependencies.DependencyMongoWriteConflictNMessage, err)
+	}
+	// Try to update the user again. Expect this to pass after one retry.
```
peterjan, line 391: should we assert the PUT was successful? so check whether the password effectively updated?

api/api.go
```diff
@@ -82,33 +98,85 @@ func (api *API) ListenAndServe(port int) error {
...
+				select {
+				case <-req.Context().Done():
+					// If the request context has expired we won't retry anymore.
+				default:
+					api.staticLogger.Tracef("Retrying call because of WriteConflict (%d out of %d). Request: %+v", numRetriesLeft, DBTxnRetryCount, req)
+					numRetriesLeft--
+					return true
+				}
+			}
+			// If the call failed with a non-WriteConflict error  or we ran out
```
peterjan, line 165:
```suggestion
			// If the call failed with a non-WriteConflict error or we ran out
```

peterjan, line 165: I think we definitely need an error log if we're out of retries to warn us that for whatever reason we can't get the transaction to pass even after retrying it. The 5 retries feels very arbitrary to me, especially since there's no exponential backoff mechanism or random x ms sleep.

api/handlers.go
```diff
@@ -742,6 +743,14 @@ func (api *API) userPUT(u *database.User, w http.ResponseWriter, req *http.Reque
 		changedEmail = true
 	}
 
+	if api.staticDeps.Disrupt("DependencyUserPutMongoDelay") {
+		time.Sleep(100 * time.Millisecond)
+	}
+	if api.staticDeps.Disrupt("DependencyMongoWriteConflictN") {
```
peterjan, line 749: This is a bit funky? This is essentially a disrupt that returns a custom error message, and then you have a unit test that sets the dependency, call the route and verify you get a failure N times, and then a success? That's really only a test to make sure this `Disrupt` call works, and I guess aside from that it's also asserting the code works if you ensure there's only one transaction to the user object.

api/api.go
peterjan: we avoid named return variables in `skyd` at all cost

api/handlers.go
peterjan: I expected to see the dependency closer to the database, ideally before an update or right after it, so it kicks in after `N` times and returns the error after an interaction with a database, rather than at the top of a handler.

peterjan (approved):
approved since all comments I have are non-critical

test/dependencies/mongo.go
```diff
@@ -0,0 +1,49 @@
...
+	// DependencyMongoWriteConflictNMessage is the error message with which the
+	// DependencyMongoWriteConflictN will cause a function to fail.
+	DependencyMongoWriteConflictNMessage = "(WriteConflict) This is a WriteConflict caused by DependencyMongoWriteConflictN."
+)
+
+type (
+	// DependencyMongoWriteConflictN causes the caller to fail with a
+	// WriteConflict N times in a row.
+	DependencyMongoWriteConflictN struct {
+		remainingFailures uint
```
MSevey, line 15: If we are testing concurrency issues, should this dependency have a mutex? Might be good to prevent future data races. 

Fine as a follow up.

api/api.go
ro-tex: Interesting. Why is that? I am not aware of any drawbacks to using named return values.
There is nothing said about those in our Engineering Guidelines.

Named return values are a nice and unobtrusive way to document a function.
Here's what Effective Go has to say about them: https://go.dev/doc/effective_go#named-results

api/handlers.go
ro-tex: Done.

test/dependencies/mongo.go
ro-tex: Good idea. Added.

api/api.go
peterjan: Well I like them too to be honest, but in `skyd` we avoid them when we can because it's often a foot gun and can lead to unexpected bugs 🤷‍♂️ , it's easy to miss a place where you overwrite it or don't set it when you want it to be set.

I personally am in your camp and like the documenting aspect of it, and the fact that you skip the var initialisation, but I'm just doing `skyd` reviews here, I don't really mind whether you update it or not.

MSevey (approved):

api/api.go
ChrisSchinnerl: The way it is used here it is probably fine. But for future reference we usually try to avoid them, especially in longer methods. Unless they significantly simplify code. e.g. when you need to close something in an error case.

The reason is that for longer functions they can really harm readability and can lead to weird errors. e.g. we had a surprising number of issues where we'd miss assignments to named return vars or shadow them and then miss errors and such.

If it was just for the sake of documentation this could be pulled out into a separate function `WithRetry` that has its own docstring and an explanation of what the return value is. Which should probably happen anyway in a f/u since it's enough code.

api/mongowriter.go
ChrisSchinnerl: Yeah me neither. Just something to remember in case we ever run into it. No immediate action item here.

ChrisSchinnerl (approved):

PR #228, opened 2022-07-05
Stripe promote on checkout

ro-tex:
# PULL REQUEST

## Overview

Our typical payment flow starts with us creating a checkout session and redirecting the user to a Stripe-controlled page where they make the actual payment. After the payment is done, we wait for a push notification on our webhook endpoint. We process that notification and promote/demote the user accordingly.
This works well and is reliable because Stripe will keep sending the webhook event until they get a successful response.
The problem with it is that it might take a while and the user will be left without any notification what is going on.

This PR solves that problem by introducing a new endpoint which we can proactively call in order to check the status of a given checkout session and promote the user, if required. Note that we won't demote the user by calling this endpoint - that will be handled by the webhook endpoint. This new endpoint will only be used in order to speed up the user promotion process, so we get happy users.

## Example for Visual Changes

Response JSON example:
```json
{
    "created": 1658417273,
    "currentPeriodStart": 1658417273,
    "discount": {
        "amountOff": 0,
        "currency": "",
        "duration": "repeating",
        "durationInMonths": 3,
        "name": "test50%",
        "percentOff": 50
    },
    "id": "sub_1LO1I5IzjULiPWN6zatWjjxl",
    "plan": {
        "amount": 2000,
        "currency": "usd",
        "interval": "month",
        "intervalCount": 1,
        "product": {
            "description": "Pin up to 4TB",
            "name": "Skynet Pro"
        }
    },
    "startDate": 1658417273,
    "status": "active"
}
```

## Checklist

- [ ] Changelog file created.

## Issues Closed
Closes SKY-1056


api/stripe.go
```diff
@@ -74,6 +72,21 @@ type (
 		ProductID   string  `json:"productId"`
 		LiveMode    bool    `json:"livemode"`
 	}
+	// SubscriptionInfoGET is a basic description of the subscription.
+	SubscriptionInfoGET struct {
```
ro-tex, line 76: This is a subset of the Stripe's Subscription object. What gets in here is entirely up to @meeh0w, what I've put here for the moment is just a demo.

test/api/stripe_test.go
```diff
@@ -147,6 +151,134 @@ func testStripeCheckoutPOST(t *testing.T, at *test.AccountsTester) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	if u.Tier != database.TierPremium20 {
+		t.Fatalf("Expected tier %d, got %d", database.TierPremium20, u.Tier)
+	}
+	// Get the info on a checkout session that hasn't been completed and
+	// doesn't have a subscription assigned to it, yet.
+	info, status, err = at.StripeCheckoutIDGET(sessionIDWithoutSub)
+	errStr := "this checkout session does not have an associated subscription"
```
peterjan, line 262: I think it would be good to extract these error strings in a constant and then reuse the error constant in the tests, that way the test won't fail if you update the error message later, also it's more readable if it would've been `errNoAssociatedSubscription` or w/e

test/api/stripe_test.go
```diff
@@ -147,6 +151,134 @@ func testStripeCheckoutPOST(t *testing.T, at *test.AccountsTester) {
...
+	defer func(c *http.Cookie) {
+		at.SetCookie(c)
+		_, _ = at.UserDELETE()
+	}(c)
+	u, _, err := at.UserGET()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Fixture values.
```
peterjan, line 174: maybe move it into constants on top of the file so it's reusable by other tests and it doesn't clutter the actual test

test/api/stripe_test.go
```diff
@@ -147,6 +151,134 @@ func testStripeCheckoutPOST(t *testing.T, at *test.AccountsTester) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	defer gock.Off()
+	// We need to enable networking in order to allow the Tester to call our
+	// own API.
+	gock.EnableNetworking()
+	// Set up a response that will upgrade the user to tier 20.
+	gock.New("https://api.stripe.com").
```
peterjan, line 200: could be a helper called `mockSessionResponse` 

```go
// mock tier 20 success
res := strings.NewReader(fixtures.StripeCheckoutSessionWithSubTier20)
mockSessionResponse(sessionIDWithSub20, http.StatusOK, res)

// mock tier 20 failure
mockSessionResponse(sessionIDWithSub20, http.StatusOK, nil)
```

peterjan: I usually also move this to right before I'm doing the assertions instead of a big block in the beginning of the test:

```go
// mock tier 20 success
res := strings.NewReader(fixtures.StripeCheckoutSessionWithSubTier20)
mockSessionResponse(sessionIDWithSub20, http.StatusOK, res)

// write tier 20 success assertions here
...

// mock tier 20 failure

// write tier 30 failure assertions here
```

api/stripe.go
```diff
@@ -226,6 +239,67 @@ func (api *API) stripeCheckoutPOST(u *database.User, w http.ResponseWriter, req
...
+	// Get the subscription price.
+	if coSub.Items == nil || len(coSub.Items.Data) == 0 || coSub.Items.Data[0].Price == nil {
+		api.WriteError(w, errors.New("subscription does not have a price"), http.StatusBadRequest)
+		return
+	}
+	coSubPrice := coSub.Items.Data[0].Price
+	tier, exists := StripePrices()[coSubPrice.ID]
+	if !exists {
+		err = fmt.Errorf("invalid price id '%s'", coSubPrice.ID)
+		api.WriteError(w, err, http.StatusBadRequest)
```
peterjan, line 280: I think this should be an internal server error

peterjan (commented):
looks good, would clean it up a bit but definitely close to done

api/stripe.go
```diff
@@ -226,6 +239,67 @@ func (api *API) stripeCheckoutPOST(u *database.User, w http.ResponseWriter, req
...
+		build.Critical(errors.AddContext(err, "We somehow received an invalid price ID from Stripe. This might be caused by mismatched test/prod tokens or a breakdown in our Stripe setup."))
+		return
+	}
+	// Promote the user, if needed.
+	if tier > u.Tier {
```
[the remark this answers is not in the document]
ro-tex: 1. Stripe won't report on unfinalised payment. I don't think it's possible for the user to manipulate this in such a way. And even if they managed to do that, we'd still get a webhook call from Stripe, during which we'd set the right tier. This usually happens within a couple of seconds.
2. In failure case the entire txn gets rolled back. I can add a test case.

api/stripe.go
```diff
@@ -226,6 +239,67 @@ func (api *API) stripeCheckoutPOST(u *database.User, w http.ResponseWriter, req
...
+			Expand: []*string{&subStr},
+		},
+	}
+	cos, err := cosession.Get(checkoutSessionID, params)
+	if err != nil {
+		api.WriteError(w, err, http.StatusInternalServerError)
+		return
+	}
+	if cos.Customer.ID != u.StripeID {
+		api.WriteError(w, errors.New("checkout session does not belong to current user"), http.StatusBadRequest)
```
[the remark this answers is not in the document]
ro-tex: Hm, I see what you mean. I would still go for Bad Request on the assumption that the caller messed up their request on accident, rather than trying to abuse the API. Kind of "assume a mistake, rather than bad intentions".

That said, I'm perfectly fine with changing this to a 403 if that seems more appropriate.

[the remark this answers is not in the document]
peterjan: yeah that assumption you propose is one that's usually not made in this case. 
if you try to do things using somebody else's credentials usually results in forbidden

peterjan (approved):

api/stripe.go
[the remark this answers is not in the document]
ro-tex: OK, changed.

ChrisSchinnerl (approved):

peterjan (approved):

PR #240, opened 2022-08-25
Implement a better health endpoint.

ro-tex:
# PULL REQUEST

## Checklist

- [ ] Changelog file created.

### DIFF
--- a/email/sender.go
+++ b/email/sender.go
@@ -8,6 +8,8 @@ import (
 	"strconv"
 	"time"
 
+	"go.mongodb.org/mongo-driver/mongo"
+
 	"github.com/SkynetLabs/skynet-accounts/database"
 	"github.com/sirupsen/logrus"
 	"gitlab.com/NebulousLabs/errors"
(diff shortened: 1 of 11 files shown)


jwt/jwt.go
```diff
@@ -134,58 +134,58 @@ func TokenSerialize(t jwt.Token) ([]byte, error) {
 //
 // Header:
 //
-// {
-//  "alg": "RS256",
-//  "kid": "a2aa9739-d753-4a0d-87ee-61f101050277",
-//  "typ": "JWT"
-// }
+//	{
```
ro-tex, line 137: This whole formatting thing happens during our linting run. I'll dig into it later.

email/sender.go
```diff
@@ -97,9 +99,13 @@ func NewSender(ctx context.Context, db *database.DB, logger *logrus.Logger, deps
...
 		return Sender{}, errors.AddContext(err, "failed to parse email config")
 	}
+	sess, err := db.NewSession()
+	if err != nil {
+		return Sender{}, err
+	}
 	return Sender{
 		staticConfig: c,
-		staticCtx:    ctx,
+		staticCtx:    mongo.NewSessionContext(ctx, sess),
```
ro-tex, line 108: This is needed, so when the sender is counting how many emails it needs to send, it doesn't need to create an ad-hoc session. Creating an ad-hoc session causes a potential data race with mongo's `Client().NumberSessionsInProgress()` which can't cause any damage but it breaks our tests.

ChrisSchinnerl (approved):

PR #242, opened 2022-08-30
Promoter: set tier endpoint

ro-tex:
# PULL REQUEST

## Overview

The `POST /promoter/settier/:sub` endpoint is a private (not exposed by nginx) endpoint which allows the caller to directly set the tier of the user with the given `sub`.

## Checklist

- [ ] Changelog file created.


api/promoter.go
```diff
@@ -0,0 +1,45 @@
...
+type (
+	// PromoterSetTierPOST describes the body of a POST request that sets the
+	// user's tier.
+	PromoterSetTierPOST struct {
+		Tier int `json:"tier"`
+	}
+)
+
+// _promoterSetTierPOST sets the given user's tier.
+func (api *API) _promoterSetTierPOST(_ *database.User, w http.ResponseWriter, req *http.Request, ps httprouter.Params) {
```
ChrisSchinnerl, line 21: Is there a reason for this new convention? The `_`? Gives me very pythony vibes. Like "don't ever use this".

api/promoter.go
```diff
@@ -0,0 +1,45 @@
...
+		api.WriteError(w, fmt.Errorf("invalid tier %d", body.Tier), http.StatusBadRequest)
+		return
+	}
+	ctx := req.Context()
+	u, err := api.staticDB.UserBySub(ctx, sub)
+	if errors.Contains(err, database.ErrUserNotFound) {
+		api.WriteError(w, err, http.StatusNotFound)
+		return
+	}
+	err = api.staticDB.UserSetTier(ctx, u, body.Tier)
```
ChrisSchinnerl, line 39: I noticed that this upserts a user. Is that intended behaviour? Might lead to a race if the user is deleted between here and `UserBySub`.

Also we need to guarantee that we can't update tiers for users which are linked to an active stripe subscription.

test/tester.go
```diff
@@ -756,3 +756,13 @@ func (at *AccountsTester) StripePricesGET() ([]api.StripePrice, int, error) {
 	}
 	return resp, r.StatusCode, nil
 }
+
+/*** Promoter helpers ***/
+
+// PromoterSetTierPOST performs a `POST /promoter/settier/:sub`
+func (at *AccountsTester) PromoterSetTierPOST(sub string, tier int) (int, error) {
+	body := api.PromoterSetTierPOST{tier}
+	bodyBytes, err := json.Marshal(body)
```
ChrisSchinnerl, line 765: Seems like this error is never actually checked

api/promoter.go
ro-tex: I wanted to do it for the endpoints themselves, so it's painfully obvious that those are internal and should never be exposed. Then I backtracked on that and the convention remained for the handlers. But that doesn't make much sense now. I'll remove it.

api/promoter.go
ro-tex: Good catch. No, this wasn't intended and it should not happen. I made that behaviour controllable via a parameter.

ChrisSchinnerl (commented):
Looks pretty good but what happens when you call the setter on an account managed by stripe or vice versa? e.g. it might be dangerous to have a promoter create a user and manage it but not prevent attaching a stripe subscription to a user.

database/user.go
```diff
@@ -551,13 +551,13 @@ func (db *DB) UserSetStripeID(ctx context.Context, u *User, stripeID string) err
 }
 
 // UserSetTier sets the user's tier to the given value.
-func (db *DB) UserSetTier(ctx context.Context, u *User, t int) error {
+func (db *DB) UserSetTier(ctx context.Context, u *User, t int, upsert bool) error {
 	if t <= TierAnonymous || t >= TierMaxReserved {
```
MSevey, line 555: Follow Up: worth connecting my other comment here as well.

test/api/promoter_test.go
```diff
@@ -0,0 +1,53 @@
...
+	if err != nil || status != http.StatusNoContent {
+		t.Fatal(status, err)
+	}
+	u1, err := at.DB.UserBySub(at.Ctx, u.Sub)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if u1.Tier != database.TierPremium20 {
+		t.Fatalf("Expected tier %d, got %d", database.TierPremium20, u1.Tier)
+	}
```
MSevey, line 52: REQ: we need to test downgrading the tier as well.

MSevey (dismissed):
Needs the test case of downgrading the tier as well.  Otherwise good to go.

api/promoter.go
```diff
@@ -0,0 +1,45 @@
...
+// promoterSetTierPOST sets the given user's tier.
+func (api *API) promoterSetTierPOST(_ *database.User, w http.ResponseWriter, req *http.Request, ps httprouter.Params) {
+	sub := ps.ByName("sub")
+	var body PromoterSetTierPOST
+	err := parseRequestBodyJSON(req.Body, LimitBodySizeLarge, &body)
+	if err != nil {
+		api.WriteError(w, err, http.StatusBadRequest)
+		return
+	}
+	if body.Tier < database.TierFree || body.Tier >= database.TierMaxReserved {
```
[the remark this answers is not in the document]
ro-tex: The only purpose of the `database.TierMaxReserved` is to point to the first tier that doesn't exist.
I'm fine with creating an issue but I'm not sure what's a better way to handle this. Maybe replace `TierFree` with a zero?

test/api/promoter_test.go
ro-tex: Added.

database/user.go
ro-tex: I revisited the need for the upsert flag and removed it.

api/promoter.go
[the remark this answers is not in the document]
ro-tex: It doesn't hurt to put this in a function. It will help people to reason about it.

F/U: https://github.com/SkynetLabs/skynet-accounts/issues/245

ro-tex:
> Looks pretty good but what happens when you call the setter on an account managed by stripe or vice versa? e.g. it might be dangerous to have a promoter create a user and manage it but not prevent attaching a stripe subscription to a user.

F/U: #246

MSevey (approved):

PR #249, opened 2022-09-06
Promoter / Stripe switch

ro-tex:
# PULL REQUEST

## Overview

Add an env var that switches between Stripe and Promoter as payment processor.

## Checklist

- [ ] Changelog file created.

### DIFF
--- a/test/api/api_test.go
+++ b/test/api/api_test.go
@@ -29,7 +29,7 @@ func TestUserTierCache(t *testing.T) {
 		t.SkipNow()
 	}
 	dbName := test.DBNameForTest(t.Name())
-	at, err := test.NewAccountsTester(dbName, nil)
+	at, err := test.NewAccountsTester(dbName, "", nil)
 	if err != nil {
 		t.Fatal(err)
 	}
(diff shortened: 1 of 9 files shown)


test/api/api_test.go
```diff
@@ -29,7 +29,7 @@ func TestUserTierCache(t *testing.T) {
 		t.SkipNow()
 	}
 	dbName := test.DBNameForTest(t.Name())
-	at, err := test.NewAccountsTester(dbName, nil)
+	at, err := test.NewAccountsTester(dbName, "", nil)
```
MSevey, line 32: nit: An approach that requires fewer line changes is having a `NewAccountsTesterCustom` that is used to add any new options that are necessary.  Then just the tests that need custom options have to be edited and all these default cases don't need to be touched. 

Something to consider for the future.

test/tester.go
```diff
@@ -78,11 +78,14 @@ func NewDatabaseWithDeps(ctx context.Context, dbName string, deps lib.Dependenci
 
 // NewAccountsTester creates and starts a new AccountsTester service.
 // Use the Close method for a graceful shutdown.
-func NewAccountsTester(dbName string, deps lib.Dependencies) (*AccountsTester, error) {
+func NewAccountsTester(dbName string, promoter api.Promoter, deps lib.Dependencies) (*AccountsTester, error) {
 	// Make sure we have valid dependencies.
 	if deps == nil {
 		deps = &lib.ProductionDependencies{}
 	}
+	if promoter == "" {
```
MSevey, line 86: then here, the custom handler requires none nil values for all of these and `NewAccountsTester` just calls the custom handler with defaults.

MSevey (approved):

ChrisSchinnerl (approved):

test/api/api_test.go
ro-tex: Good advice. I will probably add that in a F/U.

